PR #1, opened 2020-08-04
initial version

edewit:


package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "test",
```
ssilvert, line 2: Should be "keycloak-admin"

src/http-service/http-client.ts
```diff
@@ -0,0 +1,149 @@
...
+  constructor(public response: HttpResponse) {
+    super(response.statusText);
+  }
+}
+
+/**
+ *
+ * @author Stan Silvert ssilvert@redhat.com (C) 2018 Red Hat Inc.
+ */
+export class HttpClient {
```
ssilvert, line 23: I wonder if this should become its own library?

edewit: there are multiple packages already:
* https://github.com/elbywan/wretch
* https://www.npmjs.com/package/@ianwalter/http
* https://www.npmjs.com/package/@schirrel/request
* https://www.npmjs.com/package/@hellosirandy/rest-api-wrapper
* https://www.npmjs.com/package/fetch-json

ssilvert: I know.  But if our HttpClient is what we want to use then we might want to make it into a library because we are using it on two projects.

PR #26, opened 2020-08-18
added initial version of the "help system"

edewit:
fixes: #25


src/components/help-enabler/HelpHeader.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <TextContent>
+        <Text component={TextVariants.small}>
+          <Trans>
+            This toggle will enable/disable part of
+            <br />
+            the help info in the console. Includes
+            <br />
+            any help text, links and popovers.
+          </Trans>
+        </Text>
```
srambach, line 74: It looks like the text color is because the small text color is still set as if it were a dark component (i.e. the page header) even though the dropdown is a light element. I'll see if this is a PF bug.
Also, we should probably set a max-width somewhere rather than putting `<br>` tags into the text, wdyt?

src/components/help-enabler/help-header.module.css
```diff
@@ -0,0 +1,8 @@
+
+.dropdown > ul {
+  right: 0;
+}
```
srambach, line 4: I think we can remove this if we use `position="right"` on the `Dropdown` component.

src/components/help-enabler/help-header.module.css
```diff
@@ -0,0 +1,8 @@
...
+
+.helpDropdownItem {
+  background-color: transparent;
+}
```
srambach, line 8: Can you tell me what this is doing? I can't see a difference when it is removed.

src/components/help-enabler/HelpHeader.tsx
```diff
@@ -0,0 +1,96 @@
...
+      id="enable"
+      component="div"
+      className={style.helpDropdownItem}
+    >
+      {t('Enable help mode') + ' '}
+      <Switch
+        id="enableHelp"
+        isChecked={help.enabled}
+        onChange={() => help.toggleHelp()}
+      />
```
srambach, line 64: I think we need to add an `aria-label="Message when on"` to add accessible text to the switch. The Patternfly site says "Adds accessible text to the Switch, and should describe the isChecked="true" state. When label is defined, aria-label should be set to the text string that is visible when isChecked is true." So I think it should say something like "Help text enabled"

src/components/help-enabler/help-header.module.css
edewit: It removes the hover background from the dropdown element, as it's more then just a simple action / link

src/components/help-enabler/help-header.module.css
edewit: Right, that is way better, didn't know about this property

srambach:
It looks like the text color of small text within a light element within a dark element IS a bug - but there is a good workaround as of PF 2020.19 where a dropdown item description was introduced. I think that will work well, but it looks like maybe we are one version behind that so would need to update PF to the current version.
https://www.patternfly.org/v4/documentation/react/components/dropdown#dropdown-with-descriptions

edewit:
I now solved it by removing the `<TextContent>` element that seemed to fix it as well

srambach:
> I now solved it by removing the `<TextContent>` element that seemed to fix it as well

Maybe we can add it as an issue to fix after updating Patternfly? It would take care of the additional styling that you've had to add for wrapping as well.

edewit:
closer to design:
image

ssilvert:
> > I now solved it by removing the `<TextContent>` element that seemed to fix it as well
> 
> Maybe we can add it as an issue to fix after updating Patternfly? It would take care of the additional styling that you've had to add for wrapping as well.

@srambach If this is the only thing holding up the PR then please open the issue, assign to Erik, and merge this.  I'll make sure the PatternFly upgrade gets done very soon.

edewit:
@srambach rebased now has the latest patternfly, but the problems still there

edewit:
it doesn't work though

edewit:
@srambach can we merge this? Maybe we can add an issue for the use of the 'description' attribute, because at the moment I can't seem to get that to work.

srambach:
@edewit You are correct - the same root problem is hitting the description property. The issue is bumped up and @mcoker is putting in a fix. https://github.com/patternfly/patternfly/issues/3411

srambach:
My preference would be to use `description` and add the class there to adjust the color until we have a version of PF with the fix, but as long as there's an issue to track it so we don't lose it, I don't care if this PR is merged as is.

edewit:
right the problem is when I add it to the description I can't add a class to change the colour or the spacing and it will look like this:
image
So I suggest we leave it like this for now and wait on a patternfly fix

PR #27, opened 2020-08-18
enhanced rendering of client list

edewit:


src/clients/ClientList.tsx
```diff
@@ -30,11 +44,25 @@ export const ClientList = ({ clients }: ClientListProps) => {
     />
   );
 
+  const enabled = (): IFormatter => (data?: IFormatterValueType) => {
+    return data ? (
+      <Chip isReadOnly isOverflowChip>
+        Enabled
+      </Chip>
+    ) : (
+      <Chip isReadOnly>Disabled</Chip>
```
srambach, line 53: In the final design it looks like these are badges next to the name instead of chips in their own column. (If I'm looking at the right thing!) Is there a plan to move these later?

edewit:
Right, I was looking at the wrong one, sorry about that. Have to look at
how to do that, I'm new to this table component.

edewit:
closer to design:
image

srambach:
Nice! I do think that the chip element should be a badge, however. `isRead` will make it the gray color shown in the mockups.
I'm checking with the content team on `Client Id` versus `Client ID`.

edewit:
changed to badge and title to "Client ID"

PR #30, opened 2020-08-19
Implement masthead

ssilvert:
This PR implements the following in masthead:
- Adds hamburger icon to show/hide left nav menu
- Hamburger menu replaces nav on smaller screen widths
- Adds regular dropdown for Manage Account/Server Info/Signout
- Adds kabab menu for Manage Account/Server Info/Help/Signout
- Kabab replaces regular menu and help menu on smaller screen widths
- User Avatar
- Change the way brand gets its height

PR also makes changes not really related to masthead:
- Update all patternfly dependencies
- Fix linting errors

### DIFF
--- a/src/PageHeader.tsx
+++ b/src/PageHeader.tsx
@@ -1,13 +1,149 @@
-import React, { useContext } from 'react';
-import { PageHeader, Brand, PageHeaderTools } from '@patternfly/react-core';
+import React, { useContext, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+        Avatar,
+        Button,
+        ButtonVariant,
+        Brand,
+        Dropdown,
+        DropdownItem,
+        DropdownSeparator,
+        DropdownToggle,
+        KebabToggle,
+        PageHeader,
+        PageHeaderTools,
+        PageHeaderToolsItem,
+        PageHeaderToolsGroup} from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
 import { KeycloakContext } from './auth/KeycloakContext';
 
 export const Header = () => {
-  const keycloak = useContext(KeycloakContext);
   return (
     <PageHeader
-      logo={<Brand src="/logo.svg" alt="Logo" style={{ height: '35px' }} />}
-      headerTools={<PageHeaderTools>{keycloak?.loggedInUser}</PageHeaderTools>}
+      showNavToggle
+      logo={<Brand src="/logo.svg" alt="Logo" />}
+      headerTools={headerTools()}
     />
   );
 };
+
+const ManageAccountDropdownItem = () => {
+  const keycloak = useContext(KeycloakContext);
+  const { t } = useTranslation();
+  return (
+    <DropdownItem key="manage account" onClick={() => keycloak?.account()}>{t('Manage account')}</DropdownItem>
+  );
+}
+
+const SignOutDropdownItem = () => {
+  const keycloak = useContext(KeycloakContext);
+  const { t } = useTranslation();
+  return (
+    <DropdownItem key="sign out" onClick={() => keycloak?.logout()}>{t('Sign out')}</DropdownItem>
+  );
+}
+
+const ServerInfoDropdownItem = () => {
+  const { t } = useTranslation();
+  return (
+    <DropdownItem key="server info">{t('Server info')}</DropdownItem>
+  )
+}
+
+const HelpDropdownItem = () => {
+  const { t } = useTranslation();
+  const help = t('Help');
+  return (
+    <DropdownItem><HelpIcon />{` ${help}`}</DropdownItem>
+  )
+}
+
+const kebabDropdownItems = [
+  <ManageAccountDropdownItem key='kebab Manage Account'/>,
+  <ServerInfoDropdownItem key='kebab Server Info'/>,
+  <HelpDropdownItem key='kebab Help'/>,
+  <DropdownSeparator key="kebab sign out seperator" />,
+  <SignOutDropdownItem key='kebab Sign out'/>
+];
+
+const userDropdownItems = [
+    <ManageAccountDropdownItem key='Manage Account'/>,
+    <ServerInfoDropdownItem key='Server info'/>,
+    <DropdownSeparator key="sign out seperator" />,
+    <SignOutDropdownItem key='Sign out'/>
+];
+
+const headerTools = () => {
+  return (
+    <PageHeaderTools>
+      <PageHeaderToolsGroup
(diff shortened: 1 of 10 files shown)


src/components/data-loader/DataLoader.tsx
```diff
@@ -19,7 +19,7 @@ export function DataLoader<T>(props: DataLoaderProps<T>) {
     loadData();
   }, [props]);
 
-  if (!!data) {
```
ssilvert, line 22: @edewit I changed this to quiet a linting error.  If this really needs to be a double-negative please let me know.

src/messages.json
```diff
@@ -11,7 +11,13 @@
...
       "fullName": "{{givenName}} {{familyName}}",
       "unknownUser": "Anonymous",
-      "Keycloak Administration Console": "RH-SSO Administration Console"
+      "Keycloak Administration Console": "RH-SSO Administration Console",
+
+      "********* MASTHEAD *********": "",
+      "Sign out": "Sign out",
+      "Manage account": "Manage account",
+      "Server info": "Server info",
+      "Help": "Help"
```
ssilvert, line 20: @edewit @srambach @dlabaj @jenny-s51 
I'd like feedback on this if you have any.  The message bundle will eventually get pretty big and we need a way to at least break it into sections.

We could divide messages into sections using a "fake json comment" like I did here or break it up into multiple files.  If you guys really hate this little trick I can go ahead and break it into multiple files.

Also, I rather like the pattern where the default bundle key is the plain English text.  That doesn't always work, such as the fullName example above.  But this makes it easier when you go to translate the text.  The downside is that it's a little harder to find missing keys.

jenny-s51: Just want to make sure I understand -- would you mind kindly clarifying why we need this file @ssilvert ? I see it imported in https://github.com/keycloak/keycloak-admin-ui/blob/d6f2a014bf8310bed2effec0613a5b01b0ec373a/src/i18n.ts but I'm not too clear on what its purpose is 🙂

src/PageHeader.tsx
```diff
@@ -1,13 +1,149 @@
...
+  const keycloak = useContext(KeycloakContext);
+  const { t } = useTranslation();
+  return (
+    <DropdownItem key="manage account" onClick={() => keycloak?.account()}>{t('Manage account')}</DropdownItem>
+  );
+}
+
+const SignOutDropdownItem = () => {
+  const keycloak = useContext(KeycloakContext);
+  const { t } = useTranslation();
```
jenny-s51, line 40: what does this do/why do we need it?

src/messages.json
ssilvert: It's the English message bundle for i18next.  When you want to support another language you supply a different bundle with the same keys but different values.  Spanish for instance:
`"Help": "Ayuda"`

src/PageHeader.tsx
ssilvert: It allows you to log out of Keycloak.

src/components/data-loader/DataLoader.tsx
edewit: guess it's fine, I guess I wanted to make sure it's a boolean...

src/forms/realm/NewRealmForm.tsx
```diff
@@ -14,11 +13,11 @@ import {
   Divider,
 } from "@patternfly/react-core";
 
-type NewRealmFormProps = {
-  realm: string;
-};
+//type NewRealmFormProps = {
+//  realm: string;
+//};
```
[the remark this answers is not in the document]
edewit: yeah, it's something that is not done yet I guess... commenting it out for now is fine

src/messages.json
edewit: @ssilvert this package is very mature and of course they have thought of that ;) We can use "namespaces" to divide it up into "sections"

edewit (approved):

src/PageHeader.tsx
jenny-s51: Oops @ssilvert , I meant the `const { t } = useTranslation();`! if you remove the `t(...)` from the `t('Sign Out')` it seems to behave the same

edewit: @jenny-s51 yes because the "key" is the same as the "value" when there is no value it uses the "key"

src/PageHeader.tsx
```diff
@@ -1,13 +1,149 @@
...
+  return (
+    <DropdownItem key="server info">{t('Server info')}</DropdownItem>
+  )
+}
+
+const HelpDropdownItem = () => {
+  const { t } = useTranslation();
+  const help = t('Help');
+  return (
+    <DropdownItem><HelpIcon />{` ${help}`}</DropdownItem>
```
yordis, line 57: Shouldn't this space be added with CSS instead of using spaces for it?

Also, I definitely recommend inline translations where it is being used instead of using an anchor variable.

ssilvert: > Shouldn't this space be added with CSS instead of using spaces for it?
Just following the PatternFly example, but that doesn't mean it's actually best.
> 
> Also, I definitely recommend inline translations where it is being used instead of using an anchor variable.
The variable just makes it easier to add the space.

I don't understand why either of your recommendations provide any benefit, but if you can explain the "why" part I'll consider a change.

src/messages.json
ssilvert: Ah, I should have guessed there was already a solution.  I'll look into that.  Thanks.

src/PageHeader.tsx
ssilvert: @yordis I need to go ahead and merge this because others are waiting on part of it.  Feel free to continue your thoughts.  There is plenty of time to make more changes.

srambach: @ssilvert  I know I'm late to the game after being on PTO, but next time I would recommend adding `className="pf-u-mr-md"` (or whatever the correct spacer size is) to the Icon in lieu of the character space; this will add a medium right margin to the icon to ensure proper spacing. As a one-off I'd use this utility class but if it's happening elsewhere we can add a custom class to keep it consistent.

ssilvert: > @ssilvert I know I'm late to the game after being on PTO, but next time I would recommend adding `className="pf-u-mr-md"` (or whatever the correct spacer size is) to the Icon in lieu of the character space; this will add a medium right margin to the icon to ensure proper spacing. As a one-off I'd use this utility class but if it's happening elsewhere we can add a custom class to keep it consistent.

Thanks @srambach 

Can you submit a PR for this?  I'll leave it up to you to decide which solution you think is best at the present time.

PR #46, opened 2020-08-27
added an external link component

edewit:


src/components/external-link/ExternalLink.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+
+export const ExternalLink = ({
+  title,
```
ssilvert, line 5: Looks like title should be marked optional here?

edewit: this is the variable the type is defined in `React.HTMLProps<HTMLAnchorElement>` and it's optional indeed

PR #48, opened 2020-08-28
add realm selector

### LINKED ISSUE #11: Realm Selector
Implement Real Selector component.

jenny-s51:
Closes #11.


src/PageNav.tsx
```diff
@@ -1,14 +1,30 @@
...
+  const [realms, setRealms] = useState([] as Realm[]);
+
+  const getRealms = async () => {
+    return await httpClient?.doGet('/realms')
+      .then((r) => r.data as Realm[]);
+    }
+  
+
+  getRealms().then((result) => {
+    setRealms(result !== undefined ? result: []);});
```
ssilvert, line 20: I don't think this stuff belongs in the PageNav.  The only place that the realm list gets used is in the dropdown.  So it probably just belongs somewhere inside the RealmSelector.

dlabaj: Agreed makes sense.  Let's add this to the dropdown.

src/auth/keycloak.service.ts
```diff
@@ -45,6 +45,8 @@ export class KeycloakService {
     return this.keycloakAuth.realm;
   }
 
+  
+
```
dlabaj, line 49: Not sure these extra spaces are need.  Can you run yarn format to make sure all files are formatted correctly.

src/components/realm-selector/RealmSelector.tsx
```diff
@@ -1,27 +1,33 @@
-import React, { useState } from 'react';
+import React, { useState, useContext } from 'react';
 import { useHistory } from 'react-router-dom';
+import { Realm } from "../../models/Realm";
```
dlabaj, line 3: @ssilvert @edewit I like the idea of having the models folder for all the data models we get back from the API.  Is this good with you?

src/context/RealmContext.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { createContext } from "react";
```
dlabaj, line 1: @ssilvert @edewit We split the Realms up into their own context so components that need them can pull in that context.  I think it's easier to manage then one giant global context.  Thoughts?  We did this on apicurio and liked it.

src/models/Realm.ts
```diff
@@ -0,0 +1,4 @@
+export interface Realm {
+    id: string
+    realm: string
+}
```
dlabaj, line 4: Looks good a few comments.

src/components/realm-selector/RealmSelector.tsx
ssilvert: @dlabaj Feels right to me.

src/context/RealmContext.tsx
ssilvert: @dlabaj I'm definitely against having one big global glob.  So that part sounds good to me.

But the realm list is only used in that one dropdown component.  I don't see why we need to create a context for it.  You can just put it in local state and let the component refresh the list whenever it needs it.

src/components/realm-selector/RealmSelector.tsx
edewit: I have generated these models based on the java classes on the backend

src/context/RealmContext.tsx
edewit: Yeah also against a big context with global state, I have done something similar for the help, the only thing that is needed for components down in the hierarchy is to know if help is enabled or not.

ssilvert:
@jenny-s51 I just ran your code.  A few problems:

- There is an infinite loop that keeps calling localhost:8080/admin/realms
- The list should be in alphabetical order
- Each realm name should be capitalized.

I don't think we have a wireframe for the ream list?  If we do, please post a link.  If we don't, you can use the old admin console as a general guide.

jenny-s51:
@ssilvert thank you for your helpful feedback! Sounds good - I'll make sure to address these points you mentioned. Here is a link to the mockups that include the realm selector ... I'll work on making it look more like this as well.

https://marvelapp.com/prototype/ei9eebh/screen/71477803

ssilvert (approved):

PR #60, opened 2020-09-03
remove env vars and use simple copy task to switch

edewit:
to switch to a `rh-sso` "version" of the console, you run `npx grunt switch-sso` and run `git checkout public` to switch back.


ssilvert (changes requested):
@edewit In the README, please add how to build RH-SSO.

ssilvert (approved):

PR #64, opened 2020-09-04
syntax highlighting on json import

edewit:
image


ssilvert:
@edewit I'm going to close this for now.  I have found out that there is a good chance we will need a full javascript editor anyway.  Since OpenShift has one I was told we should try to find out what package they are using.

edewit:
I know what they are using, they use https://microsoft.github.io/monaco-editor/ it's vs code. Do we really need it? It will make the console huge!

PR #65, opened 2020-09-04
Adjusts toolbar help menu items

srambach:
This adjusts a few things on the toolbar help menu and kebab menu. It fixes alignment and spacing of icons and moves to use the DropdownItem description attribute for the descriptive text. The color of the descriptive text is manually styled pending a bug fix in Patternfly.

Related to comments in #30 as well as other discussions.

### DIFF
--- a/src/PageHeader.tsx
+++ b/src/PageHeader.tsx
@@ -61,7 +61,7 @@ const ServerInfoDropdownItem = () => {
 const HelpDropdownItem = () => {
   const { t } = useTranslation();
   const help = t("Help");
-  return <DropdownItem icon={<HelpIcon />}>{`${Help}`}</DropdownItem>;
+  return <DropdownItem icon={<HelpIcon />}>{`${help}`}</DropdownItem>;
 };
 
 const kebabDropdownItems = [
--- a/src/PageHeader.tsx
+++ b/src/PageHeader.tsx
@@ -61,12 +61,7 @@ const ServerInfoDropdownItem = () => {
 const HelpDropdownItem = () => {
   const { t } = useTranslation();
   const help = t("Help");
-  return (
-    <DropdownItem>
-      <HelpIcon />
-      {` ${help}`}
-    </DropdownItem>
-  );
+  return <DropdownItem icon={<HelpIcon />}>{`${help}`}</DropdownItem>;
 };
 
 const kebabDropdownItems = [
--- a/src/PageHeader.tsx
+++ b/src/PageHeader.tsx
@@ -61,7 +61,7 @@ const ServerInfoDropdownItem = () => {
 const HelpDropdownItem = () => {
   const { t } = useTranslation();
   const help = t("Help");
-  return <DropdownItem icon={<HelpIcon />}>{`${help}`}</DropdownItem>;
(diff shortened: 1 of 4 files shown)


ssilvert (changes requested):
@srambach We should not be doing workarounds for PatternFly bugs.  Like we said earlier today, we want to eliminate custom css.

Please submit a bug report to PatternFly.  Then create an issue in keycloak-admin-ui and link to the bug report.

srambach:
The dropdown description text color will be fixed by https://github.com/patternfly/patternfly/issues/3411

@stianst do you still want me to create an issue, given that there is nothing to change on our side?

src/PageHeader.tsx
```diff
@@ -61,12 +61,7 @@ const ServerInfoDropdownItem = () => {
 const HelpDropdownItem = () => {
   const { t } = useTranslation();
   const help = t("Help");
-  return (
-    <DropdownItem>
-      <HelpIcon />
-      {` ${help}`}
-    </DropdownItem>
-  );
+  return <DropdownItem icon={<HelpIcon />}>{`${help}`}</DropdownItem>;
```
yordis, line 64:
```suggestion
  return <DropdownItem icon={<HelpIcon />}>{t("Help")}</DropdownItem>;
```

srambach: Good catch, thanks!

ssilvert:
> The dropdown description text color will be fixed by [patternfly/patternfly#3411](https://github.com/patternfly/patternfly/issues/3411)
> 
> @stianst do you still want me to create an issue, given that there is nothing to change on our side?
@srambach 

Thanks.  Yes.  Please create an issue so we will remember to follow up on it.  Granted, with the way it looks I don't think we will forget.

BTW, you messaged @stianst  instead of me.  Happens a lot.  At least you didn't call him Stain.

ssilvert:
@srambach This code breaks the kabab menu in the masthead.  Reduce viewport size then try to click kabab and it breaks.

src/PageHeader.tsx
srambach: @yordis, I misread this earlier - thought you were pointing out that the menu item name should be capitalized. I will defer to @ssilvert there because I don't know if it's better to store the text in a variable for future modification/customization maybe? Reverted for now, sorry!

edewit:
for me it looks like:
image
is that what we want until patternfly has a fix?

edewit (approved):

ssilvert:
> for me it looks like:
> image
> is that what we want until patternfly has a fix?

It won't be long before we have a fix.  Let's just leave it as-is for now.

@srambach Do you know why the kabab broke?

ssilvert:
@srambach Given that we are waiting on the PF fix, do we still need to merge this or should we close it?

srambach:
@ssilvert The kebab broke because I mistakenly changed the letter case on the help variable after the comment above. It's fixed now.
And yes, we should still merge this because I corrected alignment as well by including a layout element to right justify the icon and switch, as shown in the designs. This PR also moves the text to the description attribute as it will need to be to pick up the PatternFly fix, and removes unnecessary CSS.

ssilvert (approved):

PR #67, opened 2020-09-04
Use isBlock instead of css

ssilvert:


jenny-s51 (approved):
Good update! Tested and LGTM - CI is sad though

PR #69, opened 2020-09-05
Run tsc on build to check for typescript errors

ssilvert:
I requested all of you as reviewers because I wanted you to at least be aware that we are now checking for typescript errors.

For a couple of the stories it looked like there was more broken than just typescript errors.  So I renamed them to *.bak.  If you want to fix them then please submit a new PR for them.  If you don't want them any more please submit a PR to remove the *.bak files.


src/components/realm-selector/RealmSelector.tsx
```diff
@@ -28,7 +28,7 @@ export const RealmSelector = () => {
 
   useEffect(() => {
     getRealms().then((result) => {
-      setRealms(result) !== undefined ? result : [];
+      setRealms(result!);
```
ssilvert, line 31: @jenny-s51 I'm not sure exactly what this was for, but the realm list is guaranteed to at least have one item.  I changed it to make the typescript checker happy.

edewit: even better option would be to do `setRealms(result || [])`

package.json
```diff
@@ -8,6 +8,7 @@
     "start": "snowpack dev",
     "build": "snowpack build",
     "test": "jest",
+    "check-types": "tsc --noImplicitAny --strictNullChecks --jsx react -p ./",
```
edewit, line 11: I wouldn't add these flags here as they are already defined in the `tsconfig.json`

edewit (changes requested):

edewit (approved):

Issue #71, opened 2020-09-07
Realm selector layout is not centered


edewit:
![image](https://user-images.githubusercontent.com/51133/92377056-d3dea080-f103-11ea-9cdf-cea7e1d9c811.png)
Seems that the dropdown is to much to the right / not centered

PR #73, opened 2020-09-07
fixed size of dropdown menu

edewit:
fixes: #71

### DIFF
--- a/src/components/realm-selector/realm-selector.module.css
+++ b/src/components/realm-selector/realm-selector.module.css
@@ -8,13 +8,9 @@
   outline-offset:-32px;    
 }
 
-.menu {
-    /* setting this to 85% for now */
-    min-width: 85%;
-}
-
-.pf-c-button.pf-m-primary.realmSelectButton {
-  width: 100%;
+:global(#realm-select > .pf-c-dropdown__menu) {
+    /* setting this to 78% for now */
+    min-width: 78%;
 }
 
 .toggle {


ssilvert:
@srambach @edewit Given what we discussed in the meeting, is this PR still useful?

src/components/realm-selector/realm-selector.module.css
```diff
@@ -8,13 +8,9 @@
...
-.menu {
-    /* setting this to 85% for now */
-    min-width: 85%;
-}
-
-.pf-c-button.pf-m-primary.realmSelectButton {
-  width: 100%;
+:global(#realm-select > .pf-c-dropdown__menu) {
+    /* setting this to 78% for now */
+    min-width: 78%;
```
jenny-s51, line 13: 78% is a bit too small for my viewport ... did it look different for you when you tested? 85% worked best for me

Screen Shot 2020-09-08 at 1 22 37 PM

srambach: I would recommend a different approach. The dropdown should fill its space 100%. This might be done with a `.pf-m-fill` modifier (side note, it looks like we need to get patternfly-addons imported for utility and modifier classes) or by placing it within a layout like `<Split>` and then <SplitItem isFilled>` to make it span the width of the container. 
Then the space around it should be provided by its parent. If this is a different element and not contained within the navigation pane, then it should mirror that padding using the standard spacer variables.
There's also an odd space between the toggle and dropdown list coming from I think the outline-offset that went away when I took out some of the extra styles.
I'll paste in a picture of the inspector showing some of this, but I'm happy to have a call if that's easier.
image

edewit:
Your version looks a lot better indeed @srambach could you create a PR instead of this one?

srambach:
> Your version looks a lot better indeed @srambach could you create a PR instead of this one?

Yeah, sure!

PR #80, opened 2020-09-10
Added missing logic to realm form

edewit:


ssilvert (commented):
Per our conversation yesterday, let's use camel-case keys for the translations.  That way we always know that:

- If translations are broken we see it right away.
- We can be sure that all the translation keys are present in the translation files. (Makes it easier to translate to another language)

ssilvert (approved):

PR #87, opened 2020-09-11
Restyle realm selector and fix centering

### LINKED ISSUE #71: Realm selector layout is not centered

srambach:
Fixes #71 

This fixes the width and position of the realm selector in the left navigation.
NOTE: There are TODOs noting where class names must be corrected to our format after the CSS modules processor is removed. The processor does not support double dashes and the :global declaration is also acting wonky, which is why I have used only underlines as separators and left one as a local style.


edewit:
Removing the css modules is easy you only have to rename the file to not have the word `module` in it and change the import to `import realm-selector.css`

dlabaj (approved):
LGTM

PR #88, opened 2020-09-14
changed messages to keys

edewit:


src/components/help-enabler/__tests__/__snapshots__/HelpHeader.test.tsx.snap
```diff
@@ -53,7 +53,7 @@ exports[`<HelpHeader /> enable help 1`] = `
             <div
               class="pf-l-split__item pf-m-fill"
             >
-              Documentation
+              documentation
```
ssilvert, line 56: @edewit So this is what I was talking about the other day.  I'm thinking it's probably not best to have un-translated strings in the snapshots.  We do want to be able to catch if something failed in the translation.

However, I'm not sure what it takes to do that?  If it requires changes to all the components then I wouldn't want to do it.  For instance, I wouldn't want to be forced to always use "withTranslation" instead of "useTranslation".

But if i18next can be set up to "just work" in jest then that would be good.  WDYT?

ssilvert (approved):

src/components/help-enabler/__tests__/__snapshots__/HelpHeader.test.tsx.snap
edewit: Right now I have it mocked, the `withTranslation` was just to specify a default namespace. I can have a look to see if I can make it work in jest as well... I think it's doable

PR #90, opened 2020-09-14
updated the realm selector

edewit:
1. realm context that works on other requests
2. added search bar when realm list > 5 realms


src/clients/import/ImportForm.tsx
```diff
@@ -49,7 +51,7 @@ export const ImportForm = () => {
 
   const save = async () => {
     try {
-      await httpClient.doPost("/admin/realms/master/clients", client);
+      await httpClient.doPost(`/admin/realms/${realm}/clients`, client);
```
yordis, line 54: Wouldn't be better to generate these things with some Codegen? I know that keycloak is trying to add Swagger, but even before getting that you can prepare yourself to remove the manual generated SDK.

And also abstract the HTTP/Transport layer out of the UI.

```suggestion
      await sdk.getClients(realm);
```

src/components/realm-selector/RealmSelector.tsx
```diff
@@ -1,62 +1,120 @@
...
+
+  const toUpperCase = (realmName: string) =>
+    realmName.charAt(0).toUpperCase() + realmName.slice(1);
+
+  const RealmText = ({ value }: { value: string }) => (
+    <Split>
+      <SplitItem isFilled>{toUpperCase(value)}</SplitItem>
+      <SplitItem>{value === realm && <CheckIcon />}</SplitItem>
+    </Split>
+  );
```
yordis, line 40: Wouldn't be simpler and cleaner to avoid deconstruction in TypeScript land?

The datatype already defines the structure of the data, and deconstruction habits create larger code base and harder to use refactoring tools from IDEs since they require runtime execution for rename and so on, but static information like datatypes don't.

🤷‍♂️

```suggestion
  const RealmText = (props: { value: string }) => (
    <Split>
      <SplitItem isFilled>{toUpperCase(props.value)}</SplitItem>
      <SplitItem>{props.value === realm && <CheckIcon />}</SplitItem>
    </Split>
  );
```

edewit: I don't think that is true, when deconstructing the properties they will still be checked in the example above if I would to type `({ something }: TheType)` the check will be if something is defined in `TheType`

src/clients/import/ImportForm.tsx
edewit: Right there is no Swagger yet, so in the mean time we use http, I don't see a huge benefit of trying to abstract that.

src/components/realm-selector/RealmSelector.tsx
yordis: Let's agree to disagree.

Personally, such practices end up doubling the codebase size in TypeScript land (deconstruction + type spec everywhere), and make it harder to refactor in the future, but I guess we have different experiences.

🤷‍♂️ 

## Deconstructing

Sep-15-2020 02-14-12

## Non-deconstructing

Sep-15-2020 02-14-29

edewit: Are you using the argument that because it's a newer feature we should not use it? By that logic we should stick to angular.

edewit: In the your screenshots you are refactoring the name of the property in the first one and the other the name of the variable, I don't see why we would not use deconstructing just because of that.

ssilvert: @yordis If you have a problem with destructuring in TypeScript then the place to discuss that is over at the TypeScript project.

ssilvert:
@edewit Please let me know when you fix the conflicts on this one.

src/components/realm-selector/RealmSelector.tsx
yordis: Has nothing to do with TypeScript itself (since deconstruction for the sake of, still an over abused feature that meant to be use for something else) I think you misinterpreted the feedback.

Regardless, no point in having a discussion about it, it is your project and your guidelines, deconstructing will be 🤷

ssilvert:
@edewit The "x" button doesn't work.  If I do a search and narrow the list I can not get rid of the search without manually emptying the search box and hitting the search button again.

edewit:
@ssilvert there are no options for that, it's standard patternfly stuff and it works on my machine.
context

ssilvert:
> @ssilvert there are no options for that, it's standard patternfly stuff and it works on my machine.
That's not how I expected it to work, but it's OK for now.  I see that the component is listed as "beta" anyway.

ssilvert (approved):

PR #95, opened 2020-09-16
Empty state for a list

### LINKED ISSUE #35: General layouts - add empty state to storybook
https://marvelapp.com/prototype/504f2fh/screen/72066305

srambach:
This adds an opinionated empty state to be used for a list. Based on mockup https://marvelapp.com/prototype/504f2fh/screen/72066304

Can be viewed in the storybook.

Fixes #35


jenny-s51 (commented):
Code, styling, and storybook demo are looking good! I tested this in the groups section and was able to integrate the new component and customize the prop inputs without any issues. My only quibble here is the suggestion to rename the component to `EmptyListState` so it's a bit clearer that it's an empty state component. Otherwise LGTM :)

src/components/empty-list/empty-list.tsx
```diff
@@ -0,0 +1,38 @@
...
+  Title,
+  Button,
+} from "@patternfly/react-core";
+import { PlusCircleIcon } from "@patternfly/react-icons";
+
+type EmptyListProps = {
+  message: string;
+  instructions: string;
+  primaryActionText: string;
+  primaryAction: any;
```
ssilvert, line 15: We want to avoid using `any`
```suggestion
  primaryAction: (event: React.MouseEvent<HTMLButtonElement>) => void;
```

ssilvert (commented):
We have adopted CamelCase convention for ts and tsx files.

`empty-list.tsx` file should be named `EmptyList.tsx`

src/components/empty-list/empty-list.tsx
edewit: Also I would recommend to name actions "on(action name)" so in this case `onPrimaryAction` that way you can also try them out in storybook
```suggestion
  onPrimaryAction: MouseEventHandler<HTMLButtonElement>;
```

edewit (changes requested):
A cool new feature of this new storybook is to add meta data so you can play with the attributes of a component like this:
```javascript
import React from "react";
import { Meta, Story } from "@storybook/react";
import { Page } from "@patternfly/react-core";
import { EmptyList, EmptyListProps } from "../components/empty-list/empty-list";

export default {
  title: "Empty list",
  component: EmptyList,
} as Meta;

const Template: Story<EmptyListProps> = (args) => (
  <Page>
  <EmptyList {...args} />
  </Page>
);

export const View = Template.bind({});
View.args = {
  message:"No things",
  instructions:"You haven't created any things for this list.",
  primaryActionText:"Add a thing"
};
```
storybook

edewit:
also rename the file `EmptyList.tsx`

ssilvert:
> A cool new feature of this new storybook is to add meta data so you can play with the attributes of a component like this:

That is cool indeed.  Gives a good reason for using storybook.

srambach:
Thank you @jenny-s51, @edewit, and @ssilvert for your helpful comments! I've updated everything and added a test, which is new to me so please let me know if that's ok or not.

ssilvert (approved):

jenny-s51 (approved):
LGTM 🚀

PR #100, opened 2020-09-17
Now uses http to load translations and bundles in tests

edewit:


src/__mocks__/react-i18next.js
```diff
@@ -0,0 +1,64 @@
...
+  });
+};
+
+import common from "../../public/locales/en/common.json";
+import clients from "../../public/locales/en/clients.json";
+import realm from "../../public/locales/en/realm.json";
+import groups from "../../public/locales/en/groups.json";
+import help from "../../public/locales/en/help.json";
+
+const translation = { ...common, ...help, ...clients, ...realm, ...groups };
```
ssilvert, line 43: So if I understand, all components running inside tests will now be able to access `useTranlation` without any additional setup except to add new bundles here?

edewit: yes that is right

ssilvert: @edewit Perfect.  

Looks like we still have test failures.  Please fix so we can get this merged.

edewit:
@ssilvert already tried to make the test not fail for the last 4 commits of this PR

ssilvert:
> @ssilvert already tried to make the test not fail for the last 4 commits of this PR

OK.  Do you want to try some more or just close this PR?  We can come back and do this later if you want.

PR #104, opened 2020-09-18
Small client form fixes

edewit:


ssilvert (commented):
Just noticed a misspelling.  "Implicid flow" should be "Implicit flow".

Checkboxes still look too far apart.  Wonder what would happen if you use `span={1] hasGutter="true"`.

I bet there is a way to get the 24px spacing that is called for in the design without resorting to custom css.  Maybe @srambach has an idea.

src/clients/add/Step2.tsx
```diff
@@ -52,7 +52,7 @@ export const Step2 = ({ form }: Step2Props) => {
       </FormGroup>
       <FormGroup label={t("authenticationFlow")} fieldId="kc-flow">
         <Grid>
-          <GridItem span={6}>
+          <GridItem lg={4} sm={6}>
```
srambach, line 55: Instead of marking each of the `GridItem`s you can use something like `<Grid sm={12} lg={6} xl={3}>`
The vertical alignment is off on these forms and I'm not sure why. I can put it on my list to investigate.

edewit: the idea was to bring them more together

src/clients/ClientDescription.tsx
```diff
@@ -1,21 +1,26 @@
...
   const { t } = useTranslation("clients");
+  const { register, errors } = form;
   return (
     <>
-      <FormGroup label={t("clientID")} fieldId="kc-client-id">
+      <FormGroup
+        label={t("clientID")}
+        fieldId="kc-client-id"
+        helperTextInvalid={t("common:required")}
+        validated={errors.clientId ? "error" : "default"}
```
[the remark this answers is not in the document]
edewit: there no constants as far as I know, but they are type checked so you can't just put any string

src/clients/add/Step2.tsx
ssilvert: @srambach Please create an issue for this and put the issue number in a comment here.  It looks like this is the only thing holding up this PR, so I'd like to merge and let you investigate out of band.  We will merge once we have your issue.

srambach: Issue created https://github.com/keycloak/keycloak-admin-ui/issues/113

ssilvert (approved):

PR #107, opened 2020-09-21
initial version of the multi line input

edewit:


ssilvert (changes requested):
This doesn't let you remove the first item.  See behavior of old console.

srambach:
There are problems with keyboard navigation also. Is there a UX design for this or just from the old console?

srambach:
The reason I ask about a UX design is that there are many novel interaction issues here to work out - tab order, arrow keys, enter key, labeling, and placeholders are just the ones off the top of my head. Good keyboard controls will be essential not just for a11y but for "power" users.

ssilvert:
> Is there a UX design for this or just from the old console?

@srambach The UX design just shows this component and it looks like the one in the old console.  As far as I know, there is no PatternFly component for this and I don't know what we would use in its place.

ssilvert (approved):

ssilvert:
@edewit I approved this PR, but we should address @srambach concerns about tab order and enter key at some point.  So I'll leave it up to you if you want to fix it now or log an issue and merge.

edewit:
@srambach I did address the tab order, but if you want some more things I'll be happy to add them (just don't know what arrow keys means ;) )

srambach:
@edewit This needs to have some input from interaction design. I'll take that as a task and get it spelled out. https://github.com/keycloak/keycloak-admin-ui/issues/121

PR #108, opened 2020-09-21
First page within Group section

christiemolloy:
Implemented first page in the Groups Section.

**Notes:**
-  I haven't added actual data yet so I had to add a new component called GroupsSectionData which passes in the data as a prop, that was the only way I could immediately set the data. I plan to add the data in the next round of changes.
- I updated TableToolbar component so that you can pass in the input group now instead of having one there... this is because all the pages require different input groups. I also updated client section to reflect this change.
- There's quite a few TS errors - I plan to fix those once I actully add the data because I haven't looked at the models yet from the old Keycloak code - so that will probably fix all of the errors.
- Commented out the DataLoader component for now as it was refreshing all the current states on the page on every re-render so I need to investigate that further.

**Functionality to test:**
1) Filter on groups
2) Select items in the table
3) Select multiple items in the table and click on Toolbar kebab > "Delete" to delete rows in the table.
4) Select Table kebab > "Delete" to delete individual row in the table.


ssilvert (changes requested):
@christiemolloy Nice to see a first PR.  A few things to get this going:
- Please fix conflicts
- You will need to clean up the TS errors before it will let you merge
- We are using camelCase for resource bundle keys.
i.e. `"Create group": "Create group"` should be `"createGroup": "Create group"`

src/groups/GroupsList.tsx
```diff
@@ -3,31 +3,43 @@ import {
...
-  });
+  const tableHeader = [ { title: t("Group name") }, { title: t("Members") }];
+  
+  const actions = [
+    {
+      title: t("Move to"),
+      onClick: () => console.log('TO DO: Add move to functionality')
+    },
+    {
+      title: t("Delete"),
```
dlabaj, line 29: Can we call this something other than t.  t has no meaning when you look at it without looking the useTranslate above.

src/groups/GroupsSection.tsx
```diff
@@ -1,53 +1,219 @@
...
+  PageSectionVariants,
+  Title,
+  TitleSizes,
+  ToolbarItem
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+import { UsersIcon } from '@patternfly/react-icons';
+import './GroupsSection.css';
+
+type GroupsSectionProps = {
```
dlabaj, line 27: Shouldn't this be an interface instead of a type?

src/groups/GroupsSection.tsx
```diff
@@ -1,53 +1,219 @@
...
+        if (groupName.toLowerCase().includes(newInput.toLowerCase())) {
+          localRowData.push(obj);
+      }
+    })
+    setFilteredData(localRowData);
+  };
+
+  function onSelect(_, isSelected: boolean, rowId: number) {
+    let localRow;
+    if (rowId === -1) {
```
dlabaj, line 109: Can we use a constant instead of -1

src/groups/GroupsSection.tsx
```diff
@@ -1,53 +1,219 @@
...
-      <DataLoader loader={loader}>
-        {(groups) => (
+    <React.Fragment>
+      <PageSection variant={PageSectionVariants.light}>
+        <Title headingLevel="h3" size={TitleSizes['2xl']}>
+          {t("Groups")}
+        </Title>
+      </PageSection>
+      <Divider/>
+      {/* <DataLoader loader={loader}>
```
dlabaj, line 165: Can we remove the commented out code.

src/groups/GroupsSection.tsx
```diff
@@ -1,53 +1,219 @@
...
+      {(groups) => ( */}
           <TableToolbar
-            count={groups!.length}
+            count={10}
             first={first}
             max={max}
             onNextClick={setFirst}
             onPreviousClick={setFirst}
-            onPerPageSelect={(f, m) => {
+            onPerPageSelect={(f,m) => {
```
dlabaj, line 173: what's f and m.  We should use full names instead of a single letter.

src/groups/GroupsSection.tsx
```diff
@@ -1,53 +1,219 @@
...
+            }
           >
-            <GroupsList list={groups} />
+            <GroupsList list={formattedData} onSelect={onSelect} onDelete={onDelete} />
           </TableToolbar>
-        )}
-      </DataLoader>
-    </>
+      {/* )}
+      </DataLoader> */}
```
dlabaj, line 216: Remove dead code.

dlabaj (changes requested):

src/groups/GroupsList.tsx
ssilvert: @dlabaj That a convention commonly used with the i18next package.

src/groups/GroupsSection.tsx
ssilvert: I've done some reading on this, and while you could use either, `type` is probably a better choice here.  `type` is a little more flexible.  In our codebase, we are only using interfaces for data transfer definitions (representations).

src/groups/GroupsSection.css
```diff
@@ -0,0 +1,10 @@
+.pf-icon-group-members {
+  display: flex;
+  align-items: center;
+}
+
+.pf-icon-group-members svg {
+  color: var(--pf-global--Color--200);
+  font-size: var(--pf-global--FontSize--md);
+  margin-right: var(--pf-global--spacer--sm);
+}
```
srambach, line 10: We decided to prefix all CSS classes using the pattern `.keycloak-admin--block[__element][--modifier][--state][--breakpoint][--pseudo-element]` so maybe instead of `.pf-icon-group-members`, something more like `.keycloak-admin--groups__member-count`? 
Since this is used twice already, is it worth making that a component? Especially if there are going to be variations in color or icon. I can't find where it is in the Groups mockups so IDK if there are or not?

src/groups/GroupsList.tsx
dlabaj: We can leave this since that's how it's being provided by the library.

src/groups/GroupsSection.tsx
dlabaj: That's opposite of the way it is on other projects.  What's the convention then we use type for all Props?

ssilvert: Yes, I think we should use `type`.  The main place it might come in handy for props is the ability to do unions, which you can't do with interfaces.

BTW, I'm getting my info from O'Reily book "Effective TypeScript".  Really good read.

We use interfaces for the representations because the tool we use to generate them from Java code uses interfaces.  See Realm.ts for an example.

src/clients/ClientsSection.tsx
```diff
@@ -41,6 +42,16 @@ export const ClientsSection = () => {
               setFirst(f);
               setMax(m);
             }}
+            inputGroup={
```
edewit, line 45: I think there might be a way to make this a bit more re-usable? I mean now every user of this will need to create this same piece of components just to make the search work. Maybe we can have the field and button still be in the `TableToolbar` and only expose the "action"?

edewit (changes requested):

src/groups/GroupsSection.tsx
dlabaj: Thanks stan

src/groups/GroupsSection.tsx
christiemolloy: I got this straight from PatternFly. I can update it though to be undefined

src/groups/GroupsSection.tsx
christiemolloy: I think these were from @edewit 's implementation. Erik would you be ok with changing these to full names?

src/groups/GroupsSection.css
christiemolloy: I refactored the code so that class only appears once now. I don't think we need it as a separate component unless another comes up! which it might, so I can add that in the future. I updated the naming of the class thanks for that feedback!

src/clients/ClientsSection.tsx
christiemolloy: Agreed. I updated it so that now the Input Group and Button are in the TableToolbar, and you pass in the name, placeholder, and onChange event

src/groups/GroupsSection.tsx
```diff
@@ -1,55 +1,135 @@
...
+  const columnID: (keyof GroupRepresentation) = "id";
+  const membersLength: (keyof GroupRepresentation) = "membersLength";
+  const columnGroupName: (keyof GroupRepresentation) = "name";
 
   const loader = async () => {
-    return await httpClient
-      .doGet("/admin/realms/master/groups", { params: { first, max } })
-      .then((r) => r.data as GroupRepresentation[]);
+    const groups = await httpClient.doGet("/admin/realms/master/groups", { params: { first, max } });
+    const groupsData = groups.data;
```
christiemolloy, line 37: @edewit do you mind taking a look at the TS errors that I'm getting in this block. The `<HTTPResponse>` is unknown and when I try to map it its throwing an error because its unknown. I've been reading up on async/await functions and TS  https://www.carlrippon.com/type-safe-data-fetching-with-unknown/  but I'm still struggling so wondering if you could help me out :)

edewit: right add the type on the `doGet` like this:
```javascript
const groups = await httpClient.doGet<GroupRepresentation[]>("/admin/realms/master/groups", { params: { first, max } });
```

christiemolloy:
@ssilvert @edewit @dlabaj @srambach @jenny-s51  checks finally all passed if you want to take another look :)

src/groups/GroupsSection.tsx
edewit: I think we shouldn't create constants for something like `-1` see: https://www.codereadability.com/magic-numbers/

src/groups/GroupsSection.tsx
edewit: I'll do that in another PR we can rename them to `first` and `max` still pretty short

edewit (approved):

edewit:
I get an error in the console about the inline property:
```
Warning: React does not recognize the `isInline` prop on a DOM element.
```
Other then that :+1: looks great

edewit:
Oh one more comment, this is more general and not specific to your implementation but fetching the members of a group is limited to 100 @ssilvert should we remove the column? I guess without it the whole thing also looks dumb

src/groups/GroupsSection.tsx
dlabaj: okay i've been on projects where I seen it used a constant instead of -1.  We can leave it then.  Thanks.

christiemolloy:
> Oh one more comment, this is more general and not specific to your implementation but fetching the members of a group is limited to 100 @ssilvert should we remove the column? I guess without it the whole thing also looks dumb

Could we limit the number of groups using pagination and then on every new page fetch the members by all the groupIDs on the page?

dlabaj (approved):
LGTM

ssilvert (changes requested):
@christiemolloy It looks like you ran prettier against some files that you didn't actually change?

tsconfig.json
```diff
@@ -1,29 +1,29 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "esnext",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
-    "module": "esnext",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
+    "module": "esnext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
```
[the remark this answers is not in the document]
christiemolloy: @ssilvert yep thats because I ran `prettier --write .` I probably shouldn't have committed those files too, so I can revert those changes, and next time just make sure that I only use `prettier` for the files that have formatting errors, do you know if there is a short cut using prettier for formatting certain files ?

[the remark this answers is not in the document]
ssilvert: @christiemolloy I set up VSCode to "format on save", with prettier as my default formatter.  That way, it only tries to format the files I've worked on.

I'm a little distressed about what prettier did to tsconfig.json.  That's a nasty-looking change.

christiemolloy:
@ssilvert do you mind taking a look again? some of the files didnt update but its only slight formatting and i fixed tsconfig.json

ssilvert:
@srambach Can you look at the changes you requested on this one and approve or disapprove ASAP?  This one is at the 7 day limit.

src/groups/GroupsList.tsx
```diff
@@ -1,35 +1,106 @@
...
+    // TO DO update the state
+  }
 
-  const data = list.map((c) => {
-    return { cells: columns.map((col) => c[col]) };
-  });
+  const tableHeader = [{ title: t("groupName") }, { title: t("members") }];
+  const actions = [
+    {
+      title: t("moveTo"),
```
ssilvert, line 79: Missing key.  Please fix ASAP.

src/groups/GroupsList.tsx
```diff
@@ -1,35 +1,106 @@
...
+      onClick: () => console.log("TO DO: Add move to functionality"),
+    },
+    {
+      title: t("delete"),
```
ssilvert, line 83: Missing key.  Should be `"common:delete"`.  Please fix ASAP.

ssilvert (changes requested):

src/clients/ClientsSection.tsx
```diff
@@ -47,6 +47,8 @@ export const ClientsSection = () => {
                 setFirst(first);
                 setMax(max);
               }}
+              inputGroupName="clientsToolbarTextInput"
+              inputGroupPlaceholder="Search for client"
```
srambach, line 51: Does this need a translation?

src/groups/GroupsList.tsx
```diff
@@ -1,35 +1,106 @@
...
-      <TableBody />
-    </Table>
+    <React.Fragment>
+      {formattedData && (
+        <Table
+          actions={actions}
+          variant={TableVariant.compact}
+          onSelect={onSelect}
+          canSelectAll={false}
+          aria-label="Selectable Table"
```
srambach, line 96: Does this need a translation?

src/groups/GroupsList.tsx
christiemolloy: There is a "moveTo" key that exists in the messages file. Are you referring to something else?

christiemolloy:
@srambach @ssilvert updated

ssilvert (approved):

PR #115, opened 2020-09-23
Generic view header

edewit:
Can be used as a header for a from or view as well as a specific item with actions


src/realm/messages.json
```diff
@@ -5,6 +5,7 @@
     "enabled":"Enabled",
     "create":"Create",
     "createRealm": "Create realm",
+    "realmExplain": "A realm is something",
```
ssilvert, line 8: You can use this text: `A realm manages a set of users, credentials, roles, and groups. A user belongs to and logs into a realm. Realms are isolated from one another and can only manage and authenticate the users that they control.`

ssilvert (changes requested):

src/clients/ClientsSection.tsx
```diff
@@ -28,40 +29,46 @@ export const ClientsSection = () => {
...
+      <PageSection variant="light">
+        <DataLoader loader={loader}>
+          {(clients) => (
+            <TableToolbar
+              count={clients!.length}
+              first={first}
+              max={max}
+              onNextClick={setFirst}
+              onPreviousClick={setFirst}
+              onPerPageSelect={(f, m) => {
```
dlabaj, line 46: Change to something other then f and m

dlabaj (commented):
One minor comment otherwise looks great.

dlabaj (approved):
LGTM

ssilvert (approved):

PR #117, opened 2020-09-23
initial version of confirm dialog

edewit:


srambach:
Should the first two stories for this display the modal?
By default, as a confirmation modal, I think this should use a primary button, and a variant added to make it a danger button.
I think we should also make the modal small by default, and allow a variant for the other sizes if needed. (unless you have mockups saying otherwise?)

src/components/confirm-dialog/ConfirmDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+          }}
+        >
+          {t(continueButtonLabel || "common:continue")}
+        </Button>,
+        <Button
+          id="modal-cancel"
+          key="cancel"
+          variant={ButtonVariant.secondary}
+          onClick={() => {
+            if (onCancel) onCancel();
```
ssilvert, line 76: If there is no cancel handler, we shouldn't snow the cancel button.

ssilvert (changes requested):

ssilvert:
> Should the first two stories for this display the modal?
> By default, as a confirmation modal, I think this should use a primary button, and a variant added to make it a danger button.
> I think we should also make the modal small by default, and allow a variant for the other sizes if needed. (unless you have mockups saying otherwise?)

Agree with @srambach.  Otherwise, LGTM.

dlabaj (approved):
LGTM other then sarah's comment

edewit:
@srambach ahh yes it should show the dialog, fixed that now

PR #119, opened 2020-09-25
initial version of the breadcrumb component

edewit:


ssilvert:
@edewit Please resolve conflict.

src/route-config.ts
```diff
@@ -0,0 +1,111 @@
...
+  },
+  {
+    path: "/realm-roles",
+    component: RealmRolesSection,
+    breadcrumb: t("roles:roleList"),
+  },
+  {
+    path: "/users",
+    component: UsersSection,
+    breadcrumb: "Users",
```
ssilvert, line 64: Please go ahead and localize these.

ssilvert (changes requested):
@edewit This looks good.  Are there any screens where I can see it in action?

srambach:
Making a component `Breadcrumbs` (plural) seems to ask for mistakes since there's already the PatternFly component `Breadcrumb` - could it be something like `PageBreadcrumb` instead?

package.json
```diff
@@ -27,7 +27,8 @@
     "react-dom": "^16.8.5",
     "react-hook-form": "^6.8.2",
     "react-i18next": "^11.7.0",
-    "react-router-dom": "^5.2.0"
+    "react-router-dom": "^5.2.0",
+    "use-react-router-breadcrumbs": "^1.0.4"
```
dlabaj, line 31: We are adding a new library?  Do we really need this?

src/components/bread-crumb/__tests__/__snapshots__/BreadCrumbs.test.tsx.snap
```diff
@@ -0,0 +1,92 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`BreadCrumbs tests couple of crumbs 1`] = `
+<BreadCrumbs>
```
dlabaj, line 4: Let's call this something else it's too close to Breadcrumb and can cause coding errors.

dlabaj (changes requested):

package.json
edewit: we don't we can copy the code this is all of the code: https://github.com/icd2k3/use-react-router-breadcrumbs/blob/master/src/index.tsx

edewit:
@srambach I made it `PageBreadcrumbs` because it renders contains multiple crumbs IMO

package.json
ssilvert: @dlabaj @edewit I'm not crazy about adding more libraries either.  But I think copying the code is even worse.  Let's just use the library.  Then at least we don't have to maintain the code.

ssilvert (changes requested):
I can see it in action now.  Looks good except that there needs to be spacing between the breadcrumbs and the page title.

ssilvert:
@edewit Oh, and also a conflict.

package.json
dlabaj: agreed I don't want to copy the code.  Was just wondering if we really needed it or not.  It's small so we can bring it in.

dlabaj (approved):

edewit:
> I can see it in action now. Looks good except that there needs to be spacing between the breadcrumbs and the page title.

@srambach I guess just adding some padding would is not the way to go for this?

srambach:
> > I can see it in action now. Looks good except that there needs to be spacing between the breadcrumbs and the page title.
> 
> @srambach I guess just adding some padding would is not the way to go for this?

Probably not ;-) Can the breadcrumb be passed to the Page? That seems to be the way it's meant to be built now. (I think there used to be a page section variant for breadcrumbs but I don't see it now) https://www.patternfly.org/v4/components/page/react-demos

ssilvert (approved):

PR #125, opened 2020-09-29
Added client setttings header with actions

edewit:


src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+              <TextInput
+                type="text"
+                id="kc-home-url"
+                name="baseUrl"
+                ref={form.register}
+              />
+            </FormGroup>
+          </Form>
+          <Form isHorizontal>
+            <FormGroup label={t("consentRequired")} fieldId="kc-consent">
```
srambach, line 180: Please add `hasNoPaddingTop` to the form group to align the switch

src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+                    id="kc-consent"
+                    label={t("common:on")}
+                    labelOff={t("common:off")}
+                    isChecked={value}
+                    onChange={onChange}
+                  />
+                )}
+              />
+            </FormGroup>
+            <FormGroup
```
srambach, line 196: Please add `hasNoPaddingTop` to the form group to align the switch

src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+                  } else if (value === "delete") {
+                    toggleDeleteDialog();
+                  }
+                }}
+              />
+            </>
+          );
+        }}
+      />
+      <PageSection>
```
srambach, line 144: I think this needs a `variant="light"`

src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+            <>
+              <DisableConfirm />
+              <ViewHeader
+                titleKey="clients:clientList"
+                subKey="clients:clientsExplain"
+                selectItems={[
+                  <SelectOption
+                    key="action"
+                    value={t("common:action")}
+                    isPlaceholder={true}
```
ssilvert, line 114: `Action` should not show as a selectable value in the dropdown.

src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+                    toggleDeleteDialog();
+                  }
+                }}
+              />
+            </>
+          );
+        }}
+      />
+      <PageSection>
+        <Alerts />
```
ssilvert, line 145: Not something to address in this PR, but I don't understand why we need to keep adding the Alerts component over and over.  I would think that we should have only one instance of the Alerts component that all screens could add alerts to.

ssilvert (changes requested):
When you go to Client Settings page, the header still says `Client list`

src/clients/ClientSettings.tsx
edewit: That is the way it is with patternfly even if you give add the `isPlaceholder` it add it to the dropdown. @srambach do you know how you can get a placeholder to not show up in the dropdown?

src/clients/ClientSettings.tsx
edewit: I don't want to have a static component that we add like in the other console.

edewit: created issue #132 for this.

src/clients/ClientSettings.tsx
edewit: ahhh... I see it's resolved in 37c1576

ssilvert:
@srambach If you agree that all your issues are resolved then please approve.

src/clients/ClientSettings.tsx
```diff
@@ -65,95 +85,152 @@ export const ClientSettings = () => {
...
+            onConfirm: () => {
+              onChange(!value);
+              save();
+            },
+          });
+          return (
+            <>
+              <DisableConfirm />
+              <ViewHeader
+                titleKey="clients:clientList"
```
srambach, line 108: Is this still the wrong page title?  
Is this the mockup, and it should be the client name and have breadcrumbs?

edewit: Changed to use the client name and the breadcrumbs are done in #119 and yes that is the mockup

srambach: I'm just seeing a blank now, even though there's a clientID shown in the list...?

edewit: strange it seems to works sometimes... now I really fixed it ;)

ssilvert (approved):

PR #133, opened 2020-09-30
added client list search

edewit:


ssilvert:
@edewit This searches on every keystroke, which is probably fine.  But every time I press a key it sends the request twice.

ssilvert:
@edewit Actually, as I think about this, we need to ask about it on the dev list.  Even for realms I'm not 100% sure it's OK to do keystroke searches.  We should ask and be sure.

ssilvert:
@edewit I think the story on realms is that today keystroke searches will be fine.  But in the future with admin console running in a SAAS environment you are more likely to have a lot more realms and keystroke searches would be tool slow.  Again, we need to ask about this.

edewit:
@ssilvert right the old console also doesn't search on every keystroke, let's change it that the user has to press the search button. There are other fancy options like using a some time after some keys have been pressed, but I have bad experiences with that.

ssilvert (approved):

PR #134, opened 2020-09-30
Some pages don't support pagination

edewit:
so introduced non paginating toolbar that can be used for pages like groups

### DIFF
--- a/src/client-scopes/ClientScopesSection.tsx
+++ b/src/client-scopes/ClientScopesSection.tsx
@@ -46,7 +46,7 @@ export const ClientScopesSection = () => {
       />
       <PageSection variant="light">
         {!rawData && (
-          <div style={{ textAlign: "center" }}>
+          <div className="pf-u-text-align-center">
             <Spinner />
           </div>
         )}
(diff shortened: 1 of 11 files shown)


ssilvert:
@edewit Please resolve conflicts

srambach:
Looks good. One axe issue noting a duplicate aria-label, but this seems to fall under an exception for pagination controls existing above and below the table. (copying below for info) 

> This section of the WAI Aria Practices actually covers that specific instance under Step 3: Label areas:
> If a specific landmark role is used more than once on a page, provide each instance of that landmark with a unique label. There is one rare circumstance where providing the same label to multiple instances of a landmark can be beneficial: the content and purpose of each instance is identical. For example, a large search results table has two sets of identical pagination controls -- one above and one below the table, so each set is in a navigation region labelled Search Results. In this case, adding extra information to the label that distinguishes the two instances may be more distracting than helpful.

src/client-scopes/ClientScopesSection.tsx
```diff
@@ -32,28 +45,25 @@ export const ClientScopesSection = () => {
...
-                  {t("createClientScope")}
-                </Button>
-              }
-            >
-              <ClientScopeList clientScopes={scopes.data} />
-            </TableToolbar>
-          )}
-        </DataLoader>
+        {!rawData && (
+          <div style={{ textAlign: "center" }}>
```
ssilvert, line 49: You should be able to use PF layout components here instead of css.

ssilvert (changes requested):

ssilvert:
@christiemolloy Can you review this, please?

ssilvert (approved):

PR #136, opened 2020-09-30
Create whoami context

ssilvert:
WhoAmIContext provides a WhoAmI object from the server.  This provides things such as the proper display name of the user, whether the user is allowed to create new realms, and the realm that the user signed into.  It also sets the locale.

A subsequent PR will provide an AccessContext.  This will use the RealmAccess data from WhoAmI to provide permission checks.  This can not be done inside WhoAmIContext because we need to know the current realm and the RealmContext is not available yet (RealmContext uses home realm as initial realm, so we have a chicken and egg problem).

Eventually, App.tsx will look like this:
```
<Router>
      <WhoAmIContextProvider>
        <RealmContextProvider>
            <AccessProvider>
                 {children}
            </AccessProvider>
         </RealmContextProvider>
     </WhoAmIContextProvider>
</Router>
```

### DIFF
--- a/src/whoami/__tests__/WhoAmI.test.tsx
+++ b/src/whoami/__tests__/WhoAmI.test.tsx
@@ -1,0 +1,29 @@
+import whoamiMock from "./mock-whoami.json";
+import { WhoAmI } from "../WhoAmI";
+
+test("returns display name", () => {
+  const whoami = new WhoAmI("master", whoamiMock);
+  expect(whoami.getDisplayName()).toEqual("Stan Silvert");
+});
+
+test("returns correct home realm", () => {
+  let whoami = new WhoAmI("myrealm", whoamiMock);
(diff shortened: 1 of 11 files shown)


src/whoami/__tests__/WhoAmI.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+});
+
+test("can not create realm", () => {
+  const whoami = new WhoAmI("master", whoamiMock);
+  expect(whoami.canCreateRealm()).toEqual(false);
+});
+
+test("getRealmAccess", () => {
+  const whoami = new WhoAmI("master", whoamiMock);
+  console.log(JSON.stringify(whoamiMock));
```
edewit, line 23: guess this is left over from testing?

edewit (commented):
wouldn't we want to put this into the `components` subfolder?

edewit (approved):

Issue #131, opened 2020-09-29
Create realm role bugs


ssilvert:
- Form allows you to create a realm role without a name.
- Description field always displays error message `Max length 255`

jenny-s51:
Closing as these were resolved in #160

PR #141, opened 2020-10-01
add logic to delete realm role

### LINKED ISSUE #135: add logic to delete realm role

jenny-s51:
towards #94 . closes #135 and #131

### DIFF
--- a/src/realm-roles/RealmRolesSection.tsx
+++ b/src/realm-roles/RealmRolesSection.tsx
@@ -1,46 +1,78 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { useHistory } from "react-router-dom";
 import { useTranslation } from "react-i18next";
-import { Button, PageSection } from "@patternfly/react-core";
+import {
+  Button,
+  Divider,
+  PageSection,
+  Spinner,
+} from "@patternfly/react-core";
 
-import { DataLoader } from "../components/data-loader/DataLoader";
-import { TableToolbar } from "../components/table-toolbar/TableToolbar";
 import { HttpClientContext } from "../context/http-service/HttpClientContext";
 import { RoleRepresentation } from "../model/role-model";
 import { RolesList } from "./RoleList";
 import { RealmContext } from "../context/realm-context/RealmContext";
 import { ViewHeader } from "../components/view-header/ViewHeader";
+import { PaginatingTableToolbar } from "../components/table-toolbar/PaginatingTableToolbar";
 
 export const RealmRolesSection = () => {
+  const [max, setMax] = useState(10);
+  const [first, setFirst] = useState(0);
   const { t } = useTranslation("roles");
   const history = useHistory();
   const httpClient = useContext(HttpClientContext)!;
+  const [roles, setRoles] = useState<RoleRepresentation[]>();
   const { realm } = useContext(RealmContext);
 
   const loader = async () => {
+    const params: { [name: string]: string | number } = { first, max };
+
     const result = await httpClient.doGet<RoleRepresentation[]>(
-      `/admin/realms/${realm}/roles`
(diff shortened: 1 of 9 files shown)


jenny-s51:
at the moment, the realm role can be successfully deleted from the server, but the table is only updated on a manual reload... @edewit would you mind clarifying how you were able to get the table component to update in `ClientList` when a client is deleted? I've been looking into how to force a reload for a component but wanted to ask you if there is a better way because it looks like you didn't end up implementing a forced update at all

edewit:
@jenny-s51 Right, because also for the client list this doesn't work yet.

ssilvert:
> @jenny-s51 Right, because also for the client list this doesn't work yet, but I introduced it in #143

@jenny-s51 #143 is now merged.  Please update for that and also fix typescript errors.

jenny-s51:
@ssilvert @edewit this PR is all set and ready for review. Do you have any suggestions as to how to go about fixing the check-types error? The test passes locally for me, and alerts in `RoleList` are implemented exactly how @edewit implemented them in `ClientsList.tsx`. 

```
Error: src/realm-roles/RoleList.tsx(35,9): error TS2488: Type 'AlertProps' must have a '[Symbol.iterator]()' method that returns an iterator.
```

src/realm-roles/RealmRolesSection.tsx
```diff
@@ -1,46 +1,70 @@
...
-        >
-          <DataLoader loader={loader}>
-            {(roles) => <RolesList roles={roles.data} />}
-          </DataLoader>
-        </TableToolbar>
+        {!roles && (
+          <div className="pf-u-text-align-center">
+            <Spinner />
+          </div>
+        )}
```
srambach, line 44: I think this should be a variant="light" page section. And I think the spinner would work nicely in a <Bullseye> layout in place of this div.

src/realm-roles/add/NewRoleForm.tsx
```diff
@@ -75,6 +77,7 @@ export const NewRoleForm = () => {
               render={({ onChange, value }) => (
                 <TextArea
                   type="text"
+                  validated={value.length >= 255 ? "error" : "default"}
```
edewit, line 80: you could use the `errors` object from the "react-hook-form" instead right? No need to check the length again.

src/realm-roles/RealmRolesSection.tsx
```diff
@@ -1,46 +1,70 @@
...
 import { HttpClientContext } from "../context/http-service/HttpClientContext";
 import { RoleRepresentation } from "../model/role-model";
 import { RolesList } from "./RoleList";
 import { RealmContext } from "../context/realm-context/RealmContext";
 import { ViewHeader } from "../components/view-header/ViewHeader";
+import { PaginatingTableToolbar } from "../components/table-toolbar/PaginatingTableToolbar";
 
 export const RealmRolesSection = () => {
+  const [max, setMax] = useState(10);
+  const [first, setFirst] = useState(0);
```
edewit, line 15: does the roles endpoint support pagination?

edewit:
what is the difference between this one and #158 ?

ssilvert:
Closing this.  @jenny-s51 will submit a new PR combining it with #158.

src/realm-roles/RealmRolesSection.tsx
jenny-s51: could you kindly clarify what you mean by this @edewit ?

src/realm-roles/add/NewRoleForm.tsx
jenny-s51: nice catch, thank you

src/realm-roles/RealmRolesSection.tsx
edewit: I does support pagination, sorry about that

PR #142, opened 2020-10-01
Add Create Group Functionality

### LINKED ISSUE #124: Groups section hard-coded to master realm
Just noticed on line 21 of GroupsSection.tsx, master realm is hard coded.  You need to get the current realm from the RealmContext so that your screen will change when you use the realm selector to switch realms.

### LINKED ISSUE #130: Add modal for create groups
Add modal for creating a new group + httpPost

christiemolloy:
closes #124 and closes #130 

Adds the functionality to create a group, and then updates the state so that the new group shows immediately.

**Minor fixes included in this PR, unrelated to Create Groups:**
- Added realm context for every POST and GET
- Added alerts after POST and GET
- Replaced the header I had with the common header


src/groups/GroupsCreateModal.tsx
```diff
@@ -0,0 +1,84 @@
...
+            {t("create")}
+          </Button>
+        ]}
+      >
+        <Form isHorizontal>
+          <FormGroup
+            label={t("name")}
+            fieldId="group-id"
+            helperTextInvalid={t("common:required")}
+            validated={errors.name ? "error" : "default"}
```
christiemolloy, line 68: @edewit I don't think the "required" error is working. I copied what you had done in Clients. Is this the correct way to validate the form field?

edewit: There are 2 ways either you submit the form and use `handleSubmit` from the `react-hook-form` or you use `trigger` to trigger the validation. Maybe here we can also add server side validation to check if the group you want to create doesn't already exists?

srambach:
If I hit return on the modal instead of the button, it reloads the page, but the group doesn't seem to have been created. 
Also, when I do use the button, I get an alert saying the group has been created, but it's not showing in the list until I reload the page.

edewit:
@srambach the reload problem is handled in #143

mfrances17:
i need to do a hard refresh to see the group appear in the list... have you merged with the latest changes from master, namely those in 143?

mfrances17 (changes requested):
Code looks good - two comments... Would like to see it adding the new group name row without refreshing, does not seem to be working yet, although it could just be a merge issue. Also seeing a unique key prop warning on the groups table, probably need to ensure we are using a unique key to each row as they are added.

ssilvert:
> @srambach the reload problem is handled in #143

@christiemolloy #143 is now merged.  Please update this PR to use it.

ssilvert:
Christie will fix and resubmit later.

src/groups/GroupsList.tsx
```diff
@@ -5,26 +5,31 @@ import {
...
-import { Button } from "@patternfly/react-core";
+import { Button, AlertVariant } from "@patternfly/react-core";
 import { useTranslation } from "react-i18next";
 import { GroupRepresentation } from "./models/groups";
 import { UsersIcon } from "@patternfly/react-icons";
 import { HttpClientContext } from "../context/http-service/HttpClientContext";
+import { RealmContext } from "../context/realm-context/RealmContext";
+import { useAlerts } from "../components/alert/Alerts";
 
 type GroupsListProps = {
```
dlabaj, line 16: This should be exported since you are exporting GroupList

dlabaj (dismissed):
Looks good just that one export for the type and marks comments.

dlabaj:
@christiemolloy Also add some jest / unit tests or create a PR to add them.

srambach:
@christiemolloy Can you check the Axe report? There are some form labels it's complaining about, and a landmark for the page, which I don't know if that's new for this page or not.

src/groups/GroupsList.tsx
christiemolloy: This isn't something that is happening across the project, should we start doing that ?

christiemolloy:
@srambach looks like this is actually a problem with PatternFly. I can file an issue but I don't know that we need to worry about it for now - we can follow up if PF does something about it!

Screen Shot 2020-10-14 at 3 11 47 PM

src/groups/GroupsList.tsx
ssilvert: @christiemolloy I think it's probably a good idea to go ahead and export it.  I don't see any down side.

mfrances17 (changes requested):
I'm not sure what the purpose is of the Delete command to the right of the 'Create group' button. At first I thought that if you had multiple groups selected it would be a way to perform a bulk delete but that doesn't seem to be working.

**Edit:** This feature is coming later, not part of this PR

src/groups/GroupsCreateModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+        actions={[
+          <Button key="confirm" variant="primary" onClick={() => submitForm()}>
+            {t("create")}
+          </Button>,
+        ]}
+      >
+        <Form isHorizontal>
+          <FormGroup
+            label={t("name")}
+            fieldId="group-id"
```
srambach, line 80: I think this might need to be "name" to fix the a11y error?

mfrances17:
LGTM, just fix your conflicts :)

mfrances17 (approved):

jenny-s51 (approved):
create functionality is looking good @christiemolloy !! +1 about what @mfrances17 about bulk-delete... but it would be okay to just make a new PR for that I think 🙂

src/groups/GroupsCreateModal.tsx
christiemolloy: added a name - we could follow up if this doesnt fix it

christiemolloy: a lot of the FormGroups dont have names so might be a universal issue

src/groups/GroupsList.tsx
christiemolloy: Going to address this in a separate issue so that we handle all of the exports in one issue and make that a requirement going ahead.

PR #143, opened 2020-10-02
added function to refresh the dataloader data

edewit:


jenny-s51 (approved):
LGTM, clients list gets updated appropriately once a client is deleted. just fix merge conflict

ssilvert:
@edewit Looks like this one is holding up @jenny-s51?  

Will this be superseded by the nodejs admin client work?

jenny-s51 (commented):
just tested on my branch - the `refresh()` doesn't seem to work for me when I delete a realm role, and the state still will not update until I manually refresh

edewit:
@ssilvert It's not holding up anything and it's useful also when the new admin client is there

PR #144, opened 2020-10-02
added actions for client scopes

edewit:
[marvelapp](https://marvelapp.com/prototype/5cf7cfd/screen/72040350)

### DIFF
--- a/src/client-scopes/ClientScopesList.tsx
+++ b/src/client-scopes/ClientScopesList.tsx
@@ -1,53 +1,63 @@
-import React from "react";
-import { useTranslation } from "react-i18next";
+import React, { useContext } from "react";
+import { AlertVariant } from "@patternfly/react-core";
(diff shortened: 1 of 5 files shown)


ssilvert:
@edewit Did you mean for this to be a draft PR?  If so, why?

Also, there is a conflict.

ssilvert:
@edewit I think we need to talk with Stian about this.  As I see it, there are really three types of Default Client Scope.  They are something like, "Default", "Optional", and "Ignore on client creation".  "Ignore on client creation" is, to my understanding, the state where it is not assigned to "Default" or "Optional".  I don't remember what "Always" and "Required" were supposed to mean.

If you agree that we need more clarification, let's close this PR and resubmit later.

edewit:
@ssilvert I can add "Ignore on client creation" that is not hard

PR #147, opened 2020-10-05
Add empty state

### LINKED ISSUE #128: Add empty state to first Groups page
Add empty state component for when there are no groups, and for when filtering on groups

christiemolloy:
closes #128 

Adds empty state for No Groups:
Screen Shot 2020-10-05 at 4 15 39 PM

Adds empty state for no search results:
Screen Shot 2020-10-05 at 4 13 35 PM


srambach:
Instead of extending the ListEmptyState to handle what we want a search empty state to look like, what do you think about creating a SearchEmptyState component? 
I'm also wondering if that should have the magnifying glass icon, so if we make it a separate component we could decide later and update that all at once for all empty searches.

christiemolloy:
> Instead of extending the ListEmptyState to handle what we want a search empty state to look like, what do you think about creating a SearchEmptyState component?
> I'm also wondering if that should have the magnifying glass icon, so if we make it a separate component we could decide later and update that all at once for all empty searches.

I think its a good idea to have a variation for the empty state that is specific to search, but i dont know that we need to create a new component, i think we could just pass in a prop and that prop would change the icon to the search icon, how does that sound?

srambach:
That's fine with me I think.

edewit (changes requested):
The text needs translation

edewit (approved):
:+1:

srambach:
Can you make it so that you can't get both icons if you turn them both on?

edewit:
@srambach @christiemolloy Instead of the 2 booleans you could maybe introduce a variant property like `ButtonVariant` or `ModalVariant`

srambach:
@christiemolloy @edewit Is adding lots of variants and props just recreating the Patternfly empty state? My intention when I created `ListEmptyState` was just to make an opinionated empty state that did most everything for you when you have a list that is empty - correct icon, etc. If we make it so flexible then we might as well just use the PF component. WDYT? Is adding a separate opinionated empty state for search results a bad idea?

dlabaj (approved):
LGTM.  Do we need any updates to jest tests? If so either create a new issue and add them or add them to this PR.

christiemolloy:
> LGTM. Do we need any updates to jest tests? If so either create a new issue and add them or add them to this PR.

I wasn't sure that we were adding any tests yet ? are there requirements that we need to follow for the tests? If we are adding them ill create a separate follow-up issue to write them. @edewit @ssilvert

edewit:
@srambach I don't think 2 is a lot. We want a simple / small abstraction to quickly create the patterns that we need for this app

mfrances17 (approved):
LGTM... let's get this in and add a new issue for unit tests that references this PR as a TBD until we discuss how we're handling the jest tests

ssilvert:
> > LGTM. Do we need any updates to jest tests? If so either create a new issue and add them or add them to this PR.
> 
> I wasn't sure that we were adding any tests yet ? are there requirements that we need to follow for the tests? If we are adding them ill create a separate follow-up issue to write them. @edewit @ssilvert

@christiemolloy 

We don't have hard requirements saying what kind of jest tests to create.  But I'm open to suggestions on general principles we can adopt.

There are definitely instances where jest tests can help you develop the code so those are always helpful to include with a PR.

As for this PR, I'm fine with merging it and doing a follow up task if there are more tests that need to be written.

PR #152, opened 2020-10-06
Field level help

srambach:
I have a question about dynamic translation of an item passed into a component. In HelpItem.tsx you can see a couple of comments with questions about this. In the form, the item would be translated and come up with the field name. However, in the HelpItem component it can't translate it because it doesn't know that components json. I think... @edewit do you have a suggestion for handling this?

Also, there was an attribute I don't recognize `data-testid={item}` and I wasn't sure if that needed to be carried forward or not.


src/components/help-enabler/HelpItem.tsx
```diff
@@ -1,24 +1,35 @@
...
-          </span>
-        </Tooltip>
+        // <Tooltip position="right" content={t(`help:${item}`)}>
+        //   <span id={item} data-testid={item}> *** I don't know what data-testid was being used for ***
+        //     <HelpIcon />
+        //   </span>
+        // </Tooltip>
+        <Popover bodyContent={t(`help:${item}`)}>
+          <button
+            aria-label={t(`helpLabel`) + " " + item} // *** this should say "more help for the [field label]"
```
ssilvert, line 25: You need to use interpolation here to pass the item as a param to t().  See https://www.i18next.com/translation-function/interpolation

srambach: @ssilvert That's cool. I still get something like `More help for clientScope.name` though instead of getting the correct label "name". Can you see if I'm doing it right? Thanks!

src/components/help-enabler/HelpItem.tsx
```diff
@@ -1,24 +1,35 @@
...
   return (
     <>
       {enabled && (
-        <Tooltip position="right" content={t(`help:${item}`)}>
-          <span id={item} data-testid={item}>
-            <HelpIcon />
-          </span>
-        </Tooltip>
+        // <Tooltip position="right" content={t(`help:${item}`)}>
+        //   <span id={item} data-testid={item}> *** I don't know what data-testid was being used for ***
```
edewit, line 19: It was used in the tests but that is no longer the case.

src/components/help-enabler/HelpItem.tsx
```diff
@@ -1,24 +1,35 @@
 import React, { useContext } from "react";
-import { Tooltip } from "@patternfly/react-core";
+import { Button, Popover } from "@patternfly/react-core";
```
edewit, line 2: Unused import
```suggestion
import { Popover } from "@patternfly/react-core";
```

src/components/help-enabler/HelpItem.tsx
```diff
@@ -1,24 +1,35 @@
...
-          </span>
-        </Tooltip>
+        // <Tooltip position="right" content={t(`help:${item}`)}>
+        //   <span id={item} data-testid={item}> *** I don't know what data-testid was being used for ***
+        //     <HelpIcon />
+        //   </span>
+        // </Tooltip>
+        <Popover bodyContent={t(`help:${item}`)}>
+          <button
+            aria-label={t(`helpLabel`, { label: item })} // *** this should say "more help for the [field label]"
```
edewit, line 25: Okay I see 3 solutions to that problem:
1. change `itemFor` from id to name
1. change id to name in `NewClientScopeForm`
1. add another parameter for the name of the field this help item is for.

src/components/help-enabler/HelpItem.tsx
edewit: @srambach see my comment on the comment ;)

edewit:
@srambach is it really an improvement that you have to click the help icons?

ssilvert:
> @srambach is it really an improvement that you have to click the help icons?

I can see where it might be an improvement.  "Hover help" goes away when you stop hovering.  Sometimes it is helpful to see the help text while you fill in the field.  So clicks to show/hide are nice to have in some cases.

But I'm wondering if this is a PatternFly recommendation or not?

edewit:
It might also not be clear that you have to click them as they aren't / look like buttons

srambach:
@ssilvert @edewit This is the Patternfly recommendation. https://www.patternfly.org/v4/components/form
I've been in touch with a couple of a11y people about this and although not perfect, this is the best we have at the moment to accommodate a wide variety of users. Disclosure of information on hover presents a multitude of issues, from keyboard only users, to those whose mouse pointer might be shaky, those using screen readers, and those who may need or want the text to remain visible while looking at something else.

edewit:
@srambach so that brings up the question why does patternfly even have a hover tooltip component

ssilvert:
> @srambach so that brings up the question why does patternfly even have a hover tooltip component

My guess is that the hover tooltip will go away.  I think what @srambach said makes a lot of sense and it will probably become an industry standard practice that you should never rely solely on hover because so many have trouble with it.  

I put it in the same category as the rule that you should never rely solely on color to impart information because over 4% of the world is color blind.

ssilvert:
> It might also not be clear that you have to click them as they aren't / look like buttons

If you do hover over it then the cursor changes to indicate it is clickable.  But I think you are right that it needs to look like a more clickable component.  There is no reason that this could not look like a round button.

edewit (approved):
:+1:

PR #159, opened 2020-10-08
added client scope details screen

edewit:
[marvel app](https://marvelapp.com/prototype/5cf7cfd/screen/72040359)


ssilvert (changes requested):
`Include in token scope` is missing from this form.

Also, `type` is missing.

edewit:
@ssilvert don't know why I did it, but I'm rendering `type` only when you re creating a client-scope not when editing. Changed it back to always rending it.

edewit:
ahh found it here: https://marvelapp.com/prototype/5cf7cfd/screen/72040359 seems I need a different way to know when not to render what.

ssilvert:
@edewit Doesn't matter if we address this here or later, but throughout Client scopes, `GUI order` should be called `Display order`.  This is something we changed in design discussions, but some of the wireframes still don't reflect that decision.

ssilvert (approved):

PR #158, opened 2020-10-07
Delete role modal

jenny-s51:
closes #157

PR #160, opened 2020-10-08
Adds functionality to delete realm role

### LINKED ISSUE #135: add logic to delete realm role

jenny-s51:
closes #135 , #131 and extends PR review comments from @srambach and @edewit on #141 and #158

### DIFF
--- a/src/realm-roles/RealmRolesSection.tsx
+++ b/src/realm-roles/RealmRolesSection.tsx
@@ -1,7 +1,13 @@
 import React, { useContext, useEffect, useState } from "react";
 import { useHistory } from "react-router-dom";
 import { useTranslation } from "react-i18next";
-import { Button, Divider, PageSection, Spinner } from "@patternfly/react-core";
+import {
+  Bullseye,
+  Button,
+  Divider,
+  PageSection,
+  Spinner,
+} from "@patternfly/react-core";
 
 import { HttpClientContext } from "../context/http-service/HttpClientContext";
 import { RoleRepresentation } from "../model/role-model";
@@ -36,11 +42,11 @@ export const RealmRolesSection = () => {
   return (
     <>
       <ViewHeader titleKey="roles:title" subKey="roles:roleExplain" />
-      <PageSection padding={{ default: "noPadding" }}>
+      <PageSection variant="light">
         {!roles && (
-          <div className="pf-u-text-align-center">
+          <Bullseye className="pf-u-text-align-center">
             <Spinner />
-          </div>
+          </Bullseye>
         )}
         <Divider component="li" key={1} />
         {roles && (
--- a/src/realm-roles/RealmRolesSection.tsx
+++ b/src/realm-roles/RealmRolesSection.tsx
@@ -1,46 +1,78 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { useHistory } from "react-router-dom";
 import { useTranslation } from "react-i18next";
-import { Button, PageSection } from "@patternfly/react-core";
+import {
+  Button,
+  Divider,
+  PageSection,
+  Spinner,
+} from "@patternfly/react-core";
 
-import { DataLoader } from "../components/data-loader/DataLoader";
-import { TableToolbar } from "../components/table-toolbar/TableToolbar";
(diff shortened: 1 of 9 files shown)


src/realm-roles/RealmRolesSection.tsx
```diff
@@ -1,46 +1,76 @@
...
-            </>
-          }
-        >
-          <DataLoader loader={loader}>
-            {(roles) => <RolesList roles={roles.data} />}
-          </DataLoader>
-        </TableToolbar>
+      <PageSection variant="light">
+        {!roles && (
+          <Bullseye className="pf-u-text-align-center">
```
srambach, line 47: Jenny, you shouldn't need to have this utility class since you are using the Bullseye. Otherwise, LGTM!

src/realm-roles/RoleList.tsx
```diff
@@ -43,37 +58,83 @@ export const RolesList = ({ roles }: RolesListProps) => {
...
+              setSelectedRowId(rowId);
+            },
+          },
+        ]}
+        aria-label="Roles list"
+      >
+        <TableHeader />
+        <TableBody />
+      </Table>
+      <Modal
```
edewit, line 98: Instead of this you could use the ConfirmDialog like this:

```
  const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
    titleKey: "roles:deleteConfirmTitle",
    messageKey: "roles:deleteConfirm",
    continueButtonLabel: "common:delete",
    continueButtonVariant: ButtonVariant.danger,
    onConfirm: () => {
      try {
        httpClient.doDelete(`/admin/realms/${realm}/roles/${data[rowId].role.name}`);
        addAlert(t("roleDeletedSuccess"), AlertVariant.success);
      } catch (error) {
        addAlert(`${t("roleDeleteError")} ${error}`, AlertVariant.danger);
      }
    },
  });
```

edewit (changes requested):

src/realm-roles/RoleList.tsx
jenny-s51: ah, thank you for the suggestion and for making this component @edewit  - will do!

ssilvert:
@jenny-s51 I found a nasty bug in this code.

1. Create a realm role
2. Delete the realm role
3. Create another realm role
4. Delete the second realm role
5. App crashes with `RoleList.js:68 Uncaught TypeError: Cannot read property 'role' of undefined`

src/realm-roles/RealmRolesSection.tsx
```diff
@@ -1,46 +1,76 @@
...
-            {(roles) => <RolesList roles={roles.data} />}
-          </DataLoader>
-        </TableToolbar>
+      <PageSection variant="light">
+        {!roles && (
+          <Bullseye>
+            <Spinner />
+          </Bullseye>
+        )}
+        <Divider component="li" key={1} />
```
ssilvert, line 51: I think this is creating a second divider.  There are two on the screen.

src/realm-roles/RoleList.tsx
```diff
@@ -65,8 +60,29 @@ export const RolesList = ({ roles, refresh }: RolesListProps) => {
   const selectedRoleName =
     selectedRowId != -1 ? data[selectedRowId].role.name : "";
 
+  const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
+    titleKey: "roles:roleDeleteConfirm",
+    messageKey: t(
+      `This action will permanently delete the role ${selectedRoleName} and cannot be undone.`
```
edewit, line 66: move this text to the `message.json` file. You can add the `selectedRoleName` in there as a parameter like: "dialogBody": "This action will permanently delete the role {{selectedRoleName}} and cannot be undone." and then use it: `t("dialogBody", {selectedRoleName})`

ssilvert: @jenny-s51 Just to make sure you understand why Ed wants you to do it that way, whenever there is a variable inside the text you should always put the variable in the message bundle instead of resolving it in your code.  Otherwise, the text in the message bundle can not be translated to another language.

jenny-s51: makes sense - thank you for clarifying @ssilvert !

ssilvert (approved):

PR #162, opened 2020-10-09
initial version of the mapping list

edewit:
[marvel](https://marvelapp.com/prototype/f0e8fih/screen/72041678)


src/client-scopes/details/MapperList.tsx
```diff
@@ -0,0 +1,141 @@
...
+        mapper,
+        cells: [
+          mapper.name,
+          mapperType.category,
+          mapperType.name,
+          mapperType.priority,
+        ],
+      };
+    })
+    .sort((a, b) => (a.cells[3] as number) - (b.cells[3] as number));
```
ssilvert, line 70: I think you can probably get rid of these casts to `as number` and `as string`?  Maybe define a tuple type?

ssilvert (commented):
Other than my one possible suggestion, LGTM.

src/client-scopes/details/MapperList.tsx
edewit: The `Table` component needs an array and you can't use the types to specify that the 3rd element in the array is a `string`. I've added an extra step in the transform, but that means iterating another time to change from the introduced `Row` type back to array.

ssilvert (approved):

PR #163, opened 2020-10-09
added ability to add mulitple actions

edewit:


ssilvert (commented):
I may be reading the code wrong but it looks like the primary action will also be rendered as a secondary action?

Since primary and secondary actions are treated differently, I suggest changing the props to accept a primary action as one prop, and an array of secondary actions as a different prop.  It doesn't make sense to me to have the first element in the array be "special".

edewit:
no the primary action is the fist one in the list of actions, right but let's change it to only add secondary actions

ssilvert (approved):

PR #164, opened 2020-10-09
added download dialog to client actions

edewit:
[marvel](https://marvelapp.com/prototype/908c3c3/screen/72040453)

### DIFF
--- a/src/clients/ClientSettings.tsx
+++ b/src/clients/ClientSettings.tsx
@@ -31,6 +31,7 @@ import { useAlerts } from "../components/alert/Alerts";
 import { ViewHeader } from "../components/view-header/ViewHeader";
 import { exportClient } from "../util";
 import { useConfirmDialog } from "../components/confirm-dialog/ConfirmDialog";
+import { useDownloadDialog } from "../components/download-dialog/DownloadDialog";
 
 export const ClientSettings = () => {
   const { t } = useTranslation("clients");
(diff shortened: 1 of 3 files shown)


ssilvert:
"Download adapter" bleeds off the screen:
menu-bleed

ssilvert (changes requested):
Other than the bleed issue, this looks great!

edewit:
Right I did see that of course, thought we could add it as an issue, because we need some custom css I think to fix that @srambach ?

srambach:
@edewit I think you can just add `position={DropdownPosition.right}` to your dropdown

edewit:
@srambach it's a `<Select>` it only has `direction="up | down"`

srambach:
Hmm. Semantically I would expect that to be a dropdown. Typically a select should be used when the element needs to hold a value, whereas a dropdown is preferred for an action menu. I did not notice that before; I think it's worth an issue to swap that out and then the alignment can easily be fixed as well.

edewit:
It's in your ViewHeader... so...

ssilvert (approved):

ssilvert:
I'm going to go ahead and merge this.  Created issue so @edewit and/or @srambach can address the "bleed" issue.  #171

srambach:
> It's in your ViewHeader... so...

Yeah, I know 🤦 I can take #171 to switch it if you want @edewit

edewit:
be my guest...

PR #166, opened 2020-10-12
adds optional link to the subkey of a view header

srambach:
This adds props to the ViewHeader to pass in link text and a url, which will be appended to the subKey text.

Added to allow for a link in the help text of a header, as shown here: https://marvelapp.com/prototype/ecjc94b/screen/72071043

### DIFF
--- a/src/components/view-header/ViewHeader.tsx
+++ b/src/components/view-header/ViewHeader.tsx
@@ -16,11 +16,14 @@ import {
 import { HelpContext } from "../help-enabler/HelpHeader";
 import { useTranslation } from "react-i18next";
 import { PageBreadCrumbs } from "../bread-crumb/PageBreadCrumbs";
+import { ExternalLink } from "../external-link/ExternalLink";
 
 export type ViewHeaderProps = {
   titleKey: string;
   badge?: string;
   subKey: string;
+  subKeyLinkText?: string;
+  subKeyLinkAddress?: string;
   selectItems?: ReactElement[];
   isEnabled?: boolean;
   onSelect?: (value: string) => void;
@@ -31,6 +34,8 @@ export const ViewHeader = ({
   titleKey,
   badge,
   subKey,
+  subKeyLinkText,
+  subKeyLinkAddress,
   selectItems,
   isEnabled = true,
   onSelect,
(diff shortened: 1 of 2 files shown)


src/components/view-header/ViewHeader.tsx
```diff
@@ -16,11 +16,14 @@ import {
 import { HelpContext } from "../help-enabler/HelpHeader";
 import { useTranslation } from "react-i18next";
 import { PageBreadCrumbs } from "../bread-crumb/PageBreadCrumbs";
+import { ExternalLink } from "../external-link/ExternalLink";
 
 export type ViewHeaderProps = {
   titleKey: string;
   badge?: string;
   subKey: string;
+  subKeyLinkText?: string;
```
dlabaj, line 25: ExternalLink takes ButtonProps.  You can use that instead of these two props to get the title and hyperlink.  

`subKeyLinkProps?: ButtonProps`

src/components/view-header/ViewHeader.tsx
```diff
@@ -98,7 +103,17 @@ export const ViewHeader = ({
         </Level>
         {enabled && (
           <TextContent>
-            <Text>{t(subKey)}</Text>
+            <Text>
```
dlabaj, line 106: I believe you can the following here: 
`<TextContent>
            <Text>
              {t(subKey)}
              {subKeyLinkProps && (
                <ExternalLink 
                  {...subKeyLinkProps} 
                  isInline
                  className="pf-u-ml-md"
                />)}
            </Text>
          </TextContent>`

src/stories/ViewHeader.stories.tsx
```diff
@@ -22,6 +22,8 @@ Extended.args = {
   titleKey: "This is the title",
   badge: "badge",
   subKey: "This is the description.",
+  subKeyLinkText: "More information",
```
dlabaj, line 25: Then you can do this here: 
`const subKeyLinkProps:{title: "More information", href: 'http://www.google.com'},`

dlabaj (changes requested):
See if we can use ButtonProps instead of having these two props.

dlabaj (approved):
LGTM!

dlabaj:
Can rebase master and then the failure will go away.

PR #167, opened 2020-10-12
Fixed build by making secondaryActions optional

dlabaj:
Secondary actions in ListEmptyState.tsx is an optional prop.  This fix updates the code to make it optional, so that secondary actions are available when they are needed on specific pages.

Here's an example:

https://marvelapp.com/prototype/f0e8fih/screen/72043135


srambach (approved):
Looks good as far as I can tell 😄  👍

mfrances17 (approved):
LGTM

PR #169, opened 2020-10-13
Introduced a predefined mapper dialog

edewit:
[marvel app](https://marvelapp.com/prototype/f0e8fih/screen/72041878)


edewit:
made it a little different as I don't see how to do clickable rows in a table... I think like this it's more "in style" with patternfly

srambach:
I think this is meant to be a data list with selectable rows rather than a table. As it is, there's no indication that anything is clickable but the data list has that built in.  https://www.patternfly.org/v4/components/data-list#selectable-rows

ssilvert:
> I think this is meant to be a data list with selectable rows rather than a table. As it is, there's no indication that anything is clickable but the data list has that built in. https://www.patternfly.org/v4/components/data-list#selectable-rows

Yes, that is what was intended by Kun.  If you hover over the pattern link in the design you see it is datalist#selectableRows.  The link is broken but that's clearly what was intended.

The question I have about it is how do you dismiss the dialog after a row is selected.  There doesn't seem to be any way to do that.

ssilvert (approved):

ssilvert (approved):

PR #170, opened 2020-10-13
introduced "/" for keycloak.json

edewit:
fixes: #165


ssilvert (commented):
@edewit We need to think about this some more.  This solution will only work for the master realm.

edewit:
I don't know why you say that, but I'll take your word for it.

edewit:
For me it works on every realm, this just solves the 404 you get for the `keycloak.json` when on a details page and pressing refresh

ssilvert:
Hmm.  OK.  I'll merge it then.  Sounds like this is at least a step forward.

ssilvert (approved):

PR #175, opened 2020-10-14
Fixes for eslint and prettier

mfrances17:
The linting configuration using eslint and prettier was not working with the VSCode intellisense, you could not visually see the issues as you were working, although you could force a format with keystrokes. This fix will allow you to see the issues in the IDE in real-time as you are working, and hover over them >  _Quick Fix_ > _Fix this prettier/prettier problem_.

lint-int-found

lint-int-fix

### DIFF
--- a/src/client-scopes/details/MapperList.tsx
+++ b/src/client-scopes/details/MapperList.tsx
@@ -25,16 +25,16 @@ import { HttpClientContext } from "../../context/http-service/HttpClientContext"
 import { RealmContext } from "../../context/realm-context/RealmContext";
 import { useAlerts } from "../../components/alert/Alerts";
 
-interface MapperListProps {
+type MapperListProps = {
   clientScope: ClientScopeRepresentation;
-}
+};
 
-interface Row {
+type Row = {
   name: string;
   category: string;
   type: string;
   priority: number;
-}
+};
 
 export const MapperList = ({ clientScope }: MapperListProps) => {
   const { t } = useTranslation("client-scopes");
--- a/src/client-scopes/details/MapperList.tsx
+++ b/src/client-scopes/details/MapperList.tsx
@@ -4,20 +4,20 @@ import {
   AlertVariant,
   Dropdown,
   DropdownItem,
-  DropdownToggle,
+  DropdownToggle
 } from "@patternfly/react-core";
 import {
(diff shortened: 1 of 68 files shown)


src/client-scopes/details/MapperList.tsx
```diff
@@ -4,37 +4,37 @@ import {
...
 import { RealmContext } from "../../context/realm-context/RealmContext";
 import { useAlerts } from "../../components/alert/Alerts";
 
-type MapperListProps = {
+interface MapperListProps {
   clientScope: ClientScopeRepresentation;
-};
+}
 
-type Row = {
```
edewit, line 32: I like the type notation when it's a simple type why do we want to change that?

edewit (changes requested):
Is this something that changed by mistake, because I only see this one change and I'm sure that we use it all over the place? And could you elaborate on what the actual problem is, because I don't have any issues with VS code

src/client-scopes/details/MapperList.tsx
ssilvert: @mfrances17 We have decided to use `type` for everything except the (data transfer) Representation types.  The only reason we use `interface` for that is because many are auto-generated and the tool that generates these types uses `interface`.

According to "Effective Typescript", `type` has advantages over `interface` because it can express:
- Union types
- Mapped types
- Tuple types

The only thing an `interface` can do that a `type` can not do is "declaration merging", which is weird and tends to only be used in *.d.ts files.

mfrances17: i removed the offending type/interface rule and reverted the related changes that were made to MapperList.tsx

dlabaj (approved):

dlabaj:
@mfrances17 You need to run prettier fix.  The build is failing.

mfrances17:
> @mfrances17 You need to run prettier fix. The build is failing.

Yep, I noticed that earlier and am in the process of fixing them now. Thanks.

mfrances17:
Not sure why this is still failing on the server, it passes locally. There may be a difference in versions, although nothing is flagged in the diff. We might need to hold off on this until I get some help on figuring out why the exact same files would fail on the server and pass on my local system.

PR #177, opened 2020-10-15
initial version of the role mapping create form

edewit:
[marvel](https://marvelapp.com/prototype/f0e8fih/screen/72042619)

I think this screen needs some discussion / explanation on how it's supposed to work. As I see from the screen shots the first dropdown has realm roles and clients grouped and a search that only filters clients? I think this would be very confusing for the user. Also that leaves the second dropdown to show the client roles once a client is chosen? On top of this we can't do a filter input on a one select dropdown as it's only supported on a mutli select.


edewit:
now uses typeahead and new design: [marvel](https://marvelapp.com/prototype/f0e8fih/section/1103591/screen/72042619)

srambach:
@edewit How do you get to this in the interface?

edewit:
@srambach right now only through the story `RoleMappingForm.stories.tsx` still need the other PRs to get merged to do the entire flow

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+        titleKey="client-scopes:addMapper"
+        subKey="client-scopes:addMapperExplain"
+      />
+      <PageSection variant="light">
+        <Form isHorizontal onSubmit={handleSubmit(save)}>
+          <FormGroup
+            label={t("protocolMapper")}
+            labelIcon={
+              <HelpItem
+                helpText="client-scopes-help:protocolMapper"
```
srambach, line 148: Looks like the helpText on all these needs the translation.

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+              name="name"
+            />
+          </FormGroup>
+          <FormGroup
+            label={t("role")}
+            labelIcon={
+              <HelpItem
+                helpText="client-scopes-help:role"
+                forLabel={t("role")}
+                forID="role"
```
srambach, line 190: `forID` should match a `toggleId` on the select.

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+            helperTextInvalid={t("common:required")}
+            fieldId="role"
+          >
+            <Gallery hasGutter>
+              <GalleryItem>
+                <Select
+                  onToggle={() => setClientsOpen(!clientsOpen)}
+                  isOpen={clientsOpen}
+                  variant={SelectVariant.typeahead}
+                  aria-label={t("selectASourceOfRoles")}
```
srambach, line 203: This looks like it's coming in blank maybe?

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+                      onToggle={() => setRoleOpen(!roleOpen)}
+                      isOpen={roleOpen}
+                      variant={SelectVariant.typeahead}
+                      placeholderText={
+                        selectedClient && selectedClient.name !== "realmRoles"
+                          ? t("clientRoles")
+                          : t("selectARole")
+                      }
+                      isDisabled={!selectedClient}
+                      aria-label="Select Input"
```
srambach, line 249: I think this looks like it's coming in blank as well, and should be translated as well.

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+                helpText="client-scopes-help:role"
+                forLabel={t("role")}
+                forID="role"
+              />
+            }
+            validated={errors["config.role"] ? "error" : "default"}
+            helperTextInvalid={t("common:required")}
+            fieldId="role"
+          >
+            <Gallery hasGutter>
```
srambach, line 197: This is probably better as a Split than a Gallery.

src/client-scopes/add/RoleMappingForm.tsx
edewit: Yes it's not rendering it, is this a know patternfly bug?

edewit: ahh no found it should use `typeAheadAriaLabel` instead of `aria-label`

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,297 @@
...
+              />
+            }
+            validated={errors["config.role"] ? "error" : "default"}
+            helperTextInvalid={t("common:required")}
+            fieldId="role"
+          >
+            <Split hasGutter>
+              <SplitItem>
+                <Select
+                  id="role"
```
srambach, line 202: this needs to be toggleId so that the icon can refer to it

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,297 @@
...
+                      onToggle={() => setRoleOpen(!roleOpen)}
+                      isOpen={roleOpen}
+                      variant={SelectVariant.typeahead}
+                      placeholderText={
+                        selectedClient && selectedClient.name !== "realmRoles"
+                          ? t("clientRoles")
+                          : t("selectARole")
+                      }
+                      isDisabled={!selectedClient}
+                      aria-label={t("selectARole")}
```
srambach, line 252: And this would also be `typeAheadAriaLabel` then

srambach:
One question - if I select something in the first dropdown, then the second, then clear the first - should the second dropdown be cleared or left alone?
image

If that's good as is, then 👍 LGTM.

ssilvert:
> [marvel](https://marvelapp.com/prototype/f0e8fih/screen/72042619)
> 
> I think this screen needs some discussion / explanation on how it's supposed to work. As I see from the screen shots the first dropdown has realm roles and clients grouped and a search that only filters clients? I think this would be very confusing for the user. Also that leaves the second dropdown to show the client roles once a client is chosen? On top of this we can't do a filter input on a one select dropdown as it's only supported on a mutli select.

So after trying this out, I really agree with your original comments.  I think this needs to go back to UXD.  It's very confusing.  Frankly, the old console has better usability for this.

ssilvert:
Closing because the UX design needs review.

edewit:
Added group and divider to make it more obvious
image

src/client-scopes/add/RoleMappingForm.tsx
```diff
@@ -0,0 +1,297 @@
...
+        <SelectOption
+          key="realmRoles"
+          value={
+            {
+              name: "realmRoles",
+              toString: () => t("realmRoles"),
+            } as ClientRepresentation
+          }
+        >
+          {t("realmRoles")}
```
ssilvert, line 111: I think this should be the name of the realm, such as "Master" or "Sales and Marketing".  The way this turns out is that it says "Use a realm role from: Realm roles"

ssilvert (changes requested):

src/client-scopes/add/RoleMappingForm.tsx
edewit: Right missed that change... fix now

ssilvert (approved):

PR #179, opened 2020-10-15
Change list view header action menu from select to dropdown

### LINKED ISSUE #171: Fix bleed issue with "Download Adapter" item
See PR #164. "Download Adapter" is cut off.

srambach:
Fixes #171 
The action menu should have been a dropdown rather than select, and a dropdown allows a right alignment, so this fixes the overflow off the right side as well as correcting the semantics of the element.

To verify, look at the ViewHeader story and also the Client settings page (actions menu in the header)

### DIFF
--- a/src/components/view-header/ViewHeader.tsx
+++ b/src/components/view-header/ViewHeader.tsx
@@ -87,7 +87,6 @@ export const ViewHeader = ({
                     />
                   </ToolbarItem>
                   <ToolbarItem>
-                    {/* This is the dropdown to replace the select.  It operates but does not initiate the actions. See it by going to Clients and click on a client. All 3 menu actions should work but don't. */}
                     <Dropdown
                       position={DropdownPosition.right}
                       toggle={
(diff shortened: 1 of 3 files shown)


src/components/view-header/ViewHeader.tsx
```diff
@@ -81,19 +87,17 @@ export const ViewHeader = ({
...
-                      onSelect={(_, value) => {
-                        if (onSelect) {
-                          onSelect(value as string);
-                        }
-                        setOpen(false);
-                      }}
-                    >
-                      {selectItems}
-                    </Select>
+                    {/* This is the dropdown to replace the select.  It operates but does not initiate the actions. See it by going to Clients and click on a client. All 3 menu actions should work but don't. */}
```
jenny-s51, line 90: delete this line

jenny-s51 (changes requested):
Looking good @srambach ! Actions menu working as expected per @mfrances17 . Just delete that comment above :)

edewit (approved):
:tada: so much better

jenny-s51 (approved):
👍

PR #184, opened 2020-10-21
Change the input fields based on access rights

edewit:
Possible way to implement how to change the input fields based on access rights

## Brief Description

This is a poc to get some feedback it's not intented to merge as is

## Verification Steps

Have a look at the client settings screen when you don't hve the "manage-clients" role 3 field should be disabled

## Checklist:

### DIFF
--- a/src/components/form-access/FormAccess.tsx
+++ b/src/components/form-access/FormAccess.tsx
@@ -1,20 +1,34 @@
+import { Form, FormProps } from "@patternfly/react-core";
 import React, {
   Children,
   cloneElement,
+  createElement,
   isValidElement,
   ReactElement,
 } from "react";
+import { Controller, useForm } from "react-hook-form";
 import { useAccess } from "../../context/access/Access";
 import { AccessType } from "../../context/whoami/who-am-i-model";
 
-export type FormAccessProps = {
+export type FormAccessProps = FormProps & {
   role: AccessType;
   fineGrainedAccess?: boolean;
+  unWrap?: boolean;
   children: ReactElement[];
 };
 
-export const FormAccess = ({ children, role, fineGrainedAccess = true }: FormAccessProps) => {
+export const FormAccess = ({
+  children,
+  role,
+  fineGrainedAccess = false,
+  unWrap = false,
+  ...rest
+}: FormAccessProps) => {
   const { hasAccess } = useAccess();
+
+  const ReadOnly = (props: any) => (
+    <input type="text" value={props.value} onChange={props.onChange} disabled />
+  );
 
   const recursiveCloneChildren = (children: ReactElement[], newProps: any) => {
     return Children.map(children, (child: ReactElement) => {
@@ -23,6 +37,12 @@ export const FormAccess = ({ children, role, fineGrainedAccess = true }: FormAcc
       }
 
       if (child.props) {
+        if (child.type === Controller) {
+          return cloneElement(child, {
+            ...(child as ReactElement).props,
+            render: ReadOnly,
+          });
+        }
         newProps.children = recursiveCloneChildren(
           (child as ReactElement).props.children,
           newProps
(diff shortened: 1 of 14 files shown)


edewit:
image

ssilvert (commented):
I like what you did.  I think that general idea will work out well.

After looking at how fine grained permissions works, it looks like we will need an additional parameter.  If an entity has fine grained permissions enabled, it will have an access property.  In the case of ClientRepresentation, it looks like this:
```
{
    "id": "e0edbf2a-0e6d-4f83-8500-8099141653f0",
    "clientId": "account",
    "name": "${client_account}",
    "rootUrl": "${authBaseUrl}",
    "baseUrl": "/realms/test/account/",
...
    "access": { "view": true, "configure": true, "manage": true }
  }
```
So for example, if a user only has "query-clients" role, but his fine-grained access for a particular client gives him `manage=true`, he would still be allowed to edit.  So for FormAccess component, you would need to somehow say `<FormAccess role="manage-clients" fineGrainedAccess={client.access.manage} />`

ssilvert:
I can't see from your picture, but does this affect the Save button?

edewit:
it doesn't yet atm, but we can add that as well

yarn.lock
```diff
@@ -15237,7 +15242,7 @@ prepend-http@^1.0.0:
   resolved "https://registry.yarnpkg.com/prepend-http/-/prepend-http-1.0.4.tgz#d4f4562b0ce3696e41ac52d0e002e57a635dc6dc"
   integrity sha1-1PRWKwzjaW5BrFLQ4ALlemNdxtw=
 
-prettier@2.1.2:
+prettier@^2.0.5:
```
ssilvert, line 15245: @edewit How did this get changed to an older version of prettier?

edewit: I don't know I only added `mutationobserver-shim` for testing

src/components/form-access/FormAccess.tsx
```diff
@@ -0,0 +1,92 @@
...
+      }
+      return child;
+    });
+  };
+  return (
+    <>
+      {!unWrap && (
+        <Form {...rest}>
+          {recursiveCloneChildren(children, {
+            isDisabled: !hasAccess(role) && !fineGrainedAccess,
```
ssilvert, line 80: Need to change this attribute to `isdisabled`.  Got this warning in the console:
```
Warning: React does not recognize the `isDisabled` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isdisabled` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in textarea (created by TextArea)
    in TextArea (created by ClientSettings)
    in div (created by FormGroup)
    in div (created by FormGroup)
```

src/components/form-access/FormAccess.tsx
```diff
@@ -0,0 +1,92 @@
...
+            readOnly: !hasAccess(role) && !fineGrainedAccess,
+          })}
+        </Form>
+      )}
+      {unWrap &&
+        recursiveCloneChildren(children, {
+          isDisabled: !hasAccess(role) && !fineGrainedAccess,
```
ssilvert, line 87: Same as above.

ssilvert (commented):
Also getting this warning in console:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in DownloadDialog (created by Dialog)
    in Dialog (created by ClientSettings)
```

yarn.lock
ssilvert: I just rebuilt everything and it changed my `yarn.lock` as well.

ssilvert:
Minor bug: You can still add fields to the `Valid redirect URIs` when the form is disabled.

src/components/form-access/FormAccess.tsx
edewit: The problem is that when I add the `isDisabled` property to a component that doesn't support it it will copy all the attributes it doesn't know about to a it's childs DOM element. For instance FormGroup doesn't have a `isDisabled` so it will create a `<div` like this: `<div {...rest}` so the `isDisabled` will get copied to that div. What we need to do is exclude the child components that don't support it.

src/components/form-access/FormAccess.tsx
edewit: > Also getting this warning in console:
> 
> ```
> Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
>     in DownloadDialog (created by Dialog)
>     in Dialog (created by ClientSettings)
> ```

This also happens on master, so I created an issue for it: #190

edewit:
Created yet another workaround, this code is not getting prettier! Added an issue for `TextArea` in patternfly https://github.com/patternfly/patternfly-react/issues/5068

ssilvert (approved):

PR #188, opened 2020-10-23
Allow login to different realms using query param

ssilvert:
## Motivation
Need to be able to develop and test when user logs in using a realm other than master.

The mechanism for determining the realm and the auth URL will need to change as we figure out how this is going to be deployed.  So this is a temporary fix.

## Brief Description
You can now go to admin console with http://localhost:8080/?realm=test

## Verification Steps

1. Create a realm called "test"
2. Add a user to test realm
3. Give the user permissions using realm-management client roles
4. Optional step: Set up fine grained permissions
5. http://localhost:8080/?realm=test
6. Log in.

## Checklist:

### DIFF
--- a/src/components/realm-selector/RealmSelector.tsx
+++ b/src/components/realm-selector/RealmSelector.tsx
@@ -69,7 +69,7 @@ export const RealmSelector = ({ realmList }: RealmSelectorProps) => {
 
   const dropdownItems = realmList.map((r) => (
     <DropdownItem
-      key={r.id}
+      key={`realm-dropdown-item-${r.realm}`}
       onClick={() => {
         setRealm(r.realm);
         setOpen(!open);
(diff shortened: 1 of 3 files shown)


src/components/realm-selector/RealmSelector.tsx
```diff
@@ -69,7 +69,7 @@ export const RealmSelector = ({ realmList }: RealmSelectorProps) => {
 
   const dropdownItems = realmList.map((r) => (
     <DropdownItem
-      key={r.id}
+      key={`realm-dropdown-item-${r.realm}`}
```
ssilvert, line 72: When you are logged into a realm other than master, the REST call to /auth/admin/realms does not return an id for the realm.

edewit (approved):

PR #198, opened 2020-10-29
540 user fed settings

srambach:
## Motivation
APPDUX-540 https://issues.redhat.com/browse/APPDUX-540

## Brief Description
This lays out the sections in the User Federation LDAP Settings tab. It is NOT tied to data nor are controls like switches and dropdowns operational. And, it is not the entire page - this is the contents of one of the tabs. 
**This is intended to lay out the page and establish the structure for future work.**

## Verification Steps
This can only be viewed in Storybook. Find it under "User Federation LDAP Settings Tab". It can be compared to this mockup: https://marvelapp.com/prototype/ecjc94b/screen/72071065

## Checklist:

- [ x] User-visible strings are using the react-i18next framework (useTranslation)
- [ x] Help has been implemented
- [ x] axe report has been run and resulting a11y issues have been resolved
- [ x] Formatting has been performed via prettier/eslint
- [ x] Type checking has been performed via 'yarn check-types'

## Additional Notes
image


src/user-federation/help.json
```diff
@@ -0,0 +1,43 @@
+{
+  "user-federation-help": {
+    "generalOptions": "General options",
+    "consoleDisplayNameHelp": "Display name of provider when linked in admin console",
+    "vendorHelp": "LDAP vendor (provider)",
+
+    "consoleDisplayConnectionUrlHelp": "Connection URL to your LDAP server",
+    "enableStartTlsHelp": "Encrypts the connection to LDAP using STARTTLS, which will disable connection pooling",
+    "useTruststoreSpiHelp": "Specifies whether LDAP connection will use the truststore SPI with the truststore confifured in standalone.xml/domain.sml. 'Always' means that it will always use it. 'Never' means that it will not use it. 'Only for ldaps' means that it will use it if your connection URL use ldaps. Note even if standalone.xml/domain.xml is not configured, the default java cacerts or certificate specified by 'javax.net.ssl.trustStore' property will be used.",
```
mfrances17, line 9: "configured"

src/user-federation/help.json
```diff
@@ -0,0 +1,43 @@
...
+    "connectionPoolingHelp": "Determines if Keycloak should use connection pooling for accessing LDAP server",
+    "connectionTimeoutHelp": "LDAP connection timeout in milliseconds",
+    "bindTypeHelp": "Type of the authentication method userd during LDAP bind operation. It is used in most of the requests sent to the LDAP server. Currently only 'none' (anonymous LDAP authentication) or 'simple' (bind credential + bind password authentication) mechanisms are available.",
```
mfrances17, line 12: "used"

src/user-federation/help.json
```diff
@@ -0,0 +1,43 @@
...
+    "bindDnHelp": "DN of the LDAP admin, which will be used by Keycloak to access LDAP server",
+    "bindCredentialsHelp": "Password of LDAP admin. This field is able to obtain its value from vault, use ${vault.ID} format.",
+
+    "editModeHelp": "READ_ONLY is a read-only LDAP store. WRITABLE means data will be s ynced back to LDAP on demand. UNSYNCED means user data will be imported, but not synced back to LDAP.",
```
mfrances17, line 16: "synced"

src/user-federation/help.json
```diff
@@ -0,0 +1,43 @@
...
+    "userObjectClassesHelp": "All values of LDAP objectClass attribute for users in LDAP divided by comma. For example: 'inetOrgPerson, organizationalPerson'. Newly created Keycloak users will be written to LDAP with all those object classes and existing LDAP user records are found just if they contain all those object classes.",
+    "userLdapFilterHelp": "Additional LDAP filter for filtering searched users. Leave this empty if you don't need additional filter. Make sure that it starts with '(' and ends with ')'",
+    "searchScopeHelp": "For one level, the search applies only for users in the DNs specified by User DNs. For subtree, the search applies to the whole subtree. See LDAP documentation for more details.",
+    "readTimeoutHelp": "LDAP read timeout in milliseconds. This timeout applies for LDAP read operations.",
+    "paginationHelp": "Does the LDAP server support pagination",
+
+    "importUsersHelp": "Import users",
+    "batchSizeHelp": "Count of LDAP users to be imported from LDAP to Keycloak within a single transaction",
+    "periodicFullSyncHelp": "Whether periodic full synchroniation of LDAP users to Keycloak should be enabled or not",
```
mfrances17, line 29: "synchronization"

src/user-federation/help.json
```diff
@@ -0,0 +1,43 @@
...
+    "periodicChangedUsersSyncHelp": "Whether periodic synchronization of changed or newly created LDAP users to Keycloak should be enabled or not",
+
+    "allowKerberosAuthenticationHelp": "Enable/disable HTTP authentication of users with SPNEGO/Kerberos tokens. The data about authenticated users will be provisioned from this LDAP server",
+    "useKerberosForPasswordAuthenticationHelp": "User Kerberos login module for authenticate username/password against Kerberos server instead of authenticating against LDAP server with Directory Service API",
+
+    "cachePolicyHelp": "Cache Policy for this storage provider.  'DEFAULT' is whatever the default settings are for the global cache.  'EVICT_DAILY' is a time of day every day that the cache will be invalidated.  'EVICT_WEEKLY' is a day of the week and time the cache will be invalidated.  'MAX_LIFESPAN' is the time in milliseconds that will be the lifespan of a cache entry.",
```
mfrances17, line 35: Search for all double spaces after period like these and make single space.

src/user-federation/messages.json
```diff
@@ -2,54 +2,60 @@
...
-    "loginTheme": "Login theme",
-    "consentRequired": "Consent required",
-    "searchForClient": "Search for client",
     "userFederationExplanation": "Keycloak can federate external user databases. Out of the box we have support for LDAP and Active Directory.",
     "getStarted": "To get started, select a provider from the list below.",
     "providers": "Add providers",
     "addKerberos": "Add Kerberos providers",
-    "addLdap": "Add LDAP providers"
+    "addLdap": "Add LDAP providers",
+    "Ldap": "LDAP",
```
mfrances17, line 10: "ldap" will never be translated, it's still ldap

mfrances17 (changes requested):
Found some typos/misspellings... easy fixes. 

Looking at the marvel spec and the storybook, it looks like there are some discrepancies that probably make sense to discuss rather than mark up. I'm not sure how closely we are trying to match the marvel spec. For example, a lot of the controls like the help icons are gray in the spec, not black, and probably should be gray so they don't stand out as much. 

Also, many of the fields have default entries that are missing in the implementation (Select a mode, LDAP user DN, etc). I'm also having trouble with storybook in that none of the fields are actionable (e.g. radios/toggles/dropdowns aren't working, so I can't verify values in the dropdown). I'm not convinced of the value for using storybook as opposed to just a screen cap.

src/user-federation/messages.json
srambach: Not even in languages with different alphabets or those that are non-phonetic? (I don't think this is actually the right term) Are you saying to leave "LDAP" in the code?

srambach:
@mfrances17 This is just getting the fields there with the help text. For one thing, I am assuming that the values for at least some dropdowns are probably coming from data somewhere but idk. My next step was to get placeholder text, default values, and dropdown options in.
As for the help icons, I need to contact the design team because I am using the PF help icon here. I want to check if this is a conscious choice or result of changes over the past year. Regardless, those are all using the HelpItem component, so it's one change to fix it globally if so. (in a separate PR probably)

src/user-federation/messages.json
ssilvert: @srambach `LDAP` will never be localized.  So, yes, you can just leave it in the code.

ssilvert (changes requested):
This looks good.  I think if you just put LDAP back in the code then we can merge.

There is one other thing that is broken though.  The scroll spy doesn't actually work.  It jumps OK, but it should always stay visible when you scroll to the bottom.  Optional if you want to fix that in this PR or just log an issue.

srambach:
> This looks good. I think if you just put LDAP back in the code then we can merge. Looks like you address all of @mfrances17 other issues, right?
So even though it's a display string, it should be "LDAP" in the code? That's what you both are saying?
> 
> There is one other thing that is broken though. The scroll spy doesn't actually work. It jumps OK, but it should always stay visible when you scroll to the bottom. Optional if you want to fix that in this PR or just log an issue.
I noticed that but it's broken elsewhere as well. I'll log an issue.

srambach:
> There is one other thing that is broken though. The scroll spy doesn't actually work. It jumps OK, but it should always stay visible when you scroll to the bottom. Optional if you want to fix that in this PR or just log an issue.

Actually, there's an issue to take the PatternFly component when it becomes available so that seems sufficient to me.

mfrances17:
> So even though it's a display string, it should be "LDAP" in the code? That's what you both are saying?
Yes. Terms like LDAP won't not localized, they are the same in every language. So it doesn't make sense to add them.

srambach:
OK, LDAP is removed from the strings. It was actually not used after I removed the header for later anyway.
@mfrances17 @ssilvert

mfrances17:
I noticed the scrolling issue as well. This IS actually working correctly in Client Details, when you scroll it bumps down.

client-details

mfrances17 (approved):
LGTM - I think it's fine to just merge this in as a known WIP while work continues on it.

ssilvert (approved):

PR #205, opened 2020-11-03
added validated property to fields instead of only on FormGroup

edewit:
## Motivation
This will make the fields look nicer / indicate there is an error


ssilvert (approved):
LGTM

PR #208, opened 2020-11-04
User Federation card view implementation

mfrances17:
## Motivation
Related JIRA/Github issues:
https://issues.redhat.com/browse/APPDUX-538
https://github.com/keycloak/keycloak-admin-ui/issues/146

Design spec:
https://marvelapp.com/prototype/ecjc94b/screen/72071045

## Brief Description
Adds a card view for the user federation databases. Utilizes the user federation empty state if no user federation databases exist. Preliminary delete card functionality is working, but there should be a confirm dialog TBD. Create new provider functionality (aka button in header, LDAP & Kerberos) is not part of this PR.

## Verification Steps
Creating providers is not yet available in the new UI, so if they do not exist on your locally running server, you'll need to create them with the old admin UI (aka the one running on localhost:8180)

**In the old admin console:**
1. Go to Configure > User Federation.
2. In the Add provider dropdown, select either kerberos or ldap.
3. Enter values for all required (*) fields and then click Save. For the purposes of testing, you can just enter dummy string values, they aren't verified.
4. Repeat steps 1-3 to create however many providers you want to test.

**In the new admin UI:**
1. Go to Configure > User federation.
2. Verify that each of the providers you created in steps 1-4 now appear as cards in the card view.
3. Choose a card and select Delete from the card menu. Verify that a notification appears to say that the card was successfully deleted, and the card no longer appears in the card view.
4. Delete each of the cards from the view. Verify that the empty state view appears instead of the card view.

## Checklist:

## Additional Notes

**User fed DBs exist:**
uf-populated

**User fed DBs empty:**
uf-empty-state

**Delete user fed DB/card:**
uf-delete


src/user-federation/UserFederationCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+  configEnabled,
+}: UserFederationCardProps) => {
+  const { t } = useTranslation();
+  const [isDropdownOpen, setDropdownOpen] = useState(false);
+  const onDropdownToggle = () => {
+    setDropdownOpen(!isDropdownOpen);
+  };
+
+  return (
+    <>
```
edewit, line 36: you don't need the extra "Fragment" tag here right?

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,176 @@
...
-    title: t("common:learnMore"),
-    href: "http://google.com",
+  const [userFederations, setUserFederations] = useState<
+    UserFederationRepresentation[]
+  >();
+  const { addAlert } = useAlerts();
+
+  const loader = async () => {
+    const testParams: { [name: string]: string | number } = {
+      parentId: `${realm}`,
```
edewit, line 41:
```suggestion
      parentId: realm,
```

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,176 @@
...
+                  `/admin/realms/${realm}/components/${userFederation.id}`
+                )
+                .then(() => loader());
+              addAlert(
+                t("userFederation:userFedDeletedSuccess"),
+                AlertVariant.success
+              );
+            } catch (error) {
+              addAlert(
+                `${t("userFederation:userFedDeleteError")} ${error}`,
```
edewit, line 89: if you change the message bundle to have a param, like this:
```
"userFedDeleteError": "Could not delete user federation: '{{error}}'"
```
then you can change this to:
```suggestion
                t("userFedDeleteError", {error})},
```
also no need to prefix the namespace `userFederation` as that is the default

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,176 @@
...
   return (
     <>
-      <ViewHeader
-        titleKey="user-federation:userFederation"
-        subKey={"user-federation:descriptionLanding"}
-        subKeyLinkProps={linkArgs}
-      />
-      <PageSection>
+      <PageSection variant={PageSectionVariants.light}>
+        <Title headingLevel="h3" size={TitleSizes["2xl"]}>
```
edewit, line 116: you should use the `<ViewHeader` here as it has some logic for showing / hiding page explanation

src/user-federation/messages.json
```diff
@@ -57,5 +57,19 @@
...
+    "userFederationInfo": "Keycloak can federate external user databases. Out of the box we have support for LDAP and Active Directory.",
+    "descriptionLanding": "This is the description for the user federation landing page",
+    "getStarted": "To get started, select a provider from the list below.",
+    "learnMore": "Learn more",
+    "addNewProvider": "Add new provider",
+    "providers": "Add providers",
+    "addKerberos": "Add Kerberos providers",
+    "addLdap": "Add LDAP providers",
+    "userFedDeletedSuccess": "The user federation has been deleted",
+    "userFedDeleteError": "Could not delete user federation:",
```
edewit, line 71:
```suggestion
    "userFedDeleteError": "Could not delete user federation: '{error}'",
```

edewit (changes requested):
Looking good nice PR

src/user-federation/UserFederationSection.tsx
srambach: It also handles the "learn more" link. Can we refactor the ViewHeader to allow additional contents like the dropdown toggle here to be included?

src/user-federation/user-federation.css
```diff
@@ -0,0 +1,8 @@
+.keycloak__user-federation__dropdown {
+    margin-top: 24px;
+    width: 20%;
```
srambach, line 3: Specifying width doesn't seem to be necessary?

src/user-federation/user-federation.css
```diff
@@ -0,0 +1,8 @@
+.keycloak__user-federation__dropdown {
+    margin-top: 24px;
```
srambach, line 2: Let's never use pixels. This is a standard spacer `var(--pf-global--spacer--lg)`

src/user-federation/user-federation.css
```diff
@@ -0,0 +1,8 @@
...
+    width: 20%;
+  }
+
+  .keycloak__user-federation__provider-label {
+    margin-left: 24px;
```
srambach, line 7: Same here, let's never use pixels. This is a standard spacer var(--pf-global--spacer--lg)

src/user-federation/UserFederationCard.tsx
mfrances17: Good catch, removed it. Thanks.

src/user-federation/UserFederationSection.tsx
mfrances17: Fixed, ty.

src/user-federation/UserFederationSection.tsx
mfrances17: ah, that's nice, i changed it accordingly. also merged my translation strings with existing ones, there were some dups, i temporarily created my own section in the messages to ease the initial merge.

src/user-federation/UserFederationSection.tsx
mfrances17: the problem with ViewHeader is that it does not allow me to add the Add New Provider dropdown to the header, i attempted but failed, so i created my own header

src/user-federation/messages.json
mfrances17: fixed, thanks

srambach:
Looks good other than the few comments. I did notice one thing that I'm not sure about - if I add a provider in the other console, then refresh the page in the new console, it doesn't update until I switch off the page and back again.

srambach:
Oh, also - I guess there's no sorting to these? I don't see any in the old console and honestly, no rhyme or reason to the order.

mfrances17:
> Looks good other than the few comments. I did notice one thing that I'm not sure about - if I add a provider in the other console, then refresh the page in the new console, it doesn't update until I switch off the page and back again.

hmmm, that's different than the behavior i'm seeing. if you refresh the page, the new ones appear.

mfrances17:
> Oh, also - I guess there's no sorting to these? I don't see any in the old console and honestly, no rhyme or reason to the order.

there wasn't anything in the spec about that, it could be added easily enough in a future PR.

srambach:
The delete alert just says "The user federation.." Can we add the name of what was deleted?

src/user-federation/user-federation.css
mfrances17: old habits die hard... fixed

src/user-federation/user-federation.css
mfrances17: yes, and looks better without it... at some point i needed to do this. removed.

src/user-federation/user-federation.css
mfrances17: used the var instead, thanks!

mfrances17:
> The delete alert just says "The user federation.." Can we add the name of what was deleted?

Yes, we could easy enough, but i'm staying consistent with what the rest of the UI is currently doing. When you delete a client, realm role, or anything else that's been implemented so far, it displays a similar message. So if we do that, we should do it across the board.

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,170 @@
...
-            {t("userFederationExplanation")}
-          </Text>
-          <Text component={TextVariants.p}>{t("getStarted")}</Text>
+          {t("userFederationExplanation")}
+          <Button
+            variant="link"
+            icon={<ExternalLinkAltIcon />}
+            iconPosition="right"
+            component="a"
+            href="http://www.google.com"
```
jenny-s51, line 120: Currently "learn more" redirects the user to the home page of Google. While this presents a lot of creative freedom (and 100% made my day 😂) we should probably change this to something like [https://www.keycloak.org/docs/latest/server_admin/index.html#_user-storage-federation](https://www.keycloak.org/docs/latest/server_admin/index.html#_user-storage-federation)

jenny-s51 (commented):
Behavior is looking great! Just one comment from me 😄

jenny-s51 (approved):

src/user-federation/UserFederationSection.tsx
mfrances17: thanks for the better URL 😄  was hoping this would come from an annotation from the spec, so the google was a TBD, but i like your solution better. also added a target because i don't think linked pages should replace the UI in the current window.

src/user-federation/UserFederationCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+  CardBody,
+  CardFooter,
+  Dropdown,
+  KebabToggle,
+  Label,
+} from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+
+export type UserFederationCardProps = {
+  title: string;
```
dlabaj, line 16: We should start documenting the props.

src/user-federation/UserFederationCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+  configEnabled,
+}: UserFederationCardProps) => {
+  const { t } = useTranslation();
+  const [isDropdownOpen, setDropdownOpen] = useState(false);
+  const onDropdownToggle = () => {
+    setDropdownOpen(!isDropdownOpen);
+  };
+
+  return (
+    <Card>
```
dlabaj, line 36: Will we need cards elsewhere ... if so could this be a more generic component moved under components?

src/user-federation/UserFederationCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const UserFederationCard = ({
+  dropdownItems,
+  title,
+  providerId,
+  configEnabled,
+}: UserFederationCardProps) => {
+  const { t } = useTranslation();
+  const [isDropdownOpen, setDropdownOpen] = useState(false);
```
dlabaj, line 30: would call it setIsDropdownOpen

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,171 @@
...
+    const result = await httpClient.doGet<UserFederationRepresentation[]>(
+      `/admin/realms/${realm}/components`,
+      {
+        params: testParams,
+      }
+    );
+    setUserFederations(result.data);
   };
 
+  useEffect(() => {
```
dlabaj, line 53: What's the reason for the empty array as the second parameter?

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,171 @@
...
+    loader();
+  }, []);
+
+  const [providerOpen, isProviderMenuOpen] = useState(false);
```
dlabaj, line 57: isProviderMenuOpen and setIsProviderMenuOpen

src/user-federation/UserFederationSection.tsx
dlabaj: Can we update ViewHeader to add this ability add a dropdown?

dlabaj (changes requested):
A few comments otherwise looks good.

src/user-federation/UserFederationSection.tsx
edewit: I don't see why we couldn't add that to the view header we might need it on other pages as well, could do something like this:
```html
<ViewHeader ...>
   <Dropdown
          className="keycloak__user-federation__dropdown"
          toggle={
            <DropdownToggle
              onToggle={() => isProviderMenuOpen(!providerOpen)}
              isPrimary
              id="ufToggleId"
            >
              {t("addNewProvider")}
            </DropdownToggle>
          }
          isOpen={providerOpen}
          dropdownItems={ufAddProviderDropdownItems}
        />
</ViewHeader>
```
what do you think?

mfrances17: Yeah as the 3 of you point out, I think you are all right, it should probably be added to the ViewHeader component... at the time I thought it might be too specific to my use case. But there may be other times when it is needed. I'll rework it so it's part of the component. Thanks.

src/user-federation/UserFederationSection.tsx
mfrances17: Not sure, but i thought the same thing because i saw it this way in other parts of the code. But when I remove it, the page never loads. Add it back, and it loads fine.

src/user-federation/UserFederationCard.tsx
mfrances17: fixed, thanks

src/user-federation/UserFederationSection.tsx
mfrances17: fixed, thanks

src/user-federation/UserFederationSection.tsx
mfrances17: i've modified the ViewHeader component accordingly, and used that component inside of the user fed card view.

src/user-federation/UserFederationCard.tsx
mfrances17: Created a generic keycloak card component and used that instead of a user fed card

ssilvert (changes requested):
- Need a confirmation dialog when you delete a fed provider.
- "Add new provider" dropdown should be hidden when in empty state.  You can add by clicking one of the cards.
- When in empty state, need the "To get started..." text as shown in the wireframe.

dlabaj (approved):

mfrances17:
> * Need a confirmation dialog when you delete a fed provider.
> * "Add new provider" dropdown should be hidden when in empty state.  You can add by clicking one of the cards.
> * When in empty state, need the "To get started..." text as shown in the wireframe.

Latest commit handles both of your empty state suggestions.

mfrances17:
Added a confirm dialog for the card delete action in the latest commit. All PR review comments should be addressed at this point. Screen shots of confirm:

uf-delete-confirm

uf-delete-msg

ssilvert (approved):

src/user-federation/UserFederationSection.tsx
```diff
@@ -1,68 +1,181 @@
...
+        <GalleryItem key={index}>
+          <KeycloakCard
+            id={userFederation.id}
+            dropdownItems={ufCardDropdownItems}
+            title={userFederation.name}
+            footerText={
+              userFederation.providerId === "ldap" ? "LDAP" : "Kerberos"
+            }
+            labelText={
+              userFederation.config.enabled
```
ssilvert, line 115:
```suggestion
              userFederation.config.enabled[0] === "true"
```

ssilvert (changes requested):

ssilvert (approved):

PR #209, opened 2020-11-05
removed "Home" from navigation

edewit:
only show breadcrumb when there are 2 levels 
renamed the list clients and client scopes

[see marvel design](https://marvelapp.com/prototype/4bi07ca/section/1014937)


mfrances17 (approved):
LGTM... code looks good and it tested successfully.

PR #215, opened 2020-11-12
Update route-config.ts

mounikakella:
Fixed error "`path` must be provided in every route object" in console

## Motivation
Im trying to run this locally and got an error from withBreadCrumb: `path` must be provided in every route object and its breaking to load non existing route

## Brief Description
So the error is saying no path should be empty. so to handle that I have added * to show user not found page if the route does not exist.

## Additional Notes
Before:
Screenshot from 2020-11-12 11-43-22
After:
Screenshot from 2020-11-12 11-43-54

### DIFF
--- a/src/route-config.ts
+++ b/src/route-config.ts
@@ -158,7 +158,7 @@ export const routes: RoutesFn = (t: TFunction) => [
     access: "anyone",
   },
   {
-    path: "",
+    path: "*",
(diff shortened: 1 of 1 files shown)


ssilvert (approved):

ssilvert:
Thanks for the PR @mounikakella !

PR #217, opened 2020-11-12
Adds realm role details tab to storybook

jenny-s51:
## Motivation

Closes https://issues.redhat.com/browse/APPDUX-652. The purpose of this issue was to build the layout/UI of the realm role details tab. This PR includes the new UI component and adds a demo of the new component to Storybook.

## Checklist:

## Additional Notes

Screen Shot 2020-11-12 at 2 37 33 PM


mfrances17 (changes requested):
Ran the storybook locally. It looks like the spacing is a bit off compared to the design spec. There should be more space between the top of the tab and Role name, and less space between the bottom of the description field and the action buttons. Compare to:

spec-realm-roles

src/route-config.ts
```diff
@@ -97,6 +98,12 @@ export const routes: RoutesFn = (t: TFunction) => [
     breadcrumb: t("roles:roleList"),
     access: "view-realm",
   },
+  {
+    path: "/roles/:id",
+    component: RolesForm,
+    breadcrumb: t("roles:roleDetails"),
+    access: "view-realm",
```
ssilvert, line 105: The code that processes `route-config.ts` needs to be able to handle multiple roles.  And we don't do anything with subroutes right now anyway so I'm not sure how we need to check them.  We don't have any of that yet, but I believe that you should still be able to code it like this:  
```suggestion
    access: ["view-realm", "view-users"],
```

ssilvert (changes requested):

src/route-config.ts
jenny-s51: Thank you for your helpful review comments @ssilvert ! this doesn't seem to work at the moment, I get `Forbidden` displayed on the page instead of the details tab. what do you suggest would be an acceptable fix in the meantime? Should I just leave `view-realm`?

ssilvert: > Thank you for your helpful review comments @ssilvert ! this doesn't seem to work at the moment, I get `Forbidden` displayed on the page instead of the details tab. what do you suggest would be an acceptable fix in the meantime? Should I just leave `view-realm`?

That's fine.  I'll just make an issue for it.

src/realm-roles/RealmRoleDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <PageSection variant="light">
+        <Tabs
+          activeKey={activeTab}
+          onSelect={(_, key) => setActiveTab(key as number)}
+          isBox
+        >
+          <Tab eventKey={0} title={<TabTitleText>{t("Details")}</TabTitleText>}>
+            <FormAccess
+              isHorizontal
+              role="manage-clients"
```
ssilvert, line 41: @jenny-s51 This should be "manage-realm"

ssilvert (changes requested):

src/realm-roles/RealmRoleDetails.tsx
jenny-s51: nice catch, updated - thank you

ssilvert (commented):
LGTM.  If Sarah agrees we can merge.

ssilvert (approved):

mfrances17 (approved):
LGTM, spacing issues I flagged previously are fixed.

PR #218, opened 2020-11-12
Fix disabled label for user fed card view

mfrances17:
## Motivation
https://github.com/keycloak/keycloak-admin-ui/issues/213

## Brief Description
Fixes three issues:
- Return from REST call to get enabled/disabled was not handled correctly as an array
- Disabled label was displaying as blue instead of gray as per spec
- Add new provider dropdown was not populating with items once separate card component was created 

## Verification Steps
In the legacy Keycloak Admin Console UI:
1. Select Configure > User Federation and select kerberos or ldap from the Add provider dropdown.
2. Enter the information for a new provider... just use placeholder text in the required fields, it's not verified.
3. Repeat steps 1-2 to create a second card, but this time, make sure to set the Enabled toggle to OFF.

In the new Keycloak Admin Console:
1. Verify that one of the cards is labeled correctly as Enabled and the other as Disabled.
2. Verify that the color for the disabled label is gray and the enabled label is blue.
3. Select the Add new provider dropdown and verify that the LDAP and Kerberos items appear. Note that these are not functional as of yet, they are placeholders commands only.

## Checklist:

## Additional Notes
Screen cap showing all 3 fixes:
three-userfed-fixes

### DIFF
--- a/src/components/keycloak-card/KeycloakCard.tsx
+++ b/src/components/keycloak-card/KeycloakCard.tsx
@@ -55,7 +55,10 @@ export const KeycloakCard = ({
       <CardFooter>
         {footerText && footerText}
         {labelText && (
-          <Label color={labelColor || "gray"} className="keycloak__keycloak-card__footer-label">
+          <Label
+            color={labelColor || "gray"}
+            className="keycloak__keycloak-card__footer-label"
+          >
             {labelText}
           </Label>
         )}
--- a/src/components/keycloak-card/KeycloakCard.tsx
+++ b/src/components/keycloak-card/KeycloakCard.tsx
@@ -10,7 +10,6 @@ import {
   KebabToggle,
   Label,
 } from "@patternfly/react-core";
-// import { useTranslation } from "react-i18next";
 import "./keycloak-card.css";
 
 export type KeycloakCardProps = {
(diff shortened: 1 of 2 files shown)


edewit (approved):
:+1:

jenny-s51 (commented):
Label colors are looking good! 👍 Just wanted to bring to your attention that the alignment of the labels looks a bit different in Chrome than it does in Firefox:

Screen Shot 2020-11-13 at 3 22 16 PM

mfrances17:
Looks like an issue with the cards in FF, they look much smaller in width.

mfrances17:
Actually, I'm not seeing that issue in FF, even when resizing:
ff-userfed

srambach:
I can reproduce Jenny's screen if I set FF to "zoom text only." I'm not sure what the thinking is on the level of support for this. I think it would be a little better if we put the text and label into a `<Flex>` and `<FlexItem>`s and remove the style that's adding a left margin to the label.
If it's not acceptable to use the flex spacer (which is md but currently the custom style uses lg) then we'll need to adjust the `--pf-l-flex--spacer` locally.

jenny-s51 (approved):

mfrances17:
I've filed https://github.com/keycloak/keycloak-admin-ui/issues/223 as a separate issue to handle the card label alignment in Firefox.

PR #222, opened 2020-11-16
Add role mapping into the flow and filter the clients that don't have roles

edewit:
added a filter on the clients to only show the ones that have roles. Used `clientId` instead of `name` to show the clients in the dropdown and added a route so that it will use this screen if you use the "role mapping" in the dialog.


ssilvert (commented):
@edewit Need more detail about the PR in order to do a review.

ssilvert (approved):

PR #238, opened 2020-11-26
Initial version of the Service Account tab

edewit:
## Motivation

Doesn't have any actions yet just the basic list with a toggle function. [marvel design](https://marvelapp.com/prototype/f058374/screen/71931942)

## Brief Description

initial version of design

## Verification Steps

Go to client details when client is of type "service account" then you'll have an extra tab also called "service account"

## Checklist:

### DIFF
--- a/src/clients/ClientDetails.tsx
+++ b/src/clients/ClientDetails.tsx
@@ -103,11 +103,6 @@ export const ClientDetails = () => {
     name: "publicClient",
     defaultValue: false,
   });
-  const serviceAccount = useWatch({
-    control: form.control,
-    name: "serviceAccountsEnabled",
-    defaultValue: false,
-  });
 
   const { id } = useParams<{ id: string }>();
 
@@ -173,6 +168,7 @@ export const ClientDetails = () => {
         };
         await adminClient.clients.update({ id }, client);
         setupForm(client as ClientRepresentation);
+        setClient(client);
         addAlert(t("clientSaveSuccess"), AlertVariant.success);
       } catch (error) {
         addAlert(`${t("clientSaveError")} '${error}'`, AlertVariant.danger);
@@ -249,7 +245,7 @@ export const ClientDetails = () => {
               </Tab>
             </Tabs>
           </Tab>
-          {serviceAccount && (
+          {client && client.serviceAccountsEnabled && (
             <Tab
               eventKey={3}
               title={<TabTitleText>{t("serviceAccount")}</TabTitleText>}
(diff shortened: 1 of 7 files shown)


srambach:
I don't know if I'm doing something wrong, but I can only get the table toolbar and a spinner, no content on the page.

ssilvert:
> I don't know if I'm doing something wrong, but I can only get the table toolbar and a spinner, no content on the page.

@edewit  Same here.  There is an error in the console.

```
Uncaught (in promise) TypeError: adminClient.roles.getCompositeRolesForRealm is not a function
    at ServiceAccount.js:71
```

edewit:
You need to do an `yarn install` to get the latest version of the `keycloak-admin-client` that has the `getCompositeRolesForRealm` function

ssilvert:
> You need to do an `yarn install` to get the latest version of the `keycloak-admin-client` that has the `getCompositeRolesForRealm` function

@edewit I did that and yarn says `Already up-to-date.`  Looks like you need to update package.json.

edewit:
The package.json was updated in a separate PR #234, but I found the problem. You need to save the client before you go to the service account tab. Added a commit that makes the service account tab only show after save.

srambach (commented):
I still can't get the page to load even with your update, an install, and making sure the client is saved.

ssilvert:
Tab name should be called "Service account roles"

ssilvert:
@edewit  I'm going to break our rules and leave this open until Monday.  Normally, I would merge this PR and we can address the relatively minor issues raised by @srambach and myself separately.

However, we now have another option.  The latest PatternFly has a new Dual list selector:  https://www.patternfly.org/v4/components/dual-list-selector

It's possible that if this selector had been available then the design for this page would have ended up similar to what we have in the old console.  The nice thing is that this selector has searching, which I believe was one of the major reasons for the design change.

Regardless, we have a mandate to get this React version of admin to code complete as soon as possible.  So, if it's easier to implement like it's done in the old console then that's what we should do.  I suspect that it probably is because the REST API was designed with the old UI in mind.  Just judging by the number of REST calls required to render the old page versus the new page this looks like it's the case.

The new dual list selector is pretty nice and it does all the right things when it comes to keyboard navigation and such.  It seems to me that we could spend a lot of time perfecting the new style screens.  So if it's going to be easier to implement dual list screens like in the old console then we need to do that.

edewit:
@srambach you'll see the page when you create a new client and check the "Service Account" checkbox:
image

edewit:
As I see it the amount of requests needed to build this table is due to the fact that the old console doesn't display the client roles. Only when you select a client with the dropdown under the Dual list selector you can see the roles of one client.

If speed of development is a deciding factor, this page is working atm, if you want to use this Dual list selector we have to start over again.

edewit:
Just talked with the designers and this table layout was chosen because of better UXD not because the component wasn't available.

ssilvert (approved):

ssilvert:
Looks like cosmetic things from me and @srambach have been fixed.  Merging.

PR #241, opened 2020-11-30
Introduced generic DataList component that can handle all Lists.

edewit:
## Motivation

Instead of a lot of copy paste code in the various *List components create one that does it all.

## Brief Description

Introduces a new component called `DataList` that can render a paginated or non paginated lists. This includes searching and loading of the data.

## Checklist:


src/util.ts
```diff
@@ -29,8 +29,10 @@ const sortProvider = (
   }
 };
 
+export const deepCopy = (obj: object) => JSON.parse(JSON.stringify(obj));
```
ssilvert, line 32: I'm concerned about doing deep copy this way, especially in a generic utility where we don't know the nature of the object being cloned.

lodash is a transitive dependency in our application and it has a cloneDeep() method.  Maybe we could use that?

ssilvert (changes requested):
I like what you've done here.  We will need a lot more generic components like this if we are going to hit our deadlines.

From my experience, the problem you can run into is that you make a generic component that covers two or three use cases.  But then you try to use it in another section and you have to add more features to your generic component.  Then the same thing happens when you use it in the next section.  Before you know it, you've just recreated the PatternFly component.

I just want to make sure we don't run into that sort of thing here.  So please look at all the wireframes for the other places that this will potentially be used.

Many of our sections start with a list on the first page.  It would be really cool if we could use this to quickly implement the list on the first page of those sections.

edewit:
Right I know this pit fall all to well where you build something generic and keep adding to it to make it more and more generic. That's why only now I created this when we know the use cases and we have some code that is duplicated in 3 or more places. This component is for the List view / table views we have on multiple pages

ssilvert (approved):

PR #243, opened 2020-12-01
User Federation: populate Kerberos required and cache settings data

mfrances17:
## Motivation
Related JIRA issues:
https://issues.redhat.com/browse/APPDUX-655

Design spec:
https://marvelapp.com/prototype/ecjc94b/screen/72098368

## Brief Description
Fetches and displays settings data for a specific individual provider card that is clicked in the user federation card view. This PR does **not** cover save nor the showing/hiding of UI components based on selections. These will be handled in future PRs. This PR only covers pulling in the data and getting to the settings page from the cards.

## Verification Steps
1. Click a card in the user fed card view.
2. Verify that all settings match those that have been configured for that provider (presumably, the ones configured and saved in the old admin UI).
3. Change the values and verify that the changes are reflected in the new UI.

## Checklist:

## Additional Notes
Required settings:
kerberos-req-settings

Cache settings:
kerberos-cache-settings


src/components/bread-crumb/PageBreadCrumbs.tsx
```diff
@@ -15,8 +15,11 @@ export const PageBreadCrumbs = () => {
         <Breadcrumb>
           {crumbs.map(({ match, breadcrumb: crumb }, i) => (
             <BreadcrumbItem key={i} isActive={crumbs.length - 1 === i}>
-              {crumbs.length - 1 !== i && <Link to={match.url}>{crumb}</Link>}
-              {crumbs.length - 1 === i && <>{crumb}</>}
+              {crumbs.length - 1 !== i && match.url === "/components" ? (
+                <Link to="/user-federation">User federation</Link>
```
edewit, line 19: Why did you need this? I was hoping we could keep this component very generic and not introduce page specific logic

src/components/keycloak-card/KeycloakCard.tsx
```diff
@@ -26,17 +27,28 @@ export type KeycloakCardProps = {
...
+  const history = useHistory();
+
   const onDropdownToggle = () => {
     setIsDropdownOpen(!isDropdownOpen);
   };
 
+  const openSettings = () => {
+    history.push(`/components/${id}`);
+    // debugging
+    console.log(`Card ${title} was clicked!`);
```
edewit, line 48: you don't need this anymore right?

src/user-federation/KerberosSettingsCache.tsx
```diff
@@ -7,14 +7,67 @@ import {
...
+        setValue("cachePolicy", entry[1].cachePolicy);
+        setValue("evictionHour", entry[1].evictionHour);
+        setValue("evictionMinute", entry[1].evictionMinute);
+        setValue("evictionDay", convertToDays(entry[1].evictionDay[0]));
+        setValue("maxLifespan", entry[1].maxLifespan);
+      } else {
+        setValue(entry[0], entry[1]);
+      }
+    });
+    console.log(component);
```
edewit, line 60: can go?

src/user-federation/KerberosSettingsRequired.tsx
```diff
@@ -8,17 +9,55 @@ import {
...
-
+  const adminClient = useAdminClient();
   const [isEditModeDropdownOpen, setIsEditModeDropdownOpen] = useState(false);
-  const { register, control } = useForm<ComponentRepresentation>();
+  const { register, control, setValue } = useForm<ComponentRepresentation>();
+  const { id } = useParams<{ id: string }>();
+
+  const setupForm = (component: ComponentRepresentation) => {
+    Object.entries(component).map((entry) => {
+      if (entry[0] === "config") {
```
edewit, line 28: I guess you did this to work around the problem that the react-form-hook wants to create object when the keys have a "." right? If that is the case I have created an util function that converts the dots to "_" (e.g. `convertToFormValues`) and then you can use the config object keys like `config.claim_name`

edewit (changes requested):

src/components/bread-crumb/PageBreadCrumbs.tsx
mfrances17: Because the API is component, not 'user federation'... so the breadcrumb would be > Component > which means nothing to end users.

src/components/keycloak-card/KeycloakCard.tsx
mfrances17: I was going to leave in the debugging statements until the Save and component display/hide logic was added, which I would assume would be next sprint.

src/components/bread-crumb/PageBreadCrumbs.tsx
edewit: so why put `/components/` in the path? Or am I missing something?

src/components/keycloak-card/KeycloakCard.tsx
edewit: okay

src/components/bread-crumb/PageBreadCrumbs.tsx
mfrances17: The breadcrumb is supposed to be > User federation > ... how else would this be achieved? Why is this a problem?

src/user-federation/KerberosSettingsRequired.tsx
mfrances17: Is that solution any better than this one?

src/components/keycloak-card/KeycloakCard.tsx
```diff
@@ -26,17 +27,28 @@ export type KeycloakCardProps = {
...
+    console.log(`Its ID is: ${id}`);
+  };
+
   return (
     <Card>
```
jenny-s51, line 53: When I tested this, I initially tried to click on the console display name (kerberos, ldap) to get to the settings, which didn't take me there. Instead had to click on the center of the card. I think it might make sense to make these cards selectable/hoverable and pass in an `onClick` as it's done here to make it clearer for the user: https://www.patternfly.org/v4/components/card#selectable-and-selected

jenny-s51 (changes requested):
This is looking great @mfrances17 ! 👍 Fields are populating as expected. Would be awesome if we could make the cards selectable before merging (see comment below).

src/components/bread-crumb/PageBreadCrumbs.tsx
edewit: right so we can just remove `/componnents` in the route-config right?

mfrances17: i don't think so, no. this is the route-config:
```
  {
    path: "/components/:id",
    component: UserFederationKerberosSettings,
    breadcrumb: t("common:settings"),
    access: "view-realm",
  },

```
you can't change it to /user-federation/:id, because that URI doesn't exist. the rest call needs to be /components/:id. unless I am totally missing something?

mfrances17: it does need to change, though... clicking 'components' in the breadcrumb will fail but id will pass. so i'll try to figure this out.

src/components/keycloak-card/KeycloakCard.tsx
mfrances17: Thanks, I originally had attempted to do something similar but any click on the card kept sending it to details, including a menu click which is obviously not what we want. So I'll need to take another look.

src/components/bread-crumb/PageBreadCrumbs.tsx
edewit: The `path` in the route config does not have to match the path in the backend call it can be anything. Here you configure the path that the browser uses and also the hierarchy of our breadcrumbs.

edewit: I create a PR on you PR with the change that I mean

src/user-federation/KerberosSettingsRequired.tsx
```diff
@@ -8,17 +9,55 @@ import {
...
+          entry[1].allowPasswordAuthentication[0] === "true"
+        );
+        setValue(
+          "updateProfileFirstLogin",
+          entry[1].updateProfileFirstLogin[0] === "true"
+        );
+        setValue("editMode", entry[1].editMode);
+        setValue("kerberosRealm", entry[1].kerberosRealm);
+        // MF TODO - these two props do not seem to appear on new UI like old UI
+        // setValue("enabled", entry[1].enabled);
```
srambach, line 43: Enabled is a switch at the top in the header https://marvelapp.com/prototype/ecjc94b/screen/72098368
I don't know about priority. I just asked on the marvel.

src/components/bread-crumb/PageBreadCrumbs.tsx
mfrances17: This is perfect, thanks for the help @edewit. I'll make these changes. Feeling a bit stupid for not figuring that out.

src/user-federation/KerberosSettingsRequired.tsx
mfrances17: @srambach yes i saw it in the old UI but was wondering why it was left out of new UI, whether it was a design change or reason why you left it out

src/components/keycloak-card/KeycloakCard.tsx
mfrances17: i've made the appropriate changes so the entire card can be clicked and the card menu still works as expected

src/components/bread-crumb/PageBreadCrumbs.tsx
mfrances17: changes are in the latest commit

src/user-federation/KerberosSettingsRequired.tsx
srambach: @mfrances17 comments here from design; priority is intended to move to a different place but TBD whether it can in this iteration. Haley's going to ask at the Weds meeting. https://marvelapp.com/prototype/ecjc94b/screen/72098368/comment/55c7d99d-7e0a-477b-bdee-f975bc160272?

src/user-federation/KerberosSettingsRequired.tsx
mfrances17: Ah, I've got this working with your utility now. Will have an update soon.

mfrances17: This is working with the utility in the latest version

src/components/bread-crumb/PageBreadCrumbs.tsx
```diff
@@ -15,8 +15,7 @@ export const PageBreadCrumbs = () => {
         <Breadcrumb>
           {crumbs.map(({ match, breadcrumb: crumb }, i) => (
             <BreadcrumbItem key={i} isActive={crumbs.length - 1 === i}>
-              {crumbs.length - 1 !== i && <Link to={match.url}>{crumb}</Link>}
-              {crumbs.length - 1 === i && <>{crumb}</>}
+              <Link to={match.url}>{crumb}</Link>
```
edewit, line 18: This is my fault, sorry, but we don't need any change to the `PageBreadCrumb` right. I might have incorrectly reverted the change

jenny-s51 (approved):
Nice work @mfrances17 ! Card view looks great.

src/components/bread-crumb/PageBreadCrumbs.tsx
mfrances17: @edewit no worries, yeah we wound up not needing any changes to breadcrumb. i reverted it back to what it was originally in the latest commit.

edewit (approved):

PR #246, opened 2020-12-04
Use new patternfly jumplinks.

ssilvert:
## Motivation
ScrollForm should use new patternfly JumpLinks instead of our home-grown logic.

https://www.patternfly.org/v4/components/jump-links

## Brief Description
* Upgraded @patternfly packages.
* Refactored ScrollForm

## Verification Steps
Go to client details.  Links on the right side of the page should work the same as before.

## Checklist:

## Additional Notes
At some point we should also implement the scroll spy capability of JumpLinks.  I tried it out and couldn't get it to work right away with the Grid layout we are currently using in ScrollForm.  The example code for JumpLink uses PageSections so maybe if we converted to that it would work.  But that can be a task for another day.


src/components/scroll-form/ScrollForm.tsx
```diff
@@ -1,5 +1,12 @@
 import React, { Children, useEffect, useState } from "react";
```
edewit, line 1: `useEffect` can be removed

edewit (approved):
:+1: always good to see so much code removed

Although the new doesn't indicate the active section when scrolling down like the old did

ssilvert:
> 👍 always good to see so much code removed
> 
> Although the new doesn't indicate the active section when scrolling down like the old did

Yeah, I couldn't get the scroll spy feature working but I didn't spend much time on it. Sarah says she thinks she knows how to do it.  It's a beta component after all.  I'm confident we will get it working.

src/components/scroll-form/ScrollForm.tsx
ssilvert: Oops.  Missed this comment.  I'll submit another PR to fix it.

PR #247, opened 2020-12-04
Realm role attributes

jenny-s51:
## Motivation

Towards https://issues.redhat.com/browse/APPDUX-679

## Checklist:


src/realm-roles/RoleAttributes.tsx
```diff
@@ -39,22 +71,28 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
...
-          </Button>
-          <Button variant="link" onClick={() => history.push("/roles/")}>
-            {t("common:reload")}
-          </Button>
-        </ActionGroup>
-      </FormAccess>
-    </>
+        <TableComposable className="keyValueTable" aria-label="Table text">
+          <Thead>
+            <Tr className="labels">
```
jenny-s51, line 76: @srambach it looks like there's a bit more space than there should be between the labels and the input fields... do you think there is a good CSS fix to remedy this?

edewit (changes requested):
Why do we need `src/realm-roles/RealmRoleDetails.tsx`? I thought we changed the `RealmRoleForm.tsx` so that we could use one page in both cases? Now you introduce this page and we have 2 again.

src/realm-roles/RealmRolesSection.css
```diff
@@ -0,0 +1,17 @@
+.kc-role-attributes__table {
+  /* even though the table is borderless, make the border under the th transparent */
+  --pf-c-table--BorderColor: transparent;
```
srambach, line 3: You can remove the BorderColor and instead use
  --pf-c-table--border-width--base: 0;

That gets rid of one extra pixel between the header and body of the table.

src/realm-roles/RoleAttributes.tsx
srambach: I left a comment on the css file - There was just one pixel in the transparent border of the table. There's nothing else there, so I think it's just the line height making it appear that there's more.

src/realm-roles/RoleAttributes.tsx
```diff
@@ -0,0 +1,93 @@
...
+    [
+      <TextInput />,
+      <TextInput />,
+      <Button
+        id="plus-icon"
+        variant={ButtonVariant.link}
+        tabIndex={-1}
+        className="kc-role-attributes__plus-icon"
+      >
+        <PlusCircleIcon />
```
srambach, line 40: These should have aria labels. (You'll need to i118n the text for the button) Also, should the ids be prefixed with kc? I'm not sure of our coding convention here. I'd personally make them more descriptive to include "label" since those are identifying the labels, not the actual keys and values.
      <TextInput aria-labelledby="key" />,
      <TextInput aria-labelledby="value" />,
      <Button
        id="plus-icon"
        .......
        aria-label="Add an attribute"
      >

edewit:
@jenny-s51 seems like adding of the file is due to a bad rebase? would it make sense to create a new PR for this?

edewit (approved):

PR #251, opened 2020-12-08
Updated DataTable to be capable of select

edewit:
## Motivation

For some table views we need to be able to select multiple rows, so added this to the data table

## Brief Description

Also updated the group list to use this new feature.

## Verification Steps

Go to the groups section and see that it works like before.

## Checklist:


srambach:
Axe is reporting a critical error on the table headings that don't have text (above the selection boxes and action menus) Scope should be "column" I think and perhaps we need some screen reader only text to say what that column is?

edewit:
I don't know how / what I would need to fix it looks like this: `<th scope="col" class="" data-label="Group name" data-key="1">Group name</th>` would that be something that Patternfly needs to fix?

mfrances17 (approved):
LGTM, code looks good and the selection within the datalist works as expected.

PR #252, opened 2020-12-09
initial version of the user list

edewit:
## Motivation

with the new "DataList" (will change name in another PR) this user list is easy

## Verification Steps

See user section of the app now has a searchabe user list

## Checklist:

### DIFF
--- a/src/user/UsersSection.tsx
+++ b/src/user/UsersSection.tsx
@@ -4,9 +4,11 @@ import {
   AlertVariant,
   Button,
   ButtonVariant,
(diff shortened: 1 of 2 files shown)


ssilvert (approved):
This is good.  I'll merge it.

Just FYI, the user design is not quite done yet but should be very soon.  Users are a little strange.  They potentially come from multiple data sources and can number in the millions.  That's why in the old console we start off with an empty list.  We have been planning to improve it, especially in cases where we know there aren't so many users.

PR #254, opened 2020-12-11
Makes jumplinks work

srambach:
## Motivation
This connects up jumplinks and also pulls in a new Patternfly to get a fix to the JumpLinks component.

## Brief Description
This sets the scrollContainer so that the jumplinks component knows what to sync with. Currently, the component does not handle spaces in ids, so spaces are changed to hyphens. CSS changed to remove module css and use the BEM naming standard.

## Verification Steps
Jumplinks can be seen on the client settings page. Scroll the page and click the jump links.
Note that the highlight is based on what section is at the top of the page - so if the section clicked is far enough down that the page cannot scroll it to the top, the section highlighted will not be what was just clicked. However, user focus will be placed on the section that was selected.

## Checklist:

## Additional Notes
Note that tabOrder is put onto the section headings in order for the jump link to throw focus to it. Although this is usually not a great idea since a heading is not an interactive element, this is deemed better for screen readers so that the next thing read is the section that you jumped to. It also prevents a bug showing two highlighted sections in some cases. https://github.com/patternfly/patternfly/issues/3747

Finally, I have no idea why one line in jsonFileUpload.test.txs.snap changed. It may be related to the PatternFly update??


package.json
```diff
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "^4.65.6",
-    "@patternfly/react-core": "4.79.2",
+    "@patternfly/react-core": "4.83.3",
```
edewit, line 20: Could you also add an updated `yarn.lock` file?

edewit (changes requested):
I had the problem that when I click the link the active link was the previous only after scrolling down a little bit the clicked link would be active. I created a movie:

issue

package.json
srambach: @edewit I can't repeat this, can you? The whole concept is a little fragile imho and a little odd when it can't scroll all the way to the target, so it highlights something you didn't click on. 

If we move the scroll id from the heading to the container it scroll to a slightly different spot because of the padding on the panel, but then the focus has to go to the entire container, and I thought that overall, to the heading was a better experience.

edewit (approved):

PR #256, opened 2020-12-11
initial version of the events tab

edewit:
## Motivation

using the new list component quick implementation of the events tab list

## Brief Description

new events tab

## Verification Steps

first go to the old console and turn on event logging then create some login event and failed login events
then check the events tab section to see these events.

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] User-visible strings are using the react-i18next framework (useTranslation)
- [x] Help has been implemented
- [ ] axe report has been run and resulting a11y issues have been resolved
- [ ] Unit tests have been created/updated
- [x] Formatting has been performed via prettier/eslint
- [x] Type checking has been performed via 'yarn check-types'

## Additional Notes

image

PR #258, opened 2020-12-11
Revert "initial version of the events tab"

mfrances17:
Reverts keycloak/keycloak-admin-ui#256 - looks like this is breaking the build. This may have a dependency on the other in-progress PR, not sure.


edewit:
fixed build with #6ae437c

PR #264, opened 2020-12-16
Adds functionality to attributes tab of realm roles section

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-563

## Brief Description

Adds functionality for realm role attributes tab. Can save `key, value` attribute pairs to server and these pairs remain visible in the UI. Can also delete `key, value` pairs from UI and server.

## Verification Steps

1. Go to Realm Roles section.
2. Create a new realm role.
3. Click on the "Attributes" tab.
4. Add a new key, value pair. 
5. Add another pair.
6. Try deleting a pair. 
7. Save.
8. Go back to the realm roles list. 
9. Click on the same role you just saved, and navigate back to the "Attributes" tab. 
10. Verify that the save function works correctly and all the attributes you added are shown in the UI. 

## Checklist:


src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,80 +17,127 @@ import {
...
-            {row.map((cell, cellIndex) => (
-              <Td
-                key={`${rowIndex}_${cellIndex}`}
-                id={`text-input-${rowIndex}-${cellIndex}`}
-                dataLabel={columns[cellIndex]}
-              >
-                {cell}
-              </Td>
+    <>
+      <FormAccess role="anyone" isHorizontal onSubmit={form.handleSubmit(save)}>
```
srambach, line 52: If you remove `isHorizontal` then the ActionGroup buttons will be left justified. It's trying to move them over to align with the fields when the labels are on the left instead of above as they are here.

mfrances17 (changes requested):
General testing comments... i was not able to save attributes from a newly created role. 

Also some styling issues, I don't think the tab body should allow selectable, it serves no purpose and makes it look too busy... i tend to disable that on tabs. 

You've got a couple of linting issues just need to run the formatter on those files.

Just noticed that this was a draft PR so will wait for save functionality to be working to continue review.

src/realm-roles/RoleAttributes.tsx
```diff
@@ -31,24 +26,30 @@ type RoleAttributesProps = {
 export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
   const { t } = useTranslation("roles");
   const history = useHistory();
-  const { formState } = form;
+  const [isKeyInputValid, setKeyInputValid] = useState(form.formState.isValid);
+  const [isValueInputValid, setValueInputValid] = useState(
```
edewit, line 30: Isn't this a bit complicated?

src/realm-roles/RoleAttributes.tsx
```diff
@@ -140,7 +144,7 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
           <Button
             variant="primary"
             type="submit"
-            isDisabled={!formState.isValid}
+            isDisabled={!isKeyInputValid || !isValueInputValid}
```
edewit, line 147: Isn't `isValueInputValid` the same as `form.formState.isValid` for this to work though you have to pass `onChange` to the `useForm` hook like this:

```javascript
useForm({ mode: "onChange" })
```

edewit (changes requested):

src/realm-roles/RealmRolesSection.css
```diff
@@ -9,9 +9,22 @@
...
   --pf-c-form__group--m-action--MarginTop: calc(
     var(--pf-global--spacer--2xl) - var(--pf-global--spacer--sm)
   );
+
+  /* this works */
+  /* display: block !important;  */
+
+  /* this works too but the value for the margin isn't right */
+  /* margin-left: calc(
+    var(--pf-global--spacer--2xl)*-1); */
```
srambach, line 29: Are these comments still needed?

jenny-s51: nope! nice catch @srambach - removed

src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,82 +12,130 @@ import {
...
+                    dataLabel={columns[2]}
+                  >
+                    <Button
+                      id="minus-icon"
+                      variant="link"
+                      className="kc-role-attributes__minus-icon"
+                      onClick={() => remove(rowIndex)}
+                    >
+                      <MinusCircleIcon />
+                    </Button>
```
srambach, line 106: Since this button doesn't have text, it needs to either use aria-label or aria-labelledby. In this case, there really isn't another element to label it. Best would be to concatenate in the values you are deleting, so something like "Remove mykey with value myvalue"

Also, the id needs to be unique, so you will need to append rowIndex or something.

src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,82 +12,130 @@ import {
...
+                      },
+                    })}
+                    aria-label="value-input"
+                    defaultValue={attribute.value}
+                  />
+                </Td>
+                {rowIndex !== fields.length - 1 && fields.length - 1 !== 0 && (
+                  <Td
+                    key="minus-button"
+                    id="kc-minus-button"
```
srambach, line 96: This id needs to be unique

src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,82 +12,130 @@ import {
...
-            {row.map((cell, cellIndex) => (
-              <Td
-                key={`${rowIndex}_${cellIndex}`}
-                id={`text-input-${rowIndex}-${cellIndex}`}
-                dataLabel={columns[cellIndex]}
-              >
-                {cell}
-              </Td>
+    <>
+      <Form role="anyone" onSubmit={form.handleSubmit(save)}>
```
edewit, line 43:
```suggestion
      <FormAccess role="anyone" onSubmit={form.handleSubmit(save)}>
```

srambach, line 43: I think this needs to be `FormAccess`

src/realm-roles/RealmRoleForm.tsx
```diff
@@ -92,131 +75,10 @@ export const RoleForm = ({ form, save, editMode }: RoleFormType) => {
         <Button variant="primary" type="submit">
           {t("common:save")}
         </Button>
-        <Button variant="link" onClick={() => history.push(`/${realm}/roles`)}>
+        <Button variant="link" onClick={() => location.reload()}>
```
edewit, line 78: why not do a form reset here instead of reloading the entire application?

src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,82 +12,133 @@ import {
...
+        </TableComposable>
+        <ActionGroup className="kc-role-attributes__action-group">
+          <Button
+            variant="primary"
+            type="submit"
+            isDisabled={!form.formState.isValid}
+          >
+            {t("common:save")}
+          </Button>
+          <Button variant="link" onClick={() => history.push(`${url}/roles`)}>
```
edewit, line 137: this gives a "Page not found" also here I would use reset function of the react-form-hook

src/realm-roles/RoleAttributes.tsx
```diff
@@ -17,82 +12,133 @@ import {
...
+                  key={`${attribute}-value`}
+                  id={`text-input-${rowIndex}-value`}
+                  dataLabel={columns[1]}
+                >
+                  <TextInput
+                    name={`attributes[${rowIndex}].value`}
+                    ref={form.register({
+                      required: true,
+                      minLength: {
+                        value: 1,
```
edewit, line 87: This can be removed right? `minLenght` of 1 is the same as `required: true`

edewit (dismissed):
Don't we need error labels on the key and value fields when they are not filled out?

src/realm-roles/RealmRoleForm.tsx
jenny-s51: great idea - thank you! planning to address these as part of a separate issue

src/realm-roles/RoleAttributes.tsx
jenny-s51: great idea - thank you! planning to address these as part of a separate issue

jenny-s51:
> Don't we need error labels on the key and value fields when they are not filled out?

The "save" button is disabled when the key field is not filled out

mfrances17 (approved):

mfrances17:
LGTM - this can be merged with the stipulation that the outstanding issues will be handled as separate issues

PR #266, opened 2020-12-17
only update state when component is mounted

edewit:
## Motivation

when the component is unmounted because we have navigated to another page we shouldn't update the state of this component anymore

### DIFF
--- a/src/components/scroll-form/ScrollForm.tsx
+++ b/src/components/scroll-form/ScrollForm.tsx
@@ -1,4 +1,4 @@
-import React, { Children, useState } from "react";
+import React, { Children } from "react";
 import { useTranslation } from "react-i18next";
 import {
   Grid,
--- a/src/components/scroll-form/ScrollForm.tsx
+++ b/src/components/scroll-form/ScrollForm.tsx
@@ -19,30 +19,12 @@ type ScrollFormProps = {
 // This must match the page id created in App.tsx unless another page section has been given hasScrollableContent
 const mainPageContentId = "#kc-main-content-page-container";
 
-let spacesToHyphens = (string: string): string => {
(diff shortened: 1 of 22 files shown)


src/components/scroll-form/ScrollForm.tsx
```diff
@@ -19,44 +19,41 @@ type ScrollFormProps = {
 // This must match the page id created in App.tsx unless another page section has been given hasScrollableContent
 const mainPageContentId = "#kc-main-content-page-container";
 
-let spacesToHyphens = (string: string): string => {
```
mfrances17, line 22: unrelated to your PR, but there's an unused useState import

mfrances17 (approved):
LGTM, code looks good. although there are a lot of files touched/code changes they are all consistent. would be nice to have verification steps.

edewit:
@mfrances17 you are right it's a bit of duplication so introduced `useFetch` function that can be used like this in the simplest form:

```javascript
  useEffect(() => {
    return useFetch(
      () => adminClient.components.findOne({ id }),
      (component) => setupForm(component)
    );
  }, []);
```
It has 2 callbacks the first one you call the admin client and fetch some data in the second one you set your state with the result of the first call. `useFetch` will only do the state change when the component is mounted.

ssilvert:
@edewit Very nice fix.  Not only does if fix the flicker on realm selector, the whole UI seems to be noticeably faster when going from one section to the other.

@mfrances17 I haven't looked closely at the code but if you approve of the latest changes, go ahead and merge.

@edewit Since this touches a lot of stuff and appears to change the way we do some things, is this something that needs to be explained to the rest of the team?

edewit:
@ssilvert yes, guess we can talk about it in today's meeting

PR #268, opened 2020-12-17
switches don't work in a `FormAccess` form

edewit:
not all of their props where copied


mfrances17 (approved):
LGTM, tested with switches in kerberos settings. This is a great fix, thank you!

PR #269, opened 2020-12-18
onEnter should not post the form

edewit:
fixing: #257


ssilvert (changes requested):
@edewit This is all good, but if you don't mind, please go ahead and fix another bug I found with this dialog.  When you open the dialog the focus should be on the name field instead of the `X` button.

ssilvert (approved):

PR #270, opened 2020-12-18
refresh and message on group delete

edewit:
fixing: #129

### DIFF
--- a/src/groups/GroupsSection.tsx
+++ b/src/groups/GroupsSection.tsx
@@ -59,14 +59,16 @@ export const GroupsSection = () => {
     setIsCreateModalOpen(!isCreateModalOpen);
   };
 
-  const deleteGroup = (group: GroupRepresentation) => {
+  const deleteGroup = async (group: GroupRepresentation) => {
     try {
-      return adminClient.groups.del({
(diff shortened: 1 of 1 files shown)


src/groups/GroupsSection.tsx
```diff
@@ -59,14 +59,16 @@ export const GroupsSection = () => {
...
   };
 
-  const deleteGroup = (group: GroupRepresentation) => {
+  const deleteGroup = async (group: GroupRepresentation) => {
     try {
-      return adminClient.groups.del({
+      await adminClient.groups.del({
         id: group.id!,
       });
+      addAlert(t("groupsDeleted"), AlertVariant.success);
```
ssilvert, line 67: Should be `groupDeleted` instead of `groupsDeleted`.  Probably need a new message for singular delete?

ssilvert (changes requested):

src/groups/GroupsSection.tsx
edewit: good catch I have the message already just didn't use the right key here, fixed now

ssilvert (approved):

PR #272, opened 2021-01-04
added delete client scope functionality

edewit:


jenny-s51 (changes requested):
Nice work @edewit - the deletion functionality is working as expected! 

Wondering if we can modify the error message to match the same message that's shown under `Networks` in chrome dev tools.  

For example, when I try to delete one of the built-in scopes, this is the error message I get:
`Could not delete client scope: Error: Request failed with status code 400`

But when I look under `Networks`, I see:
`errorMessage: "Cannot remove client scope, it is currently in use"`

I think the latter message clarifies what's going on a bit better, and it might be a good idea to use that one in the alert instead... WDYT?

edewit:
@jenny-s51 for sure did that in a313116

jenny-s51 (approved):

ssilvert (approved):

PR #277, opened 2021-01-07
renamed `useFetch` to `asyncStateFetch`

edewit:
otherwise React will mistake it for a react hook and will complain.

## Motivation

Seems our webpack build in storybook does some more checks that the snowpack build does not. If a function starts with `use` react will check if it's used as a proper react hook. Our `useFetch` is not a react hook just a helper function, so renamed it to `asyncStateFetch`

## Verification Steps

Storybook build should work again

## Checklist:

### DIFF
--- a/src/context/auth/AdminClient.tsx
+++ b/src/context/auth/AdminClient.tsx
@@ -17,7 +17,7 @@ export const useAdminClient = () => {
   return adminClient;
 };
 
-export function useFetch<T>(
+export function asyncStateFetch<T>(
   adminClientCall: () => Promise<T>,
   callback: (param: T) => void
 ) {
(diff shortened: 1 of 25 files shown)


src/context/auth/AdminClient.tsx
```diff
@@ -17,7 +17,7 @@ export const useAdminClient = () => {
   return adminClient;
 };
 
-export function useFetch<T>(
+export function asyncStateFetch<T>(
```
ssilvert, line 20: Let's put some doc here so future developers will know when, how, and why to use it.

ssilvert (changes requested):

srambach:
Fixes the storybook build problem although the two user federation stories still give `Cannot read property 'includes' of undefined`

srambach (approved):
LGTM when @ssilvert is happy with documentation 👍

ssilvert (approved):

PR #278, opened 2021-01-07
move common cell formatters to util functions

edewit:


ssilvert (changes requested):
This is good, but it's not just for external links.  Let's rename it to something like "FormattedLink".  Otherwise, developers will think they can only use it for external links.

ssilvert (approved):

PR #281, opened 2021-01-07
initial version realm settings section

edewit:
## Motivation

First tab of the realm section

## Brief Description

Give all sections something to do, also makes the console more useable ;)

## Verification Steps

Go to there realm section, enable / disable / delete and save work. Partial export / imports don't

## Checklist:

### DIFF
--- a/src/clients/ClientsSection.tsx
+++ b/src/clients/ClientsSection.tsx
@@ -11,7 +11,6 @@ import {
 import { ViewHeader } from "../components/view-header/ViewHeader";
 import { useAdminClient } from "../context/auth/AdminClient";
 import { KeycloakDataTable } from "../components/table-toolbar/KeycloakDataTable";
-import { IFormatter, IFormatterValueType } from "@patternfly/react-table";
 import { emptyFormatter, exportClient, getBaseUrl } from "../util";
 import { useAlerts } from "../components/alert/Alerts";
 import ClientRepresentation from "keycloak-admin/lib/defs/clientRepresentation";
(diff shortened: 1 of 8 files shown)


ssilvert (changes requested):
Tested it out and it looks really good.  Only problem I found was that "Reload" doesn't work.

I'll review the code later.

edewit:
what should reload do?

ssilvert:
> what should reload do?

That's a very good question because I don't think we have properly defined this application-wide.  It probably means the same as the HTML input type "reset".  But we need some more clarification on refresh/reload/reset/cancel.  I'll start a discussion on it.

edewit:
right because on the realm role pages it navigates to the list again.... I could do a reset of the form back to the original state

edewit:
maybe we merge this as is and clarify the reload later?

ssilvert:
> maybe we merge this as is and clarify the reload later?

It's clarified now.  `Reload` is the word we use for the action that means, "Don't save. Reload my form data and start over."

ssilvert:
@edewit Please fix conflict.

mfrances17 (approved):
LGTM... reviewed the code and it looks good. Also tested the functionality by creating a new realm and running all the enabled commands.

One caveat, when deleting a realm it works great on the backend but the existing realm is still selectable in the realm drop-down in the new UI. This can lead to some funky behavior/invalid URLs. Worked fine after a refresh. Not sure if we may want to do an automatic reload and default back to another realm in this scenario... not sure the best solution. But i think this can be merged and maybe we file that as a separate potential issue.

edewit:
@mfrances17 this refresh was handled in another PR, I've rebased this one so that you can see that the realm list is now updated after you delete a realm

mfrances17:
@edewit yep, i just tested the rebased version and the realm list is handled nicely. great work. i already approved but am approving again. :)

ssilvert (approved):

PR #300, opened 2021-01-13
add empty input field, make value not required

### LINKED ISSUE #299: Realm roles(attributes): add empty input field to key/value array
Per @ssilvert , the attributes list should always have an empty key/value pair as the last element. This way a user is able to delete the preceding form values.

jenny-s51:
## Motivation

Closes #299 

## Brief Description

Towards https://issues.redhat.com/browse/APPDUX-749

## Verification Steps

1. Go to Realm Roles section.
2. Create a new role.
3. Click save.
4. Verify that the an empty key/value pair of input fields is rendered.

## Checklist:

### DIFF
--- a/src/realm-roles/RoleAttributes.tsx
+++ b/src/realm-roles/RoleAttributes.tsx
@@ -39,7 +39,14 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
 
   return (
     <>
-      <FormAccess role="manage-realm" onSubmit={form.handleSubmit(save)}>
+      <FormAccess
+        role="manage-realm"
+        onSubmit={form.handleSubmit(async (role) => {
+          await save(role);
+          onAdd();
+        })}
+      >
+        {" "}
         <TableComposable
           className="kc-role-attributes__table"
           aria-label="Role attribute keys and values"
@@ -79,7 +86,7 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
                   <TextInput
                     name={`attributes[${rowIndex}].value`}
                     ref={form.register({
-                      required: true,
+                      required: false,
                     })}
                     aria-label="value-input"
                     defaultValue={attribute.value}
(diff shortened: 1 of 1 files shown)


src/realm-roles/RoleAttributes.tsx
```diff
@@ -39,7 +39,15 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
 
   return (
     <>
-      <FormAccess role="manage-realm" onSubmit={form.handleSubmit(save)}>
+      <FormAccess
+        role="manage-realm"
+        onSubmit={form.handleSubmit(async (role) => {
+          await save(role);
+          {
```
edewit, line 46: why the extra curly brackets and why only call `onAdd()` when the form is dirty? If I press the `-` and press save I don't need an extra new row? Also in this case I would do an if instead of the ternary operator

src/realm-roles/RoleAttributes.tsx
```diff
@@ -79,7 +87,7 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
                   <TextInput
                     name={`attributes[${rowIndex}].value`}
                     ref={form.register({
```
edewit, line 89: you can remove this completely:

```suggestion
ref={form.register()}
```

edewit (changes requested):

src/realm-roles/RoleAttributes.tsx
jenny-s51: thank you for your feedback - will update accordingly

src/realm-roles/RoleAttributes.tsx
```diff
@@ -37,9 +37,22 @@ export const RoleAttributes = ({ form, save }: RoleAttributesProps) => {
     append({ key: "", value: "" });
   };
 
+  React.useEffect(() => {
+    append({ key: "useEffect", value: "works" });
+  }, []);
+
+  console.dir(form.formState.dirtyFields.attributes);
```
edewit, line 44: Is this meant for debugging?

edewit (changes requested):
looks simple enough

PR #301, opened 2021-01-13
Add text for user federation user attribute mapper

srambach:
## Motivation
Prepares the text for the forms within this modal https://marvelapp.com/prototype/ecjc94b/screen/72071054
Note that selecting mapper type changes what fields are in the form.
Also deduplicates text strings used in multiple places: name, role, category, type, priority, description

## Brief Description
Adds text to user fed messages and help files as well as common messages file. Moves some definitions that were in multiple places to common-messages.

## Verification Steps
New text for user fed mapper dialog can be checked by comparing to the text when adding/editing a new mapping in the old interface.
Consolidation of duplicate definitions for name, role, category, type, priority, and description should be checked anywhere they are used. Check through screens that none were missed. (which would show as lower case if the definition was removed without pointing to common now)

## Checklist:

- [x ] User-visible strings are using the react-i18next framework (useTranslation)
- [ x] Help has been implemented
- [ na] axe report has been run and resulting a11y issues have been resolved
- [na ] Unit tests have been created/updated
- [ x] Formatting has been performed via prettier/eslint
- [ x] Type checking has been performed via 'yarn check-types'


src/user-federation/help.json
```diff
@@ -64,6 +64,68 @@
     "editModeKerberosHelp": "READ_ONLY means that password updates are not allowed and user always authenticates with Kerberos password. UNSYNCED means that the user can change the password in the Keycloak database and this one will be used instead of the Kerberos password",
     "updateFirstLoginHelp": "Update profile on first login",
 
-    "kerberosCacheSettingsDescription": "This section contains a few basic options common to all user storage providers"
+    "kerberosCacheSettingsDescription": "This section contains a few basic options common to all user storage providers",
```
mfrances17, line 67: There are several issues throughout these help strings, including dropped articles (for example, 'the'), inconsistent punctuation (some have periods, some don't), abbreviations such as e.g. and i.e. which should be avoided, and pronouns like 'he'. We can overlook for now but there will need to be an editing sweep through these strings, perhaps maybe getting Doc involved.

src/user-federation/messages.json
```diff
@@ -83,6 +83,62 @@
     "userFedDeletedSuccess": "The user federation provider has been deleted.",
     "userFedDeleteError": "Could not delete user federation provider: '{{error}}'",
     "userFedDeleteConfirmTitle": "Delete user federation provider?",
-    "userFedDeleteConfirm": "If you delete this user federation provider, all associated data will be removed."
+    "userFedDeleteConfirm": "If you delete this user federation provider, all associated data will be removed.",
+
+    "id": "ID",
+
+    "mapperType": "Mapper type",
+    "mapperTypeMsadUserAccountControlManager": "msad-user-account-control-mapper",
```
mfrances17, line 91: is this section correct? these values look like keys

src/user-federation/messages.json
```diff
@@ -83,6 +83,62 @@
...
+    "mapperTypeMsadLdsUserAccountControlMapper": "msad-user-account-control-mapper",
+    "mapperTypeGroupLdapMapper": "group-ldap-mapper",
+    "mapperTypeUserAttributeLdapMapper": "user-attribute-ldap-mapper",
+    "mapperTypeRoleLdapMapper": "role-ldap-mapper",
+    "mapperTypeHardcodedAttributeMapper": "hardcoded-attribute-mapper",
+    "mapperTypeHardcodedLdapRoleMapper": "hardcoded-ldap-role-mapper",
+    "mapperTypeCertificateLdapMapper": "certificate-ldap-mapper",
+    "mapperTypeFullNameLdapMapper": "full-name-ldap-mapper",
+    "mapperTypeHardcodedLdapGroupMapper": "hardcoded-ldap-group-mapper",
+    "mapperTypeLdaoAttributeMapper": "hardcoded-ldap-attribute-mapper",
```
mfrances17, line 101: looks like a typo, should be mapperTypeLdapAttributeMapper

mfrances17 (approved):
One minor typo, otherwise just some comments on the strings overall. Verify that one mapper section, if those are really the intended values (they look like keys) then I don't think they need to be localized.

edewit (approved):
Good idea to move all these duplicate keys into the common bundle

src/user-federation/messages.json
srambach: I wasn't sure - they are dropdown options, but based on those those, the entire form changes so I thought maybe they should be there. I can remove them, just let me know for sure.

src/user-federation/help.json
srambach: I copied them from the old system and in some places I did catch and fix the grammar but then wondered if I should be doing that. Is a content manager going to go over all the text at some point?

mfrances17: @ssilvert probably a question better suited for you

src/user-federation/messages.json
mfrances17: let's keep them for now, i'll need to learn more about this part of the product, especially if these would ever need to be translated. my gut says no, but it won't hurt anything to keep them here.

srambach:
@mfrances17 I cleaned up some grammar and punctuation and fixed the typo. Someday, a content writer who understands the system should scrub through the entire app. @ssilvert

jenny-s51 (approved):

PR #304, opened 2021-01-14
Cypress tests for masthead.

ssilvert:
## Motivation
Add masthead tests, fixing https://github.com/keycloak/keycloak-admin-ui/issues/37

## Verification Steps
Run the tests.

## Checklist:


ssilvert:
Cypress tests were running locally but failing here because it brings up the old account console.  Apparently, we are running these tests against an old version of Keycloak.  That was really easy to figure out based on the mp4 output from Cypress.

I'm really, really loving Cypress!

ssilvert:
@Aboullos @edewit I've tried to upgrade the version of Keycloak we are using by bumping the version in start.js.  However, that produced a URL that doesn't exist.  Changing it to the correct download URL didn't seem to help.  Since I don't really know how all that stuff works I'm probably just making a mess of things.  Can you guys help?

Aboullos:
I am getting an error when launching the ./start.js.
The first Cypress test uses the old console to create the admin user and add the new admin console as a client.

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+  it("logs out", () => {
+    cy.get(userDropdown).click();
+    cy.get("#sign-out").click();
+    loginPage.isLogInPage();
+  });
+};
+
+const goToAcctMgtTest = (userDropdown) => {
+  it("opens manage account and returns to admin console", () => {
+    cy.get(userDropdown).click();
```
Aboullos, line 17:
```suggestion
    cy.get(userDropdown).click();
    cy.get("#manage-account").click();
```

It is better to use PageObjects to reuse code, this two lines can be replaced by
```
    headerPage.goToManageAccount();
```

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    cy.get("#manage-account").click();
```
Aboullos, line 18: It is better to use PageObjects to reuse code, this two lines can be replaced by
```suggestion
    headerPage.goToManageAccount();
```

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    cy.contains("Welcome to Keycloak Account Management");
+    cy.get("#landingReferrerLink").click({ force: true });
```
Aboullos, line 20: This can be set in a PageObject like `AccountConsolePage` as two methods `checkIsAccountConsole` and `goBackToAdminConsole`

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    cy.get(userDropdown);
```
Aboullos, line 21: This one can be removed

start.js
```diff
@@ -1,53 +1,55 @@
...
 if (!fs.existsSync(fileName)) {
   const file = fs.createWriteStream(fileName);
-  http.get(`https://downloads.jboss.org/keycloak/${version}/keycloak-${version}.tar.gz`, (response) => {
-    response.pipe(file)
-    response.on('end', () => {
-      console.log('Downloaded keycloak');
-      decompressKeycloak().then(() => run());
-    });
-  });
+  http.get(
```
edewit, line 41: The reason it fails is because of a 302 from github so we should follow that redirect:

```typescript
const request = (url, file) => {
  http.get(url, (response) => {
    if (response.statusCode == 302) {
      request(response.headers.location, file);
    } else {
      response.pipe(file);
      response.on("end", () => {
        console.log("Downloaded keycloak");
        decompressKeycloak().then(() => run());
      });
    }
  });
}

if (!fs.existsSync(fileName)) {
  const file = fs.createWriteStream(fileName);
  request(`https://github.com/keycloak/keycloak/releases/download/${version}/keycloak-${version}.tar.gz`, file);
```

start.js
```diff
@@ -1,53 +1,55 @@
...
+const decompressTargz = require("decompress-targz");
 
 const args = process.argv.slice(2);
-const version = args[0] || '11.0.0';
+const version = args[0] || "12.0.1";
 
-const folder = 'server';
-const fileName = path.join(folder, 'keycloak.tar.gz');
+const folder = "server";
+const fileName = path.join(folder, "keycloak.tar.gz");
```
edewit, line 14: Let's change this line so that we always download the right version and don't skip the download step because the old file is there
```suggestion
const fileName = path.join(folder, `keycloak-${version}.tar.gz`);
```

edewit (changes requested):

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+describe("Masthead tests with kebab menu", () => {
+  beforeEach(() => {
+    cy.viewport("iphone-6");
+    cy.visit("");
+    loginPage.logIn();
+  });
+
+  it("shows kabab and hides regular menu", () => {
+    cy.get("#user-dropdown").should("not.exist");
+    cy.get("#user-dropdown-kebab").should("exist");
```
Aboullos, line 62: This two can be added to `HeaderPage` something as `checkUserDropdown`
And it could check `#user-dropdown` is present and #user-dropdown-kebab is not when `Cypress.config().viewportWidth < 768` and viceversa

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
+import ListingPage from "../support/pages/admin_console/ListingPage.js";
+import LoginPage from "../support/pages/LoginPage.js";
+import SidebarPage from "../support/pages/admin_console/SidebarPage.js";
+
+const loginPage = new LoginPage();
+
+const logOutTest = (userDropdown) => {
+  it("logs out", () => {
+    cy.get(userDropdown).click();
+    cy.get("#sign-out").click();
```
Aboullos, line 10:
```suggestion
    headerpage.signOut();
```

And modify the signOut() method to user a different dropdown according to the viewport wide 
e.g. `Cypress.config().viewportWidth < 768`

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    listingPage.goToItemDetails("address");
+
+    cy.get("#view-header-subkey").should("exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("exist");
+
+    cy.get("#help").click();
+    cy.get("#enableHelp").click({ force: true });
+
+    cy.get("#view-header-subkey").should("not.exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("not.exist");
```
Aboullos, line 47: This can be added as two methods in `HeaderPage`, something like `helpIsShown(boolean)` and `switchShowHelp()`

start.js
ssilvert: @edewit Can you make this into a commit suggestion?  Then I'll commit it and we can give it a try.

ssilvert:
@Aboullos Please re-review.

I have refactored some of this stuff and taken some of your suggestions.  We now have a Masthead class.  I did remove code that was not being used and I moved the dialog code into a utility class because it doesn't really belong as part of the Masthead.

Also note that I have added `id` attributes to the code to make it easier to query parts of the DOM.  We should make this a best practice because `id` attributes a much less brittle than relying on more complex queries.

For various reasons, I did not think it appropriate to make a class for the account console.   We must never navigate to account console in any admin test outside of the tests I wrote here.

tests/cypress/support/pages/admin_console/Masthead.js
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+      cy.viewport();
+    }
+
+    this.isMobileMode = isMobileMode;
+  }
+
+  toggleGlobalHelp() {
+    cy.get("#help").click();
+    cy.get("#enableHelp").click({ force: true });
```
Aboullos, line 36: This would be better to add it as an object in the constructor with the others, to separate elements from actions.
Something like `enableHelpSwitch`

tests/cypress/support/pages/admin_console/Masthead.js
```diff
@@ -0,0 +1,62 @@
...
+    if (this.isMobileMode) {
+      return cy.get(this.userDrpDwnKebab);
+    } else {
+      return cy.get(this.userDrpDwn);
+    }
+  }
+
+  signOut() {
+    this.userDropdown().click();
+    cy.get("#sign-out").click();
```
Aboullos, line 49: This would be better to add it as an object in the constructor with the others, to separate elements from actions.
Something like `signOutBtn` as it was in the previous `HeaderPage`

tests/cypress/support/pages/admin_console/Masthead.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  accountManagement() {
+    this.userDropdown().click();
+    cy.get("#manage-account").click();
```
Aboullos, line 54: This would be better to add it as an object in the constructor with the others, to separate elements from actions.
Something like `manageAccountBtn` as it was in the previous `HeaderPage`

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    cy.get("#view-header-subkey").should("exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("exist");
+
+    masthead.toggleGlobalHelp();
+
+    cy.get("#view-header-subkey").should("not.exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("not.exist");
+  });
+
+  logOutTest("#user-dropdown");
```
Aboullos:
```suggestion
  logOutTest();
```

tests/cypress/integration/masthead_test.spec.js
Aboullos: > This two can be added to `Masthead` something as `checkUserDropdown`
> And it could check `#user-dropdown` is present and `#user-dropdown-kebab` is not using `isMobileMode()`

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,69 @@
...
+const masthead = new Masthead();
+
+const logOutTest = () => {
+  it("logs out", () => {
+    masthead.signOut();
+    loginPage.isLogInPage();
+  });
+};
+
+const goToAcctMgtTest = (userDropdown) => {
```
Aboullos:
```suggestion
const goToAcctMgtTest = () => {
```

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+};
+
+describe("Masthead tests in desktop mode", () => {
+  beforeEach(() => {
+    cy.visit("");
+    loginPage.logIn();
+  });
+
+  goToAcctMgtTest("#user-dropdown");
```
Aboullos:
```suggestion
  goToAcctMgtTest();
```

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+  it("shows kabab and hides regular menu", () => {
+    cy.get("#user-dropdown").should("not.exist");
+    cy.get("#user-dropdown-kebab").should("exist");
+  });
+
+  // TODO: Add test for help when using kebab menu.
+  //       Feature not yet implemented for kebab.
+
+  goToAcctMgtTest("#user-dropdown-kebab");
```
Aboullos:
```suggestion
  goToAcctMgtTest();
```

tests/cypress/integration/masthead_test.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    sidebarPage.goToClientScopes();
+    listingPage.goToItemDetails("address");
+
+    cy.get("#view-header-subkey").should("exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("exist");
+
+    masthead.toggleGlobalHelp();
+
+    cy.get("#view-header-subkey").should("not.exist");
+    cy.get(`#${CSS.escape("client-scopes-help:name")}`).should("not.exist");
```
Aboullos, line 46: To have this as a reusable method (it could be used in tests to check that all the windows that contain help can show and hide it. It could be in `Masthead` as `helpIsShown(boolean)` where the "exist" or ''not.exist" change according to the boolean value
```suggestion
   masthead.helpIsShown(true);

    masthead.toggleGlobalHelp();

    masthead.helpIsShown(false);
```

Aboullos (changes requested):

tests/cypress/integration/masthead_test.spec.js
ssilvert: already changed that one

tests/cypress/integration/masthead_test.spec.js
ssilvert: already changed that one

tests/cypress/integration/masthead_test.spec.js
ssilvert: We should never hide the assertions in a utility class like Masthead.  You should always be able to read the test and know what it is asserting.  If I put this stuff into Masthead it would hide the entire test.

Also, we don't want to refactor something just because we think we might need it later.  It's extra complexity for no benefit.  Follow the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

tests/cypress/integration/masthead_test.spec.js
ssilvert: Already done

tests/cypress/support/pages/admin_console/Masthead.js
ssilvert: No reason to do this until someone needs it.  See YAGNI comments below.

tests/cypress/support/pages/admin_console/Masthead.js
```diff
@@ -0,0 +1,62 @@
+export default class Masthead {
+  constructor() {
+    this.isMobileMode = false;
```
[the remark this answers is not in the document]
ssilvert: fixed

tests/cypress/integration/masthead_test.spec.js
Aboullos: This could be changed to use the elements from `Masthead` to remove the hardcoded strings and in case the element changes in a future
```suggestion
    cy.get(masthead.userDrpDwn).should("not.exist");
    cy.get(masthead.userDrpDwnKebab).should("exist");
```

Aboullos (approved):

PR #307, opened 2021-01-15
Add Save/Cancel buttons and logic to Kerberos settings

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-721

## Brief Description
Adds the ability to save (or cancel) changes made to the User Federation Kerberos settings page.

## Verification Steps
1. Go to User federation and select a Kerberos card.
2. Change any of the values in the Required Settings or Cache Settings sections.
3. Click Save. You should see a success alert.
4. Verify that the changes have been saved by traversing anywhere else in the UI and then re-clicking the card to make sure the values that were changed are displayed correctly.

## Checklist:

## Additional Notes
Screen cap of successful save:
kerberos-save-success

### DIFF
--- a/src/user-federation/kerberos/KerberosSettingsCache.tsx
+++ b/src/user-federation/kerberos/KerberosSettingsCache.tsx
@@ -25,7 +25,6 @@ import { WizardSectionHeader } from "../../components/wizard-section-header/Wiza
 export type KerberosSettingsCacheProps = {
   showSectionHeading?: boolean;
   showSectionDescription?: boolean;
-  // form: UseFormMethods;
 };
 
 export const KerberosSettingsCache = ({
--- a/src/user-federation/kerberos/KerberosSettingsCache.tsx
+++ b/src/user-federation/kerberos/KerberosSettingsCache.tsx
@@ -8,27 +8,19 @@ import {
 import { useTranslation } from "react-i18next";
 import { HelpItem } from "../../components/help-enabler/HelpItem";
 import React, { useState } from "react";
-import {
-  SubmitHandler,
-  UseFormMethods,
-  Controller,
-  useWatch,
-} from "react-hook-form";
-import ComponentRepresentation from "keycloak-admin/lib/defs/componentRepresentation";
+import { UseFormMethods, Controller, useWatch } from "react-hook-form";
 import { FormAccess } from "../../components/form-access/FormAccess";
 import _ from "lodash";
 import { WizardSectionHeader } from "../../components/wizard-section-header/WizardSectionHeader";
 
 export type KerberosSettingsCacheProps = {
   form: UseFormMethods;
-  save: SubmitHandler<ComponentRepresentation>;
   showSectionHeading?: boolean;
   showSectionDescription?: boolean;
 };
 
 export const KerberosSettingsCache = ({
   form,
-  save,
   showSectionHeading = false,
   showSectionDescription = false,
 }: KerberosSettingsCacheProps) => {
@@ -79,11 +71,7 @@ export const KerberosSettingsCache = ({
       )}
 
       {/* Cache settings */}
-      <FormAccess
-        role="manage-realm"
-        isHorizontal
-        onSubmit={form.handleSubmit(save)}
-      >
+      <FormAccess role="manage-realm" isHorizontal>
         <FormGroup
           label={t("cachePolicy")}
           labelIcon={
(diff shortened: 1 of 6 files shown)


jenny-s51 (approved):
LGTM @mfrances17 - tested in UI, working as expected!

srambach:
- If the user selects "Allow password authentication" but doesn't select an edit mode, they will get an error 500 with no other indication of the problem.
- Changing the cache policy to "select an option" yields an error 500 even though that field is not marked as required.
- I think the save button is not supposed to be active unless the form is "dirty" but is currently active when editing an existing entry without any changes.
- If I clear out the required text inputs, it still saves ok, and in fact I was able to save without a name, which causes an empty card back on the main user federation page.

mfrances17:
> * If the user selects "Allow password authentication" but doesn't select an edit mode, they will get an error 500 with no other indication of the problem.
> * Changing the cache policy to "select an option" yields an error 500 even though that field is not marked as required.
> * Also if cache policy is set to evict but the resulting drop downs don't have values (and they are not marked as required), then you get an error 500.

All fields that are required, including those that are only visible if you select a certain option, are now marked as required. Also, I've removed all of the 'Select an option' dropdown selections as they are causing errors and the old admin UI did not do this. 

> * If I clear out the required text inputs, it still saves ok, and in fact I was able to save without a name, which causes an empty card back on the main user federation page.

Text fields that are empty will now show a field-level error on submit, and the form won't be saved.

> * I think the save button is not supposed to be active unless the form is "dirty" but is currently active when editing an existing entry without any changes.

No other forms are doing that in the new UI.

edewit (changes requested):
Still get a 500 when switching from `DEFAULT` to `DAILY` and then hit save

src/user-federation/kerberos/KerberosSettingsCache.tsx
```diff
@@ -154,12 +120,13 @@ export const KerberosSettingsCache = ({
                 forID="kc-eviction-day"
               />
             }
+            isRequired
             fieldId="kc-eviction-day"
           >
             <Controller
               name="config.evictionDay"
               defaultValue=""
```
edewit, line 128: either add a default value here or add a "select option" with empty and make it required with: `rules={{ required: true }}`

mfrances17: good catch, i needed to add default values to those three fields since removing the choose string... working now

src/user-federation/kerberos/KerberosSettingsCache.tsx
```diff
@@ -73,18 +46,16 @@ export const KerberosSettingsCache = ({
...
   }
 
   const minuteOptions = [
-    <SelectOption key={0} value={t("common:selectOne")} isPlaceholder />,
+    <SelectOption key={0} value={[`${1}`]} isPlaceholder />,
   ];
-  for (let index = 1; index <= 60; index++) {
-    minuteOptions.push(<SelectOption key={index + 1} value={[`${index}`]} />);
+  for (let index = 2; index <= 60; index++) {
+    minuteOptions.push(<SelectOption key={index - 1} value={[`${index}`]} />);
```
ssilvert, line 58: In the dropdown, choices for eviction minute go from 1 to 60.  Look at old console.  It should be 00 to 59.

src/user-federation/kerberos/KerberosSettingsCache.tsx
```diff
@@ -73,18 +46,16 @@ export const KerberosSettingsCache = ({
...
   );
 
-  const hourOptions = [
-    <SelectOption key={0} value={t("common:selectOne")} isPlaceholder />,
-  ];
-  for (let index = 1; index <= 24; index++) {
-    hourOptions.push(<SelectOption key={index + 1} value={[`${index}`]} />);
+  const hourOptions = [<SelectOption key={0} value={[`${1}`]} isPlaceholder />];
+  for (let index = 2; index <= 24; index++) {
+    hourOptions.push(<SelectOption key={index - 1} value={[`${index}`]} />);
```
ssilvert, line 51: In the dropdown, choices for eviction hour go from 1 to 24.  Look at old console.  It should go from 00 to 23.

src/user-federation/kerberos/KerberosSettingsCache.tsx
```diff
@@ -297,14 +261,15 @@ export const KerberosSettingsCache = ({
                 forID="kc-max-lifespan"
               />
             }
+            isRequired
```
ssilvert, line 264: This is not a required field in the old console.  I suspect that if it is blank that means "forever".  You aren't failing on that anyway, so I think we can just remove `isRequired`.

ssilvert (changes requested):

PR #312, opened 2021-01-20
added roles tab to clients

edewit:
## Motivation
Roles tab under clients is the same as the roles section this PR add it see [marvel](https://marvelapp.com/prototype/bi16ff1/screen/75842763)

## Verification Steps

See that there is a new tab under client details named roles that has the same functionality as the roles section

## Checklist:

### DIFF
--- a/src/clients/ClientDetails.tsx
+++ b/src/clients/ClientDetails.tsx
@@ -31,7 +31,7 @@ import {
 } from "../components/multi-line-input/MultiLineInput";
 import { ClientScopes } from "./scopes/ClientScopes";
 import { EvaluateScopes } from "./scopes/EvaluateScopes";
-import { RealmRolesList } from "../realm-roles/RealmRolesList";
+import { RolesList } from "../realm-roles/RolesList";
 import { ServiceAccount } from "./service-account/ServiceAccount";
 import { KeycloakTabs } from "../components/keycloak-tabs/KeycloakTabs";
 
@@ -108,6 +108,10 @@ export const ClientDetails = () => {
   const { id } = useParams<{ id: string }>();
 
   const [client, setClient] = useState<ClientRepresentation>();
+
+  const loader = async () => {
+    return await adminClient.clients.listRoles({ id });
+  };
 
   const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
     titleKey: "clients:clientDeleteConfirmTitle",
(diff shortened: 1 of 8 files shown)


jenny-s51 (changes requested):
The new roles tab displays the list of roles as expected, and it looks like "create role" is also working through client details! There are a few things I noticed while testing:

- Go to clients -> click on one of the roles -> details page renders as expected -> click on "attributes" tab -> `TypeError: Cannot read property 'name' of null
    at RealmRoleTabs.js:85`
- Go to clients -> click on one of the roles -> details page renders as expected -> click on "details" tab again -> form is cleared of all data and `TypeError: Cannot read property 'name' of null at RealmRoleTabs.js:85`
- if either of the above two are done, clicking on "Client Details" in the breadcrumb results in a loading spinner without taking you back 
- if we try to create a new role, the role itself is created but the same form-clearing errors show up when we try to click on "details" or "attributes"
- role description doesn't get updated when update is attempted

edewit:
@jenny-s51 right there was a duplicated `id` parameter fixed that now.

jenny-s51 (approved):
Looks great @edewit !

ssilvert (changes requested):
@edewit The roles tab displays the wrong set of roles.  It's pulling up the realm roles instead of the client's roles.

edewit:
@ssilvert didn't know client had roles the design just said it should be the same page

ssilvert:
> @ssilvert didn't know client had roles the design just said it should be the same page

Yes they do.  I think the designer just meant it was the same design.  Different data though.  Check out old console.

mfrances17 (changes requested):
Found a couple of issues when testing:

When I initially create a role, I can't add a description - I can type one but it isn't saved. I can only enter a description later after the initial save.

When I delete a role, I get a success notification but my view is still of an empty role details screen, and in the console there is the following error:
`Uncaught (in promise) TypeError: Cannot read property 'name' of null at RealmRoleTabs.js:83`

edewit:
@mfrances17 good catch the delete problem only happened when you delete from the attributes tab, didn't notice that. Also fixed the create with description

mfrances17 (approved):
Both of the issues I had found while testing are now fixed, the Roles tab is working really well now. Just need to resolve some newly introduced conflicts but LGTM.

PR #330, opened 2021-01-26
multiple ux issues around realms

edewit:
fixes: #324


src/realm/messages.json
```diff
@@ -6,6 +6,8 @@
     "createRealm": "Create realm",
     "realmExplain": "A realm manages a set of users, credentials, roles, and groups. A user belongs to and logs into a realm. Realms are isolated from one another and can only manage and authenticate the users that they control.",
     "noRealmRoles": "No realm roles",
-    "emptyStateText": "There aren't any realm roles in this realm. Create a realm role to get started."
+    "emptyStateText": "There aren't any realm roles in this realm. Create a realm role to get started.",
+    "saveRealmSuccess": "Realm created successfully",
+    "saveRealmError": "Could not create realm {{error}}"
```
jenny-s51, line 11:
```suggestion
    "providerInfo": "Provider Info"
```

public/index.css
```diff
@@ -4,3 +4,7 @@
 .keycloak__pageheader_brand {
   height: 35px;
 }
+
+.keycloak__form {
+  max-width: 768px;
```
jenny-s51, line 9: @edewit @yih-wang 768px still seems a bit narrow... the "resource file" placeholder text is getting truncated:

Screen Shot 2021-01-26 at 3 47 56 PM

jenny-s51 (changes requested):
Save and cancel buttons are working as expected!

I'm seeing an empty sidebar when I navigate to the "create realm" page though.

Also, `providerInfo` hasn't been declared in `common-messages.json` even though it's localized here in `Dashboard`: https://github.com/keycloak/keycloak-admin-ui/blob/master/src/dashboard/Dashboard.tsx#L53

public/index.css
yih-wang: Agree. And it looks different from what I expected... Maybe the actual size in the screen should be twice 768px? Basically, the width of this form equals the form with a scroll-spy. So when users jump between different screens, the forms can keep in the same position.

edewit: seems those forms are 1024px so changed it to that size

edewit:
@jenny-s51 can't seem to get the empty navigation on the create role screen how did you do that?

jenny-s51 (commented):
@edewit this is what the left sidebar looks like for me when I navigate to "create realm". Are you seeing something else?

EDIT: Ah ok, I understand now this is the expected behavior.

Screen Shot 2021-01-27 at 11 13 16 AM

src/dashboard/Dashboard.tsx
```diff
@@ -50,7 +50,7 @@ const EmptyDashboard = () => {
         </Title>
         <EmptyStateBody>{t("introduction")}</EmptyStateBody>
         <Button variant="link" onClick={() => setRealm("master")}>
```
jenny-s51, line 52:
```suggestion
        <Button variant="link" onClick={() => setRealm("master")}>
```

Clicking on "realm info" for any of the realms always changes to realm to "master" in the realm dropdown, and takes me to the info page for the master realm.

edewit: right changed the button to say 'Server info'

edewit:
> @edewit this is what the left sidebar looks like for me when I navigate to "create realm". Are you seeing something else?
> 
> EDIT: Ah ok, I understand now this is the expected behavior.
> 
> Screen Shot 2021-01-27 at 11 13 16 AM

This is a fix, it was designed to behave that way

mfrances17 (changes requested):
I tested vs the issues that were filed and the changes look great. I noticed a couple of other issues that you may want to fix here or we can add them as new Issues:

Clear button string needs to be capitalized:
lower-case-clear

There are a bunch of errors thrown if the user starts typing directly into the resource file field, about one per keystroke. We probably want to handle this:
json-errors

edewit:
@mfrances17 right, that is not nice fixed those

src/components/json-file-upload/JsonFileUpload.tsx
```diff
@@ -85,7 +85,7 @@ export const JsonFileUpload = ({
             </Button>,
           ]}
         >
-          {t("Are you sure you want to clear this file?")}
+          {t("clearExplain")}
```
mfrances17, line 88: should be clearFileExplain

mfrances17 (changes requested):
Both of the issues i raised are fixed... just one quick change needed to fix the message reference:
clear

src/realm/add/NewRealmForm.tsx
```diff
@@ -27,14 +28,26 @@ export const NewRealmForm = () => {
...
+    control,
+    formState,
+    errors,
+  } = useForm<RealmRepresentation>({ mode: "onChange" });
 
   const handleFileChange = (value: string | File) => {
     const defaultRealm = { id: "", realm: "", enabled: true };
 
-    const obj = value ? JSON.parse(value as string) : defaultRealm;
+    let obj: { [name: string]: any } = defaultRealm;
```
ssilvert, line 43: Avoid `any`

ssilvert (changes requested):
Other than the `any` thing and the one failing test, this looks very good to me.

ssilvert (approved):

Issue #291, opened 2021-01-12
realm roles(attributes): set key fields to read-only after save


jenny-s51:
follows #264 

set key input fields to read-only on `save()` form submission

edewit:
In the old console this was done because it was the easiest way to implement it, I guess. But the keycloak API has no problem with changing keys of existing attributes, so why would we make them readonly after a `save`? Wouldn't it be better / more user friendly to allow the user to change keys?

jenny-s51:
@edewit I definitely agree. 

@yih-wang what are your thoughts?

yih-wang:
@jenny-s51 @edewit Yes that sounds great. Make the keys editable is more friendly.

jenny-s51:
Closing this based on feedback from @edewit and @yih-wang - consensus that editable keys are more user-friendly

PR #331, opened 2021-01-26
Realm roles(role attributes): adds initial empty field on load, also adds empty field on save

jenny-s51:
## Motivation

#299, #291
https://issues.redhat.com/browse/APPDUX-772

## Brief Description

This PR allows a user to successfully save, update, and delete role attributes.

The work done in this issue addresses the observation by @ssilvert that we always need to have an empty field at the bottom of the field array so a user is able to easily delete the last key/value pair in the list without having to add an additional row. 

@edewit and I also made a large effort to address #291 in this PR. We spent a great deal of time attempting many work-arounds however due to some bugs in react-hook-form this is not possible with their current API. Therefore at the moment all key/value pairs can be edited after `save`, and the last (empty) input field has functionality for both adding and deleting the row. 

## Verification Steps

1. Go to Realm Roles section.
2. Click on an existing role.
3. Go to the "Attributes" tab.
4. Try adding an attribute and saving.
5. Try adding multiple attributes and saving.
6. Try deleting one or multiple attributes and saving.
8. Navigate to a different page in the UI, and then navigate back to "Attributes" to confirm the data was saved successfully.

## Checklist:


edewit (approved):
:+1:

mfrances17 (changes requested):
The behavior of this tab seems odd... is this really the expected behavior? So I'm adding and subtracting using -/+ and the only time the Save button is active is when any of the individual fields are dirty. But I could've added 100 key-value pairs using +/-, they are all unable to be saved using Save, I navigate away, come back and they are all gone.

jenny-s51:
@mfrances17 this is the expected behavior! the user is not able to save if any of the key fields are empty. similarly, you should not be able to use the '+' if any of the key fields are empty.

if you want to '-' a row and save, you have to delete the last row on the bottom because we can't save an empty key/value pair

mfrances17:
To clarify/reiterate... the Save button is only clickable if you are in a dirty field. Once you have clicked + to add a key/value pair, the Save button is no longer clickable. As a user, I would infer that it has been saved, so I navigate away and when I come back, all of the key/value pairs I added are now gone.

In other words, the Save button needs to be smart enough to know not just when a field is dirty, but also that the key/value pair list itself is dirty.

jenny-s51:
@mfrances17 thank you for clarifying -- I agree that a user can easily think the form is saved as soon as they '+' a new row, especially if the 'save' button immediately becomes disabled. 

According to the mocks, it looks like the only time the 'save' button should be disabled is when there is only one key/value pair on the form, and both fields are empty. https://marvelapp.com/prototype/f39d464/screen/72683281

I updated the behavior to match this with `watch` from `react-hook-form`.

mfrances17 (approved):
Yes, good change with the Save button - this prevents it from being unselectable when new pairs have been added using the + button. LGTM.

PR #345, opened 2021-01-29
Limit search users when users federated or large size

edewit:
## Motivation

In order to garantee performance we don't allow to list all users when users are federated or when the size of the user list > 100

## Brief Description

When either users are federated or user size > 100 an initial search field is show [marvel](https://marvelapp.com/prototype/7i7ja65/screen/75693209)

## Verification Steps

Add a fake user federation then go to the user section to see the new dialog

## Checklist:


ssilvert:
@edewit Search does not handle 500 error on server.  For instance, I had a malformed url in my LDAP config.   Caused exception on server side and returned 500 error.  But screen froze.

ssilvert (changes requested):
If search returns less than 10, then table will report one too many elements.  For instance, if the search returns 2 users, the table will say `1-3`

Need to correct some bad text that came in through the design.

In heading, it says `This is the description`.  I think we can just remove that for now.

`Start by searching a user` should say `Start by searching for users`

`You can search for users in the local or a federation provider if there are any. To start a search, specify a provider first.` doesn't make much sense.  Ideally, this would have a description about what kind of query is accepted. Let's just get rid of this text for now.

ssilvert:
> `Start by searching a user` should say `Start by searching for users`
> 
Above is still not fixed.

ssilvert:
> If search returns less than 10, then table will report one too many elements. For instance, if the search returns 2 users, the table will say `1-3`

@edewit If I change the table to show "20 per page", the table will say `2.000000000000001 - 21`

ssilvert:
Since this is over due I'm going to merge it and I'll open issues for the remaining work.

PR #349, opened 2021-02-01
Add enable switch and action menu to LDAP settings

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-719

## Brief Description
Adds layout for Enable toggle switch, Action menu and commands, and Remove imported modal. Adds business logic for the Enable switch and the Delete provider menu command. 

Note that the logic for the 4 user-related menu commands is tracked in a different JIRA (https://issues.redhat.com/browse/APPDUX-724) and will be handled in a separate PR as they will probably require backend work.

## Verification Steps

1. Go to User federation and select an LDAP card.
2. Enable (if disabled) or disable (if enabled) the switch and click **Save**.
3. Verify that the LDAP card you enabled/disabled now appears on the User Fed card view with the correct tag of enabled or disabled.
4. Click the card and verify that the Enabled/Disabled switch is correct.
5. Select **Action > Delete provider** to delete the user fed provider.
6. Go back to the User federation card view and verify that the card is now gone and the provider has been deleted.

Marvel spec:
https://marvelapp.com/prototype/ecjc94b/screen/72379134

## Checklist:

## Additional Notes
Screen cap:
LDAP-enable-action

### DIFF
--- a/src/user-federation/UserFederationLdapSettings.tsx
+++ b/src/user-federation/UserFederationLdapSettings.tsx
@@ -41,7 +41,7 @@ const LdapSettingsHeader = ({
   onChange,
   value,
   toggleDeleteDialog,
-  toggleRemoveUsersDialog
+  toggleRemoveUsersDialog,
 }: LdapSettingsHeaderProps) => {
   const { t } = useTranslation("user-federation");
   const [toggleDisableDialog, DisableConfirm] = useConfirmDialog({
@@ -59,22 +59,31 @@ const LdapSettingsHeader = ({
         titleKey="LDAP"
         subKey=""
         dropdownItems={[
-          <DropdownItem key="sync" onClick={() => console.log("Sync users TBD")}>
-          Sync changed users
-        </DropdownItem>,
-          <DropdownItem key="syncall" onClick={() => console.log("Sync all users TBD")}>
-          Sync all users
-        </DropdownItem>,
-          <DropdownItem key="unlink" onClick={() => console.log("Unlink users TBD")}>
-          Unlink users
-        </DropdownItem>,
+          <DropdownItem
+            key="sync"
+            onClick={() => console.log("Sync users TBD")}
+          >
+            {t("syncChangedUsers")}
+          </DropdownItem>,
+          <DropdownItem
+            key="syncall"
+            onClick={() => console.log("Sync all users TBD")}
+          >
+            {t("syncAllUsers")}
+          </DropdownItem>,
+          <DropdownItem
+            key="unlink"
+            onClick={() => console.log("Unlink users TBD")}
+          >
+            {t("unlinkUsers")}
+          </DropdownItem>,
           <DropdownItem key="remove" onClick={() => toggleRemoveUsersDialog()}>
-          Remove imported
-        </DropdownItem>,
+            {t("removeImported")}
+          </DropdownItem>,
           <DropdownItem key="delete" onClick={() => toggleDeleteDialog()}>
-          {t("deleteProvider")}
-        </DropdownItem>,
-    ]}
+            {t("deleteProvider")}
+          </DropdownItem>,
+        ]}
         isEnabled={value === "true"}
         onToggle={(value) => {
           if (!value) {
@@ -133,7 +142,7 @@ export const UserFederationLdapSettings = () => {
     continueButtonLabel: "common:remove",
     onConfirm: async () => {
       try {
-        console.log("Imported users removed.");
+        console.log("Remove imported TBD");
         // TODO await remove imported users command
         addAlert(t("removeImportedUsersSuccess"), AlertVariant.success);
       } catch (error) {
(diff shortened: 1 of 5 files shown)


src/user-federation/messages.json
```diff
@@ -92,6 +92,11 @@
     "userFedDisableConfirmTitle": "Disable user federation provider?",
     "userFedDisableConfirm": "If you disable this user federation provider, you cannot make changes to the provider.",
```
ssilvert, line 93: Where did you get this text?  You can still make changes to the provider configuration when it is disabled.  Look at the help text for "Enabled" in the old console.

src/user-federation/UserFederationLdapSettings.tsx
```diff
@@ -15,16 +16,86 @@ import { LdapSettingsSynchronization } from "./ldap/LdapSettingsSynchronization"
...
+import { useAdminClient } from "../context/auth/AdminClient";
+import { useAlerts } from "../components/alert/Alerts";
+import { useTranslation } from "react-i18next";
+import { ViewHeader } from "../components/view-header/ViewHeader";
+import { useHistory, useParams } from "react-router-dom";
+import { ScrollForm } from "../components/scroll-form/ScrollForm";
+
+type LdapSettingsHeaderProps = {
+  onChange: (...event: any[]) => void;
+  value: any;
```
ssilvert, line 35: Avoid `any`

ssilvert (changes requested):

ssilvert:
In the design for the dropdown, there is a line spacer between `Remove imported` and `Delete provider`.

src/user-federation/messages.json
```diff
@@ -92,6 +92,11 @@
...
 
+    "removeImportedUsers": "Remove imported users?",
+    "removeImportedUsersMessage": "Do you really want to remove all imported users? The option \"Unlink users\" makes sense just for the Edit Mode \"Unsynced\" and there should be a warning that \"unlinked\" users without the password in the Keycloak database won't be able to authenticate.",
```
jenny-s51, line 96: This dialog message seems a bit too verbose. I think the second part can be removed with the understanding that once "Unlink Users" is implemented, the dialog can specify the password authentication warning there? Those are  my thoughts ... WDYT? I know this is how it looks in the mocks but I think we could potentially clean it up somehow
```suggestion
    "removeImportedUsersMessage": "Do you really want to remove all imported users? 
```

jenny-s51 (approved):
enable/disable toggle and actions dropdown look good @mfrances17 ! See question about the dialog message below

src/user-federation/UserFederationLdapSettings.tsx
mfrances17: I was getting errors unless using any. This is how it's done in clients and realms as well.

src/user-federation/messages.json
mfrances17: I completely agree that the wording is extremely weak, but I think it's a slippery slope on editing the text from the mocks because I can see making tens if not hundreds of changes. I think we leave that to the doc resource and/or a text sweep later in the project.

src/user-federation/messages.json
mfrances17: Added the text from the help in old console

mfrances17:
> In the design for the dropdown, there is a line spacer between `Remove imported` and `Delete provider`.

Added a separator

src/user-federation/messages.json
jenny-s51: Sounds like a good plan!

src/user-federation/UserFederationLdapSettings.tsx
ssilvert: This works but we could probably do better:
```suggestion
  onChange: (...event: unknown[]) => void;
  value: "true" | null;
```

Then you can also change line 183 to say:
`onChange={(value) => onChange(value)}`

ssilvert:
@mfrances17 You should not have to scroll down and hit `Save` button after doing enable/disable.  It should be changed immediately as this UI treats it as an action and not part of the user's form.  See how it is implemented in Clients.

mfrances17:
> @mfrances17 You should not have to scroll down and hit `Save` button after doing enable/disable. It should be changed immediately as this UI treats it as an action and not part of the user's form. See how it is implemented in Clients.

@ssilvert I don't get it... for everything else, we say make it work like the old UI. In the old UI, you needed to click Save or the Enable or Disable didn't work. So I guess I will rework this.

edewit (changes requested):
created PR with the changes: https://github.com/mfrances17/keycloak-admin-ui/pull/2

mfrances17:
> created PR with the changes: [mfrances17#2](https://github.com/mfrances17/keycloak-admin-ui/pull/2)

thanks... i actually had a fix going in today but wish i saw this sooner, yours is more concise, going with your solution, ty

mfrances17:
@ssilvert @edewit added all PR review changes/suggestions... please review and approve your requested changes and let's get this merged if possible

edewit (approved):

ssilvert (approved):

PR #366, opened 2021-02-15
fixed search text

edewit:
fixes: #363

### DIFF
--- a/src/user/messages.json
+++ b/src/user/messages.json
@@ -2,7 +2,7 @@
   "users": {
     "title": "Users",
     "searchForUser": "Search user",
-    "startBySearchingAUser": "Start by searching a user",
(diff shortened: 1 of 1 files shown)


jenny-s51 (commented):
I think "Start by searching _for_ a user" makes more sense as you can only search for one initially. What do you think?

That's also a closer match to the text on the mock for this screen: https://marvelapp.com/prototype/7i7ja65/screen/75693209

mfrances17 (approved):
LGTM... i can see arguments for both sides whether it be singular or plural. If going plural i would change the key name.

ssilvert (approved):
IMO, plural or singular is fine.  These searches often return more than one.  Adding the "for" was the important part to make it proper English.

PR #365, opened 2021-02-15
fixing round errors in paginating

edewit:
fixes: #364

PR #367, opened 2021-02-15
Realm roles: fix pagination by 10

jenny-s51:
## Motivation

#365 : Upon reviewing/testing Erik's PR I noticed a bug in the realm roles section when paginating by 10 items per page. This PR addresses and fixes the bug described below.

## Brief Description

When a user would select 10 items per page, it was possible to infinitely paginate with the same data showing on every page. 

## Checklist:

### DIFF
--- a/src/realm-roles/RealmRolesSection.tsx
+++ b/src/realm-roles/RealmRolesSection.tsx
@@ -6,9 +6,9 @@ import { RolesList } from "./RolesList";
 
 export const RealmRolesSection = () => {
   const adminClient = useAdminClient();
-  const loader = async (to?: number, max?: number, search?: string) => {
+  const loader = async (first?: number, max?: number, search?: string) => {
     const params: { [name: string]: string | number } = {
-      to: to!,
+      first: first!,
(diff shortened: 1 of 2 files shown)


edewit (changes requested):
What was the error that you got? Because the loader is passed in right?

it is here: https://github.com/keycloak/keycloak-admin-ui/blob/678ec591d61f97b0cf7b841a4ab252c8b918a07d/src/realm-roles/RealmRolesSection.tsx#L9-L16

edewit:
I see the fix would be to rename `to` to `first` because that what it's should be called for the `find` function

jenny-s51:
@edewit thank you for your feedback, that makes more sense - updated

ssilvert:
I imported 50 realm roles without this fix and there was no pagination.  Then I used this PR, which seems to have fixed it.  @edewit @jenny-s51 if you want to use the same import file I'm attaching it on Zulip.  Just go to the old console and do the import from there.

edewit (approved):

PR #369, opened 2021-02-16
updated to match ui design

edewit:
fixes: #323

### DIFF
--- a/src/components/help-enabler/HelpHeader.tsx
+++ b/src/components/help-enabler/HelpHeader.tsx
@@ -7,6 +7,7 @@ import {
   Split,
   SplitItem,
   Switch,
+  TextContent,
 } from "@patternfly/react-core";
 import { Trans, useTranslation } from "react-i18next";
 import { HelpIcon, ExternalLinkAltIcon } from "@patternfly/react-icons";
--- a/src/components/help-enabler/HelpHeader.tsx
+++ b/src/components/help-enabler/HelpHeader.tsx
@@ -71,10 +71,7 @@ export const HelpHeader = () => {
         </SplitItem>
       </Split>
       <TextContent className="keycloak_help-header-description">
-        <Trans>
-          This toggle will enable / disable part of the help info in the
-          console. Includes any help text, links and popovers.
-        </Trans>
+        {t("common-help:helpToggleInfo")}
       </TextContent>
     </DropdownItem>,
   ];
--- a/src/components/help-enabler/HelpHeader.tsx
+++ b/src/components/help-enabler/HelpHeader.tsx
@@ -69,12 +69,12 @@ export const HelpHeader = () => {
           />
         </SplitItem>
       </Split>
-      <span className="keycloak_help-header-description">
+      <TextContent className="keycloak_help-header-description">
         <Trans>
           This toggle will enable / disable part of the help info in the
           console. Includes any help text, links and popovers.
         </Trans>
-      </span>
+      </TextContent>
     </DropdownItem>,
   ];
   return (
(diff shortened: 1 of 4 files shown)


src/components/help-enabler/HelpHeader.tsx
```diff
@@ -51,27 +54,27 @@ export const HelpHeader = () => {
...
             onChange={() => help.toggleHelp()}
           />
         </SplitItem>
       </Split>
+      <span className="keycloak_help-header-description">
+        <Trans>
+          This toggle will enable / disable part of the help info in the
+          console. Includes any help text, links and popovers.
+        </Trans>
+      </span>
```
jenny-s51, line 77:
```suggestion
      <TextContent className="keycloak_help-header-description">
        <Trans>
          This toggle will enable / disable part of the help info in the
          console. Includes any help text, links and popovers.
        </Trans>
      </TextContent>
```

src/components/help-enabler/help-header.css
```diff
@@ -0,0 +1,15 @@
+
+
+.keycloak_help-header-description {
+  padding-top: var(--pf-global--spacer--lg);
+  overflow-wrap: break-word;
+  white-space: normal;
+}
```
jenny-s51, line 7:
```suggestion
.keycloak_help-header-description {
  padding-top: var(--pf-global--spacer--lg);
  overflow-wrap: break-word;
  white-space: normal;
  color: var(--pf-global--Color--100)
}
```

jenny-s51 (commented):
The lg spacer doesn't seem to work with `span` and the help text is a bit tight.

Before suggested changes:

Screen Shot 2021-02-16 at 3 54 12 PM

After suggested changes:

Screen Shot 2021-02-16 at 3 49 58 PM

edewit:
cool thanks @jenny-s51

src/components/help-enabler/HelpHeader.tsx
```diff
@@ -51,27 +55,27 @@ export const HelpHeader = () => {
...
             aria-label="Help is enabled"
             isChecked={help.enabled}
+            label=""
+            className="keycloak_help-header-switch"
             onChange={() => help.toggleHelp()}
           />
         </SplitItem>
       </Split>
+      <TextContent className="keycloak_help-header-description">
+        <Trans>
```
mfrances17, line 74: I know it's not really part of this particular bug fix, but why are we using a <Trans> tag here? This is the only place in the entire console UI that we don't use 't' for localization. Unless there is a specific reason for it, we should probably use the other method because inconsistencies will bite us, if for example, we wanted the ability to switch languages on the fly or something similar. 

Also, as-is i don't think the trans tag will do anything, i think it needs an i18n key.

mfrances17 (approved):
LGTM... small nit in that the text doesn't exactly match the spec but i'm assuming that's by choice. Also, the fix for the 40px spacing is good, but i'm not fond of the text in the explanation going farther to the right than the switch, it looks scattered... i would consider right-alignment regardless of the exact 40px spacing. But this does what they are asking for, so approving.

jenny-s51 (approved):
Looks good! I agree with @mfrances17  that the alignment for the text and switch should be fixed eventually

src/components/help-enabler/HelpHeader.tsx
ssilvert: > I know it's not really part of this particular bug fix, but why are we using a Trans tag here? This is the only place in the entire console UI that we don't use 't' for localization. Unless there is a specific reason for it, we should probably use the other method because inconsistencies will bite us, if for example, we wanted the ability to switch languages on the fly or something similar.
> 
> Also, as-is i don't think the trans tag will do anything, i think it needs an i18n key.

@edewit Did you respond to this privately?  Not sure why @mfrances17 approved without this being addressed?

PR #371, opened 2021-02-16
User fed: Add Kerberos provider and tests

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-807

## Brief Description
Decision in the 2/2/21 keycloak status meeting was to move forward with functionality to add new providers directly using the 'Add new provider' dropdown in the User Federation settings view similar to the way it was done in the old admin console UI, instead of via the wizards, at least for now. 

Note that this PR also covers creating a new kerberos provider via the user fed empty state page as well, by clicking the 'Add Kerberos providers' card.

Also included in this PR is a suite of cypress tests for the kerberos user federation as a whole, which includes tests for:

- Creating kerberos provider from empty state
- Creating kerberos provider from the card view
- Changing an existing kerberos provider and saving (update)
- Changing an existing kerberos provider and cancelling (cancel)
- Disabling a kerberos provider
- Enabling a previously-disabled kerberos provider
- Deleting a kerberos provider using the card menu
- Deleting a kerberos provider using the settings view Action menu

## Verification Steps
1. Go to user federation with no kerberos providers configured.
2. Click the 'Add Kerberos providers' card and verify that you can successfully create a new kerberos provider using the form that follows.
3. With at least one kerberos provider configured, verify that you can successfully create a new kerberos provider using the Add new provider > Kerberos command.
4. Bonus points: run yarn start:cypress and verify that the kerberos provider tests all pass.

## Checklist:

### DIFF
--- a/src/client-scopes/add/__tests__/__snapshots__/MapperDialog.test.tsx.snap
+++ b/src/client-scopes/add/__tests__/__snapshots__/MapperDialog.test.tsx.snap
@@ -21009,7 +21009,6 @@ exports[`<MapperDialog/> should return selected protocol mapping type on click 1
                             <ul
                               aria-label="chooseAMapperType"
                               className="pf-c-data-list pf-m-compact pf-m-grid-md"
-                              style={Object {}}
                             >
                               <DataListItem
                                 aria-labelledby="User Realm Role"
--- a/src/client-scopes/add/__tests__/__snapshots__/MapperDialog.test.tsx.snap
+++ b/src/client-scopes/add/__tests__/__snapshots__/MapperDialog.test.tsx.snap
@@ -21009,6 +21009,7 @@ exports[`<MapperDialog/> should return selected protocol mapping type on click 1
                             <ul
                               aria-label="chooseAMapperType"
                               className="pf-c-data-list pf-m-compact pf-m-grid-md"
+                              style={Object {}}
                             >
                               <DataListItem
                                 aria-labelledby="User Realm Role"
--- a/src/components/keycloak-card/KeycloakCard.tsx
+++ b/src/components/keycloak-card/KeycloakCard.tsx
@@ -67,7 +67,7 @@ export const KeycloakCard = ({
             />
           )}
         </CardActions>
-        <CardTitle>{title}</CardTitle>
+        <CardTitle data-cy="keycloak-card-title">{title}</CardTitle>
       </CardHeader>
       <CardBody />
       <CardFooter>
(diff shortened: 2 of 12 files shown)


edewit (approved):
:+1:

jenny-s51 (commented):
Nice work @mfrances17 ! This is looking great. 

A couple things I noticed: 

1) When I first ran your Cypress tests, my User Fed section wasn't in an empty state and most of the Cypress tests didn't pass. but when I deleted every provider and ran again they all passed. In order to avoid CI issues in the future, it might be a good idea to write a test for when there are already some existing providers. 

2) If I make a new Kerberos provider and click "Save" multiple times, I get multiple versions of the same provider. Maybe we can disable the save button when the provider is initially created and set it to be enabled when any of the fields are dirty?

WDYT? 

If you'd like to address those in separate PRs I'm happy to approve this 🙂

mfrances17:
Thanks for the review, Jenny! Good stuff.

> 1. When I first ran your Cypress tests, my User Fed section wasn't in an empty state and most of the Cypress tests didn't pass. but when I deleted every provider and ran again they all passed. In order to avoid CI issues in the future, it might be a good idea to write a test for when there are already some existing providers.

Right, these tests were written based on the assumption of a clean system, then it creates and tests a bunch of stuff, then cleans up after. It brings up some good questions about test hardening, if I can assume CI will start clean, if it needs to run and pass locally for devs, etc. I'm not sure of the answers, thinking on it, I'm wondering if the best thing to do would be to detect if providers exist and then skip the empty state test rather than fail it, and then proceed on a slightly different test path. Not sure.

> 2. If I make a new Kerberos provider and click "Save" multiple times, I get multiple versions of the same provider. Maybe we can disable the save button when the provider is initially created and set it to be enabled when any of the fields are dirty?
On this one, I don't think disabling the button and re-enabling it when the fields are dirty will fix it - it will still do a create action with the same name potentially. I've disliked how the 'create' action doesn't bring you back to the card view - i can see a 'save' staying on settings, but not a create. I think the best thing to do here might be to nav to the cards view after create. I think we might want to change the text of 'Save' button to 'Create' in this case. Also, there is another issue that we should probably only enable Save on dirty... 

> If you'd like to address those in separate PRs I'm happy to approve this 🙂
I think that might make sense. I'll file three issues and refer out to them from here.

mfrances17:
Added the following follow-up issues to address @jenny-s51 comments:
https://github.com/keycloak/keycloak-admin-ui/issues/374
https://github.com/keycloak/keycloak-admin-ui/issues/375
https://github.com/keycloak/keycloak-admin-ui/issues/376

jenny-s51 (approved):
Awesome, thank you for making those follow-up issues @mfrances17 ! That sounds like a good game plan for the tests and the create/save button functionality 👍

PR #373, opened 2021-02-17
Advanced tab

edewit:
## Motivation

Introduced advanced tab for the client section [marvel](https://marvelapp.com/prototype/9a93gef)

## Verification Steps

Open the details of a client and see that it now has an extra tab called "Advanced"
The clients of type "openid-connect" have a different tab then the once that are of type SAML

## Checklist:


ssilvert (changes requested):
@edewit Just finished testing.  Looking much better.  A couple of things:

- Reload still not working on advanced settings
- Authentication Flow Overrides should not allow selection of flows where providerId === 'client-flow'.  See https://github.com/keycloak/keycloak/blob/9356843c6c3d7097d010b3bb6f91e25fcaba378c/themes/src/main/resources/theme/base/admin/resources/js/controllers/clients.js#L1037  In this code, it is putting client flows and other flows into different arrays.  But I can't find anywhere that the clientFlows array is used.  So it looks like the proper fix is to just exclude client flows from the list.  I'll admit, I don't actually know why the logic is there in the first place.  You may want to ping the keycloak-dev list.

edewit:
@ssilvert okay maybe we need a definition of what reload should do? IMO it should set the controls back to the saved version. If no value was set then it can't / needs to rest the value, right?

edewit:
@ssilvert done

ssilvert:
> @edewit Just finished testing. Looking much better. A couple of things:
> 
> * Reload still not working on advanced settings
> * Authentication Flow Overrides should not allow selection of flows where providerId === 'client-flow'.  See https://github.com/keycloak/keycloak/blob/9356843c6c3d7097d010b3bb6f91e25fcaba378c/themes/src/main/resources/theme/base/admin/resources/js/controllers/clients.js#L1037  In this code, it is putting client flows and other flows into different arrays.  But I can't find anywhere that the clientFlows array is used.  So it looks like the proper fix is to just exclude client flows from the list.  I'll admit, I don't actually know why the logic is there in the first place.  You may want to ping the keycloak-dev list.

* Reload on advanced settings still not working for Access Token Lifespan.
* Values for Authentication flow overrides still don't match the old console.  Looking at the "account" client, the dropdown shows "http challenge" in the old console but it doesn't appear in the new console.  Also, in the old console the values are sorted.

edewit:
@ssilvert changed the time selector back to use a normal input type="number" to allow users to not set it (e.g. clear out the value), so that we have 2 states just like the old console.

ssilvert:
> @ssilvert changed the time selector back to use a normal input type="number" to allow users to not set it (e.g. clear out the value), so that we have 2 states just like the old console.

It still doesn't allow a value -1 like the old console does.

ssilvert:
@edewit I went ahead and merged this but opened an issue so we can come back to it later.
https://github.com/keycloak/keycloak-admin-ui/issues/405

PR #387, opened 2021-02-19
Search and create sub groups

edewit:
## Motivation

Adds the ability to create groups within groups and a "global" search as defined in [marvel](https://marvelapp.com/prototype/228971ig/screen/72210499)

## Verification Steps

Go to the groups section create a group and then click on the group name.
Now you can create another group that will be the child of that pref group.
Click the action dropdown and go to search to search for groups

## Checklist:


tests/cypress/support/commands.ts
```diff
@@ -23,3 +23,7 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add("getBy", (selector, ...args) => {
+  return cy.get(`[data-testid=${selector}]`, ...args);
```
mfrances17, line 28: The question is... does this save enough time to justify obscuring what cypress is doing? I'm not sure. But code looks fine. Was wondering if 'getBy' might be a little vague - maybe cy.getId or cy.getById?

tests/cypress/support/pages/ViewHeaderPage.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types=".." />
+
+export default class ListingPage {
+  private actionMenu = "viewheader-action";
+
+  clickAction(action: string) {
+    cy.get("[data-cy=action-dropdown]").click().getBy(action).click();
```
mfrances17: why use getBy for one but not the other? also, should be data-testid

mfrances17 (commented):
Couple comments just on the testing aspect... will revisit to review actual search and create sub groups code and test it.

tests/cypress/support/pages/ViewHeaderPage.ts
edewit: didn't use `getBy` as it's not using `data-testid` it was already there, but I'll rename

jenny-s51 (commented):
Subgroup functionality looks good @edewit ! 

From the mock it looks like when a group is clicked from the initial groups list, the subgroups should render in a separate "Child Groups" tab - it's fine if you want to address this in another PR.

Also, the empty state for subgroups is the same as the empty state for groups currently, which contains the text "no groups in this realm". It might be a good idea to create a separate empty state message for subgroups, e.g. "you have not created any child groups".

edewit:
@jenny-s51 I'll create the tabs in a separate PR, but changed the empty text for a empty sub group

PR #391, opened 2021-02-22
Realm roles: adds "Inherited From" column to Associated Roles tab

jenny-s51:
## Motivation

#285 

## Verification Steps

1. Go to Realm Roles section. 
2. Create a new role or click on an existing role.
3. Click on the "Action" dropdown on the top right.
4. Add "admin" as an associated role.
5. Verify that the associated roles tab now contains a list of all the roles associated with the admin role, and that for each of these roles "admin" is listed in the "Inherited From" column
6. Click "Hide inherited roles" and verify that these inherited roles are hidden when checked

## Checklist:


src/realm-roles/RealmRoleTabs.tsx
```diff
@@ -56,6 +57,7 @@ export const RealmRoleTabs = () => {
 
   const adminClient = useAdminClient();
   const [role, setRole] = useState<RoleFormType>();
+  const [, setClient] = useState<ClientRepresentation>();
```
edewit, line 60: if you only use a `set` then you don't need the state, right?

edewit (approved):
looks good

Issue #355, opened 2021-02-05
[UX review] Client - Import client


xianli123:
Hi, @edewit I've reviewed the import client page and made some fine-tuning.

- "**Import**" button in the Client list
Clicking the "**Import**" button, there will be a modal popup, as shown in the picture below. After selecting the file and clicking the "Open" button, users will jump into the "_Import client_" page.  
<img width="1266" alt="Screen Shot 2021-02-05 at 4 52 14 PM" src="https://user-images.githubusercontent.com/53931519/107011502-ad1bb780-67d2-11eb-86b0-cc5d7cb717c8.png">

- Import client page
1. It would be nice if there was a dividing line.   
2. The length of this form should be 720 px. 
3. We can add question marks to help users better understand the tags.   
4. The Description field uses the text area component.   
5. After clicking the **Save** button, users will enter the detail page of the "Import client".   
6. Click the **Cancel** button to return the client list page, but it doesn't work now. 
7. **Capability configs** - There are some different capability configs based on the different types above.

![PageBreak_1](https://user-images.githubusercontent.com/53931519/107007116-0d0f5f80-67cd-11eb-988f-336089e08cb2.png)

The Marvel file of the "[Import client](https://marvelapp.com/prototype/7aaf345/screen/72130125)" shows more details for your reference.

edewit:
what are the capability configs? The marvel shows 2 and says it varies on client type, as far as I know there are 2 client types (openid-connect and saml) so what is shown when the type is openid-connect?

xianli123:
We now have 2 types of clients namely SAML2.0 and OIDC. The capability configs are shown as below:

SAML 2.0
<img width="667" alt="Screen Shot 2021-02-23 at 11 08 13 AM" src="https://user-images.githubusercontent.com/53931519/108808909-acf92580-75e2-11eb-8dfe-7f0a0c78b20e.png">

OIDC
<img width="564" alt="Screen Shot 2021-02-23 at 11 08 54 AM" src="https://user-images.githubusercontent.com/53931519/108808927-b7b3ba80-75e2-11eb-9033-039eb3a06556.png">

In the future, some new types will be added to the admin console. And then the capability configs are going to change depending on the type.

Issue #392, opened 2021-02-23
openid-connect client create gives 500 error


edewit:
When you create a openid-connect client with the wizard and set the switches both to on on the second page you get a 500 error:

![image](https://user-images.githubusercontent.com/51133/108811195-1c542280-75ad-11eb-8abc-8ffa3cac220f.png)

in the server log:

```
07:57:53,724 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (default task-133) Uncaught server error: java.lang.RuntimeException: Client does not have a service account.
	at org.keycloak.keycloak-services@12.0.1//org.keycloak.authorization.admin.ResourceServerService.create(ResourceServerService.java:83)
	at org.keycloak.keycloak-services@12.0.1//org.keycloak.authorization.admin.AuthorizationService.enable(AuthorizationService.java:60)
	at org.keycloak.keycloak-services@12.0.1//org.keycloak.services.resources.admin.ClientsResource.createClient(ClientsResource.java:194)
```

edewit:
with the new defaults the error changed a bit:

```
10:58:09,462 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (default task-81) Uncaught server error: java.lang.RuntimeException: Only confidential clients are allowed to set authorization settings
```

PR #393, opened 2021-02-23
Fixed issue using import

edewit:
fixes: #355


jenny-s51 (commented):
The client id and description text fields are populating correctly based on the imported data. what is the best way to test the import itself @edewit ? 

I created a sample JSON file to resemble some of the attributes in the mock: 

```
{
    "clientId": "test",
    "description": "test description",
    "adminUrl": "",
    "baseUrl": "asjdlajsd.com",
    "surrogateAuthRequired": false
    
}
```

But when I click "save" I get a 500 error.

edewit:
you can create an export of an existing client, by clicking on the action menu - export:

image

And then delete that client and use the downloaded json to recreate it.

ssilvert (changes requested):
- When I try to drag a file to import, I get "Are you sure you want to clear this file?" dialog.
- I don't think you should be able to edit the json after importing, but I'm open to opposing views.  Just seems error-prone.
- Client authorization can only be set to true if accessType=confidential.  So if you set it and try to save a public client you will get an error.  Look at accessType in the old console.  There are other things that only apply to certain access types.  This is also a problem in the wizard.
- BTW, if you are wondering how you know what the accessType is from the json, it appears to be determined with a combination of the bearerOnly and publicClient flags.
- If I go to the details screen and save something, the settings for Capability config disappear.

edewit:
@ssilvert so this is an error in the design as currently there is no accessType

edewit:
@ssilvert so this is a error in the UI design we need this accessType dropdown on the pages? @yih-wang?

edewit:
can we merge this @ssilvert there is an issue with regards to the Client authorization #392

ssilvert (approved):
We can merge this and handle the rest of the feedback as issue.

PR #394, opened 2021-02-23
removed need of refresher

edewit:


jenny-s51 (approved):
nice update - works as expected 👍

Issue #376, opened 2021-02-17
User fed Kerberos provider - cypress test hardening


mfrances17:
The cypress tests for the user federation kerberos provider section assume that the system is clean and has no previously-configured kerberos providers on the system. The tests should allow for testing on systems that already have kerberos providers configured. See https://github.com/keycloak/keycloak-admin-ui/pull/371 for more info.

Issue #375, opened 2021-02-17
Prevent multiple saves of user federation kerberos provider during create


mfrances17:
When creating a new user fed kerberos provider, you can click the Save button multiple times to create multiple providers with the same name. See https://github.com/keycloak/keycloak-admin-ui/pull/371 for more info.

PR #400, opened 2021-02-23
User fed: Add LDAP provider and tests

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-806
https://issues.redhat.com/browse/APPDUX-847

## Brief Description
Decision in the 2/2/21 keycloak status meeting was to move forward with functionality to add new providers directly using the 'Add new provider' dropdown in the User Federation settings view similar to the way it was done in the old admin console UI, instead of via the wizards, at least for now.

Note that this PR also covers creating a new LDAP provider via the user fed empty state page as well, by clicking the 'Add LDAP providers' card.

Also included in this PR is a suite of cypress tests for the LDAP user federation provider as a whole, which includes tests for:

- Creating an LDAP provider from empty state
- Creating an LDAP provider from the card view
- Changing an existing LDAP provider and saving (update)
- Changing an existing LDAP provider and canceling (cancel)
- Disabling an LDAP provider
- Enabling a previously-disabled LDAP provider
- Deleting an LDAP provider using the card menu
- Deleting an LDAP provider using the settings view Action menu

## Verification Steps

1. Go to user federation with no LDAP providers configured.
2. Click the 'Add LDAP providers' card and verify that you can successfully create a new LDAP provider using the form that follows.
3. With at least one LDAP provider configured, verify that you can successfully create a new LDAP provider using the Add new provider > LDAP command.
4. Bonus points: run yarn start:cypress and verify that the LDAP provider tests all pass.

## Checklist:

## Additional Notes
The follow-up issues that were filed for the kerberos provider and tests PR (https://github.com/keycloak/keycloak-admin-ui/pull/371) also apply to this PR, and will be implemented as part of those fixes:
https://github.com/keycloak/keycloak-admin-ui/issues/374 (don't save form unless dirty)
https://github.com/keycloak/keycloak-admin-ui/issues/375 (don't allow multiple saves during create)
https://github.com/keycloak/keycloak-admin-ui/issues/376 (change tests to allow for existing providers)


edewit:
@mfrances17 don't we want to use the page objects that @Aboullos introduced

jenny-s51 (commented):
It looks like Patternfly recently released a `NumberInput` component, which is very exciting! 

Setting `TextInput` to type number seems like a reasonable way to go about the numeric fields too, although the user can currently set negative values for all numeric fields, which could be an issue. The `NumberInput` has a `min` prop that takes care of that. WDYT @mfrances17 ?

edewit:
A `<input type="number"` also has a `min` property so we could also add `min` to the `<TextInput` component. I've used the `NumberInput` already and it's a drop in replacement for `<TextInput` all that is needed is add handlers for the `+` and `-` buttons

mfrances17:
@edewit @jenny-s51 added minimum values to number fields, and greatly simplified cypress tests by reusing and abstracting code to the page object, please re-review

jenny-s51 (approved):
LGTM @mfrances17 - LDAP provider creation, updates to the provider, and min value fields work as expected with Cypress tests passing 👍

PR #401, opened 2021-02-24
Realm roles: add "users in role" tab

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-843

## Brief Description

Adds "Users in role" tab to realm roles section of Keycloak.

## Verification Steps

1. Go to Realm Roles section.
2. Create a new realm role.
3. Click on the "Users in role tab".
4. Verify that the empty state appears.
5. In the old console, go to Users -> Add user. 
6. Make sure to fill out each of the fields Username, Email, First Name, and Last Name
7. Click save and go to the Role Mappings tab.
8. Select the realm role you created in the "Realm roles" box and click "Add selected".
9. Go back to the admin console and refresh.
10. Select the "Users in Role" tab again and verify that the role information displays in the data table. 

## Checklist:


cypress/integration/realm_roles_test.spec.ts
```diff
@@ -87,21 +87,26 @@ describe("Realm roles test", function () {
...
-      cy.get('[type="checkbox"]').eq(4).check({force: true});
+      cy.get('[type="checkbox"]').eq(4).check({ force: true });
 
       cy.get("#add-associated-roles-button").contains("Add").click();
 
       cy.wait(2500);
+
+      cy.contains("Users in role")
+        .click()
+        .get("#users-empty-state > div > h4")
```
edewit, line 108: wouldn't we rather use `data-testid` instead of `id` or `data-cy`?

src/realm-roles/UsersInRoleTab.tsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  useEffect(() => {
+    loader();
+  }, []);
+
+  return (
+    <>
+      <PageSection variant="light">
+        {users.length == 0 ? (
```
edewit, line 36: you don't need this check the `<KeycloakDataTable` has a emptyState Property just add the empty state to that like this:

```tsx
      <KeycloakDataTable
          ....
          emptyState={
            <ListEmptyState
              hasIcon={true}
              message={t("noDirectUsers")}
              instructions={t("noUsersEmptyStateDescription")}
              id="users-empty-state"
          />
          }
      />
```

edewit (changes requested):

src/realm-roles/UsersInRoleTab.tsx
jenny-s51: Thank you for suggesting this @edewit - I tried this initially but it seems the tab should only contain an empty state as shown here https://marvelapp.com/prototype/f39d464/screen/72685385

Adding it to the `emptyState` prop of `KeycloakDataTable` also renders a table toolbar with a search input field, but from the mock it looks like we don't want any toolbar or search input on the empty state.

cypress/integration/realm_roles_test.spec.ts
ssilvert: @edewit  @jenny-s51 @mfrances17 We do need to standardize this.  Did we all agree that we would use data-testid?

edewit: we agreed on `data-testid` as it is less bound to the testing framework `cy` is derived from cypress

src/realm-roles/UsersInRoleTab.tsx
edewit: yes I've seen that and that and I have a fix for that in #387 but that is not merged yet

cypress/integration/realm_roles_test.spec.ts
jenny-s51: sounds good, will update with `data-testid`

src/realm-roles/UsersInRoleTab.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const role = await adminClient.roles.findOneById({ id: id });
+    const usersWithRole = await adminClient.roles.findUsersWithRole({
+      name: role.name!,
+    });
+    return usersWithRole;
+  };
+
+  useEffect(() => {
+    loader();
+  }, []);
```
edewit, line 32: This is not needed right?

edewit (changes requested):

src/realm-roles/UsersInRoleTab.tsx
jenny-s51: nice catch - updated

ssilvert (commented):
- Needs pagination, but we can do that in another PR if you want.
- The help text for "Who appears in this list" should be hidden when global help is turned off.

mfrances17 (changes requested):
Code looks good, followed the testing steps and it worked as expected.

Two comments that are both test-related:
I cannot get the final cypress test to pass, it fails on this line every time:

`cy.get("[data-testid=add-role-button]").click();`

I see you set a timeout before that action but apparently it's not enough. Maybe this can 
 be split out into its own test?
 
Also, you may want to consider creating functions in the create realm role page object for some of these test actions and then call them from the test. It seems we're standardizing on this kind of abstraction in the tests.

edewit (approved):
:+1: :1st_place_medal:

jenny-s51:
Thank you for your helpful feedback @mfrances17 ! The cypress tests are now passing and I split out the calls into separate tests like you suggested. Please kindly re-review as soon as you get a chance 🙂

mfrances17 (approved):
LGTM - tests all pass now, good job

Issue #352, opened 2021-02-03
[UX review] Client list screens


xianli123:
Hi, @edewit I've reviewed the Client list page, there are some updates about this part.

First of all, there are two tabs namely "Client list" and "Initial access token" in the latest design. The "Client list" looks like what we have designed before. The "Initial access token" is moved to the **Client page** from "**Realm settings**". Please check the marvel link to dig into the details.
- Client list design: https://marvelapp.com/prototype/f00e9jb/screen/71968754
- Initial access token design: https://marvelapp.com/prototype/fcf2450/screen/74745145

After that, there are some details that need to be updated, as shown in the screenshot below.
1.  It will be good to add the dividing lines.   
2. The space between the **Client ID** and the label should be 8 px.   
3. It would be better to display the whole name. 
4. The link and title are aligned to the left. The font size of the link is different from the other words in the table. The space between the link words and the icon should be 8 px.  
5. In the new design, users can't delete the built-in client provided by Keycloak. Therefore, the **Delete** option in the built-in client is disabled.   
6. The table should fill the entire white page. There is no gap.  

Current implementation:
![ClientList](https://user-images.githubusercontent.com/53931519/106708263-84a68880-662d-11eb-849d-a94df9198d30.png)

Specs:
![ClientList_1](https://user-images.githubusercontent.com/53931519/106708301-9720c200-662d-11eb-9a99-53aaf9bcd96b.png)

Thanks.

edewit:
@xianli123 there is currently no way to know if a client is a build in client.

xianli123:
In the old console, if users created a new realm, they would find that there were some built-in clients. These clients shouldn't be deleted in the new design.
<img width="1193" alt="Screen Shot 2021-02-09 at 11 36 55 AM" src="https://user-images.githubusercontent.com/53931519/107312627-ce7aed00-6acb-11eb-9670-9f5ff7f488eb.png">

edewit:
right, I get what you mean, but there is no api to determine if a client is
a build in one.

xianli123:
Yes. Do you know who can help us to provide the API?

ssilvert:
@xianli123 Did we discuss which clients should not be deleted?  I think there are use cases for removing at least some of them.

Though you could make the argument that disabling serves the same purpose, I'm not sure all customers would agree.

[...]

PR #409, opened 2021-03-02
added space between buttons

edewit:
## Motivation

see: https://github.com/keycloak/keycloak-admin-ui/issues/352#issuecomment-788677779


mfrances17 (approved):
LGTM - space was added and matches the spec more closely, and the cypress group test passes with these changes.

jenny-s51 (approved):
Create and Import button spacing looks good! Group search cypress test also working as expected

PR #425, opened 2021-03-09
Group details

edewit:
## Motivation
Added the group details to the group section

## Verification Steps

Go to the groups section create a couple of groups click on the group see the new group details

## Checklist:

### DIFF
--- a/src/groups/GroupAttributes.tsx
+++ b/src/groups/GroupAttributes.tsx
@@ -1,0 +1,66 @@
+import React, { useEffect } from "react";
+import { useTranslation } from "react-i18next";
+import { useFieldArray, useForm } from "react-hook-form";
+import { AlertVariant } from "@patternfly/react-core";
+
+import { useAlerts } from "../components/alert/Alerts";
+import {
+  arrayToAttributes,
+  AttributeForm,
+  AttributesForm,
+  attributesToArray,
+} from "../components/attribute-form/AttributeForm";
+import { useAdminClient } from "../context/auth/AdminClient";
+
+import { getLastId } from "./groupIdUtils";
+import { useSubGroups } from "./SubGroupsContext";
+
+export const GroupAttributes = () => {
+  const { t } = useTranslation("groups");
+  const adminClient = useAdminClient();
+  const { addAlert } = useAlerts();
+  const form = useForm<AttributeForm>({ mode: "onChange" });
+  const { fields, append, remove } = useFieldArray({
+    control: form.control,
+    name: "attributes",
+  });
+
+  const id = getLastId(location.pathname);
+  const { subGroups } = useSubGroups();
+  const getGroup = () => subGroups[subGroups.length - 1];
+
+  useEffect(() => {
+    const group = getGroup();
+    const attributes = attributesToArray(group.attributes!);
+    attributes.push({ key: "", value: "" });
+    form.setValue("attributes", attributes);
+  }, [id]);
+
+  const save = async (attributeForm: AttributeForm) => {
+    try {
+      const group = getGroup();
+      await adminClient.groups.update(
+        { id: id! },
+        { ...group, attributes: arrayToAttributes(attributeForm.attributes) }
+      );
+      addAlert(t("groupUpdated"), AlertVariant.success);
+    } catch (error) {
+      addAlert(t("groupUpdateError", { error }), AlertVariant.danger);
+    }
+  };
+
+  return (
+    <AttributesForm
+      form={form}
+      save={save}
+      array={{ fields, append, remove }}
+      reset={() =>
+        form.reset({
+          attributes: attributesToArray(
+            subGroups[subGroups.length - 1].attributes
+          ),
+        })
+      }
+    />
+  );
+};
--- a/src/groups/GroupAttributes.tsx
+++ b/src/groups/GroupAttributes.tsx
@@ -26,7 +26,7 @@ export const GroupAttributes = () => {
   });
 
   const id = getLastId(location.pathname);
-  const { currentGroup, subGroups } = useSubGroups();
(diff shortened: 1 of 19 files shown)


src/realm-roles/RealmRolesSection.css
```diff
@@ -1,30 +1,4 @@
...
-  margin-left: calc(var(--pf-global--spacer--md) * -1);
-}
-
-.pf-c-button.kc-role-attributes__minus-icon {
-  /* shift the button left to adjust for table cell padding */
-  margin-left: calc(var(--pf-global--spacer--md) * -1);
-  color: var(--pf-c-button--m-plain--Color);
-}
-
-.kc-add-role-button {
```
jenny-s51, line 18: Looks like we need this styling here

Screen Shot 2021-03-09 at 3 05 18 PM

jenny-s51 (changes requested):
Subgroup functionality is looking good @edewit !

There seems to be a small bug with the `reload()` function in the attributes section...

To reproduce:

1) add a user attribute
2) click 'save'
3) click 'reload'
4) the user attribute field disappears even though the new value should be posted to the server, reload should not remove the field in this case
5) try the same thing in the realm roles section -> when you save an attribute and click 'reload' it does not remove the field for the most recently saved attribute (which is the expected behavior)

src/realm-roles/RealmRolesSection.css
edewit: Oops seems like I moved one of the styles from here to the new component. But I don't think we even need it. I fixed it by wrapping the controls in `<ToolbarItem>`s

src/groups/Members.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <KeycloakDataTable
+      key={key}
+      loader={loader}
+      ariaLabelKey="groups:members"
+      isPaginated
+      toolbarItem={
+        <>
+          <ToolbarItem>
+            <Button data-testid="addMember" variant="primary">
```
mfrances17, line 90: intentionally skipped and will be handled separately?

src/groups/GroupTable.tsx
```diff
@@ -0,0 +1,206 @@
...
+        }
+        actions={[
+          {
+            title: t("moveTo"),
+            onRowClick: () => console.log("TO DO: Add move to functionality"),
+          },
+          {
+            title: t("common:delete"),
+            onRowClick: async (group: GroupRepresentation) => {
+              return deleteGroup(group);
```
mfrances17, line 169: we should provide a confirmation modal on delete, as we do in other parts of the UI

mfrances17 (changes requested):
Found a couple of things. Also, Action > Delete Group and the Add Member button are not working, but that functionality might not be considered part of this PR, if so, ignore.

src/groups/Members.tsx
edewit: yeah this PR was already getting big so I opted to do those in a separate PR

src/groups/GroupTable.tsx
edewit: Right this was already there I'll create an issue for that #430

mfrances17 (approved):

src/groups/GroupAttributes.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  return (
+    <AttributesForm
+      form={form}
+      save={save}
+      array={{ fields, append, remove }}
+      reset={() =>
+        form.reset({
+          attributes: convertAttributes(),
```
jenny-s51, line 60:
```suggestion
          attributes: attributes,
```

src/groups/GroupAttributes.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useTranslation } from "react-i18next";
+import { useFieldArray, useForm } from "react-hook-form";
+import { AlertVariant } from "@patternfly/react-core";
+
+import { useAlerts } from "../components/alert/Alerts";
+import {
+  arrayToAttributes,
+  AttributeForm,
+  AttributesForm,
+  attributesToArray,
```
jenny-s51, line 11:
```suggestion
  attributesToArray,
  KeyValueType,
```

src/groups/GroupAttributes.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { addAlert } = useAlerts();
+  const form = useForm<AttributeForm>({ mode: "onChange" });
+  const { fields, append, remove } = useFieldArray({
+    control: form.control,
+    name: "attributes",
+  });
+
+  const id = getLastId(location.pathname);
+  const { currentGroup, subGroups } = useSubGroups();
+
```
jenny-s51, line 30:
```suggestion
  const [ attributes, setAttributes ] = useState<KeyValueType[]>([]);

```

src/groups/GroupAttributes.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const convertAttributes = (attr?: Record<string, any>) => {
+    const attributes = attributesToArray(attr || currentGroup().attributes!);
```
jenny-s51, line 32:
```suggestion
    const attributes = attributesToArray(attr || currentGroup().attributes!);
    setAttributes(attributes);
```

jenny-s51 (changes requested):
reload is not quite working as expected - when I click "reload" after saving a new attribute the field disappears ... after some investigating/testing these suggested changes fixed the reload for me 🙂

edewit:
thanks @jenny-s51 I fixed it a little different then you suggested... the reload was still broken after a save... I fixed that now as well

jenny-s51 (approved):
LGTM! 👍

PR #427, opened 2021-03-09
Action menu commands for User Fed LDAP settings

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-724

## Brief Description
Add business logic for the following LDAP settings menu commands that were available in the legacy admin console:
* Sync changed users
* Sync all users
* Unlink users
* Remove imported

## Verification Steps
1. Go to user federation and select an LDAP provider card (or create a new one if it doesn't exist).
2. In the action menu, select 'Sync changed users'. Verify that you get an alert with the status.
3. Repeat for the 'Sync all users', 'Unlink users' and 'Remove imported' commands.

## Checklist:

## Additional Notes
There are no cypress tests yet, as we discussed, we will need to set up some kind of mock data or equivalent LDAP config to really test this. These commands will eventually need to be revisited.

### DIFF
--- a/src/user-federation/UserFederationLdapSettings.tsx
+++ b/src/user-federation/UserFederationLdapSettings.tsx
@@ -48,6 +48,8 @@ const LdapSettingsHeader = ({
 }: LdapSettingsHeaderProps) => {
   const { t } = useTranslation("user-federation");
   const { id } = useParams<{ id: string }>();
+  const adminClient = useAdminClient();
+  const { addAlert } = useAlerts();
   const [toggleDisableDialog, DisableConfirm] = useConfirmDialog({
     titleKey: "user-federation:userFedDisableConfirmTitle",
     messageKey: "user-federation:userFedDisableConfirm",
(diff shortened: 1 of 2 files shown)


jenny-s51 (commented):
Error messages for `unlink users` and `remove imported` are looking good! 

For both `sync` options, I'm hitting the `response.ignored` case and my alert looks like this:

Screen Shot 2021-03-09 at 4 01 26 PM

Is this the expected content for the alert?

mfrances17:
> Is this the expected content for the alert?

@jenny-s51 Yes i think so, at least that's what the legacy console displayed. Here's a screen shot of the old UI:

Screen Shot 2021-03-09 at 4 47 59 PM

We could change it.

src/user-federation/UserFederationLdapSettings.tsx
```diff
@@ -67,22 +131,13 @@ const LdapSettingsHeader = ({
           titleKey="LDAP"
           subKey=""
           dropdownItems={[
-            <DropdownItem
-              key="sync"
-              onClick={() => console.log("Sync users TBD")}
-            >
+            <DropdownItem key="sync" onClick={() => syncChangedUsers()}>
```
edewit, line 134: no need to wrap it could just do:
```suggestion
            <DropdownItem key="sync" onClick={syncChangedUsers}>
```

edewit (approved):

jenny-s51 (approved):
@mfrances17 , ok, approving as it matches the alert in the old console, though the text in the success alert confused me a bit - how can the sync have completed if it's ignored/currently in progress? 

I think the alert should either just contain the "sync successful" text or the "ignored/currently in progress" text using maybe the warning alert variant for the ignored case ... what do you think?

src/user-federation/UserFederationLdapSettings.tsx
mfrances17: Thanks, yes agreed. Fixed these in the latest commit.

Issue #423, opened 2021-03-09
[UX review] Client settings


xianli123:
Hi, @edewit I've reviewed the Client settings page, there are some suggestions for this part. 

1. Header: 
- A. The breadcrumb is a little far away from the title. 
- B. There should be a label to indicate the client type.  
- C. All tags should be followed by a question mark.
- D. There is a line separating the **Delete** option from the above options. 

![1](https://user-images.githubusercontent.com/53931519/110443178-9a104600-80f6-11eb-8906-c3a5ab170b17.png)

2. Footer: 
- A. The button should be fixed at the bottom. 
- B. Additionally, the **Cancel** button should be **Reload** button. Clicking **Reload** button will cancel the unsaved changes.   

![2](https://user-images.githubusercontent.com/53931519/110443684-27539a80-80f7-11eb-9ec7-d2b24592becb.png)

3. Capability configs: 
- A. If the "**_Client authentication_**" is "OFF", "**_Authorization_**" and "**_Service account roles_**" can't be edited and are disabled by default. If the "**_Client authentication_**" is "ON", "**_Authorization_**" and "**_Service account_**" roles can be edited. BTW, if the "**_Authorization_**" is enabled, "**_Service account roles_**" will be selected by default and can't be edited. Here is a prototype for your reference. [Client settings prototype](https://5lvn0v.axshare.com/#id=ol9a55&p=usability_testing&c=1)
- B.  "**_Standard flow_**" and "**_Direct access grants_**" are selected by default. The vertical distance between the two options should be larger. In the design, their vertical spacing is 16px. 

![3](https://user-images.githubusercontent.com/53931519/110443785-4520ff80-80f7-11eb-8082-d1a75e2e2a91.png)

4. General settings: The **_Description_** field is a text area component.    

![4](https://user-images.githubusercontent.com/53931519/110444501-0fc8e180-80f8-11eb-9a5c-ce7316870288.png)

5. Access settings: 
- As for the components of "**_Valid redirect URIs_**" and "**_Web origins_**", we have [a separate section](https://marvelapp.com/prototype/908c3c3/screen/73506931) to show the design.  

![5](https://user-images.githubusercontent.com/53931519/110444529-19eae000-80f8-11eb-8c31-b0c027779e8e.png)

6. Login settings: 
- A. "**_Login theme_**" is missing here. 
[...]

PR #429, opened 2021-03-10
fixed visual and logical errors described in #423

edewit:
fixing: #423


src/clients/ClientSettings.tsx
```diff
@@ -141,13 +234,18 @@ export const ClientSettings = ({ save }: ClientSettingsProps) => {
...
+              }
             />
           </FormGroup>
           <ActionGroup className="keycloak__form_actions">
             <Button variant="primary" onClick={save}>
               {t("common:save")}
             </Button>
-            <Button variant="link">{t("common:cancel")}</Button>
+            <Button variant="link" onClick={reload}>
+              {t("common:reload")}
```
mfrances17, line 247: this was changed to 'revert' in a recent bug fix

mfrances17 (changes requested):
Some quick changes:

- 'reload' is now 'revert' in common messages, as per recent bug fix
- he had asked for more spacing between the two columns of auth flow checkboxes, i think you missed it:
auth-flow-spacing

- the left margins for section heads are inconsistent between tabs - check out the difference between the settings tab and the advanced tab:
settings-left-margin

advanced-left-margin

jenny-s51 (changes requested):
Looking good @edewit ! 

A couple small things:

1) The login theme dropdown has a "blank" option that is clickable ... instead we probably want something like the client type dropdown where the default state has the "Select an option" text:
Screen Shot 2021-03-16 at 2 51 30 PM
2) In the "Advanced" tab the scroll form needs left padding. 

This can be fixed by updating form-panel.css as shown below:

```
.kc-form-panel__panel {
  padding-top: var(--pf-global--spacer--lg);
  padding-bottom: var(--pf-global--spacer--2xl);
  padding-left: var(--pf-global--spacer--lg); /* add this line */
  max-width: 768px;
}
```

edewit:
The left margins are fixed in the PR for the UXD review of the Advanced settings page

mfrances17 (approved):

jenny-s51 (approved):

Issue #417, opened 2021-03-05
[UX review] Make the search behavior consistent around the console


yih-wang:
1. The client role list filters automatically when the user typing. It's better to align this behavior to the other list - only apply the filter after the user clicks the search icon, or hit the Enter on the keyboard.
![image](https://user-images.githubusercontent.com/23331885/110091748-c0bb3d80-7dd3-11eb-9d0c-09e897d7092c.png)
2. All the searches should have applied filters display as labels/chips under the toolbar. Reference the PF4 pattern - https://www.patternfly.org/v4/components/toolbar#with-filters

edewit:
Search Groups already has some chips (e.g. `SearchGroups.tsx`)

yih-wang:
@edewit Do you think this issue can be closed?

edewit:
we don't have the attributes / values as a general thing yet, guess it's a bit hard to do that we might need to think about this a bit more

PR #433, opened 2021-03-11
Changes from Realm Roles UX Review [List]

jenny-s51:
## Motivation

#416 

## Brief Description

Towards #416 - Addresses all items in [List] category of UX Review

- Removes the margin of the Role list, reference the latest style of the Client list.
- Removes the margin of the Associated Role list
- Search box in the toolbar: search now works by hitting enter
- When searching for something that doesn't have a matched result, show the search empty state (https://github.com/keycloak/keycloak-admin-ui/pull/435)
- Adds chip group functionality to Toolbar+DataTable for filtering by category in other sections of Keycloak
- Fixes save button functionality on Role Details page

## Verification Steps

Verify that all the fixes listed above have been addressed 

## Checklist:


edewit (commented):
I'm confused about these `filterChips` how are they supposed to work? I thought it would be that when I search for one term and then search for something else after that these terms would end up beneath the search input?

yih-wang:
> I'm confused about these `filterChips` how are they supposed to work? I thought it would be that when I search for one term and then search for something else after that these terms would end up beneath the search input?

Each time the user clicks the search, the term will show as a chip beneath the toolbar. Reference this demo - https://www.patternfly.org/v4/components/table/react-demos/#filterable

mfrances17 (changes requested):
Aside from the filter chips, which we are handling as a separate issue, only one of the issues from the bulleted list of fixed items wasn't working in my testing. I am still getting the empty state when searching for a string that does not exist:

search

result

jenny-s51:
@mfrances17 thank you for your comment - I rebased and search empty state should now be addressed (https://github.com/keycloak/keycloak-admin-ui/pull/435). Please kindly re-review when you have a moment :)

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -170,32 +170,7 @@ export function KeycloakDataTable<T>({
...
-    setFilteredData(
-      rows!.filter((row) =>
-        row.cells.some(
-          (cell) =>
-            cell &&
-            getNodeText(cell).toLowerCase().includes(search.toLowerCase())
-        )
-      )
-    );
-  };
```
edewit, line 198: don't we still need this?

edewit (changes requested):
Filtering still needed right?

src/components/table-toolbar/KeycloakDataTable.tsx
jenny-s51: `filter` (which calles 'getNodeText`) is never used so eslint throws an error... this function was used to filter in real-time as the user types, right?

Per @yih-wang:

> It's better to keep consistency among all the search behaviors in a console. I looked into some other websites and think it's a common usage that return to the list by deleting the search input and search again. So let's keep using that. I update the clarification of the behaviors in this issue https://github.com/keycloak/keycloak-admin-ui/issues/417

jenny-s51:
@edewit filter (which calls `getNodeText`) is never used here so eslint throws an error... this function was used to filter in real-time as the user types

Per @yih-wang:

> It's better to keep consistency among all the search behaviors in a console. I looked into some other websites and think it's a common usage that return to the list by deleting the search input and search again. So let's keep using that. I update the clarification of the behaviors in this issue #417

mfrances17 (approved):
The issue I had flagged about empty state on not finding a role name search result is now fixed... LGTM. Nice job.

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: Yes it was used to filter in real time, but we only want to change it that we no longer filter in real time not that we don't filter at all. If you remove this searching on the client scope page will no longer work.

edewit (changes requested):
we still need filtering just not the while you type.

jenny-s51:
@edewit that makes sense - thank you for clarifying! Updated to use the filter function again

edewit (approved):

PR #435, opened 2021-03-12
added search empty state

edewit:


jenny-s51 (changes requested):
Empty state condition is getting hit when there are no groups, users, etc.

I am also able to successfully see "no search results found" if I search for a client or client scope that isn't in the list. 

However in users and realm roles, if I search for something that is not in the list, a blank white screen renders that is neither the empty state or the "no search results found" case

jenny-s51 (approved):

PR #437, opened 2021-03-12
Nonscrolling action buttons for long forms

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/398.

## Brief Description
On long forms, aka the forms that have jump links, the action buttons (Save, Cancel, Revert, etc) to the bottom of the screen should be fixed so you don't need to scroll long forms to click them.

## Verification Steps
1. Go to user fed and select an LDAP provider, or create one if it doesn't already exist.
2. Verify that the Save and Cancel buttons are available at the bottom of the screen without scrolling.
3. Verify that you can navigate the long form by scrolling or using the jump links, and the action buttons still remain available at the bottom of the screen.
4. Repeat steps 1-3 for the client details long form.

## Checklist:

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -8,3 +8,13 @@
 .keycloak__form {
   max-width: 1024px;
 }
+
+.keycloak__form_actions {
+  position: fixed;
+  bottom: 0px;
+  background-color: var(--pf-c-page__main-section--m-light--BackgroundColor);
+  padding-top: var(--pf-global--spacer--md);
+  padding-bottom: var(--pf-global--spacer--sm);
+  width: 100%;
+  border-top: 1px solid var(--pf-global--BorderColor--100);
+}
--- a/public/index.css
+++ b/public/index.css
@@ -15,6 +15,8 @@
   background-color: var(--pf-c-page__main-section--m-light--BackgroundColor);
   padding-top: var(--pf-global--spacer--md);
   padding-bottom: var(--pf-global--spacer--sm);
+  padding-left: var(--pf-global--spacer--md);
   width: 100%;
   border-top: 1px solid var(--pf-global--BorderColor--100);
+  margin-left: calc(-1 * var(--pf-global--spacer--lg));
 }
--- a/src/clients/ClientSettings.tsx
+++ b/src/clients/ClientSettings.tsx
@@ -18,8 +18,6 @@ import { MultiLineInput } from "../components/multi-line-input/MultiLineInput";
 import { FormAccess } from "../components/form-access/FormAccess";
 import { HelpItem } from "../components/help-enabler/HelpItem";
 
-import "../common-styles.css";
-
 type ClientSettingsProps = {
   save: () => void;
 };
@@ -145,7 +143,7 @@ export const ClientSettings = ({ save }: ClientSettingsProps) => {
               ref={register}
             />
           </FormGroup>
-          <ActionGroup className="nonscroll-action-group">
+          <ActionGroup className="keycloak__form_actions">
             <Button variant="primary" onClick={save}>
               {t("common:save")}
             </Button>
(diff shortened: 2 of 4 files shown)


jenny-s51 (approved):
Nonscrolling action buttons look great! 👍

Jump links are a bit buggy in a few places throughout the console (i.e. clicking on a link doesn't highlight the correct region in the jump links area) but we can definitely file another issue for that.

Nice update @mfrances17 !

src/clients/ClientSettings.tsx
```diff
@@ -18,6 +18,8 @@ import { MultiLineInput } from "../components/multi-line-input/MultiLineInput";
 import { FormAccess } from "../components/form-access/FormAccess";
 import { HelpItem } from "../components/help-enabler/HelpItem";
 
+import "../common-styles.css";
```
edewit, line 21: we already have an `index.css` for common styles

edewit:
I've introduced this already in #423

src/common-styles.css
```diff
@@ -0,0 +1,8 @@
+.nonscroll-action-group {
+    overflow: hidden;
```
srambach, line 2: I don't think you need overflow:hidden here (maybe, but it looks like it's being overwritten by the pf-m-action doing it anyway)

src/common-styles.css
```diff
@@ -0,0 +1,8 @@
...
+    background-color: white;
```
srambach, line 3: It would be better to use a PF variable here - I think `var(--pf-c-page__main-section--m-light)` is probably best.

src/common-styles.css
```diff
@@ -0,0 +1,8 @@
...
+    position: fixed;
+    bottom: 0;
+    width: 100%;
+    padding: 15px;
```
srambach: It's always best to use a global spacer variable. `var(pf-global--spacer--md` is 16px. But since this is being contained in the area that already has padding, I also think you might just want this on the top and bottom, not the left side, so that the buttons line up with the text.

src/common-styles.css
mfrances17: thanks - updated

src/common-styles.css
mfrances17: agreed, thanks... updated

src/common-styles.css
mfrances17: this is removed in the newest version - and styles are now in index.css, did not need to create a common css file as erik flagged in his review

src/clients/ClientSettings.tsx
mfrances17: thanks - i've changed it to use this css, and used your changes from the pending client ux review fixes. the only thing i changed was the bottom padding - it was clipping the bottom-most fields on full scroll in both clients and user fed.

edewit (approved):
Nice :+1:

public/index.css
```diff
@@ -7,4 +7,14 @@
...
+}
+
+.keycloak__form_actions {
+  position: fixed;
+  bottom: 0px;
+  background-color: var(--pf-c-page__main-section--m-light--BackgroundColor);
+  padding-top: var(--pf-global--spacer--md);
+  padding-bottom: var(--pf-global--spacer--sm);
+  width: 100%;
+  border-top: 1px solid var(--pf-global--BorderColor--100);
```
srambach, line 19: In order to get the border to show all the way across, you'll need to give it a negative margin that matches the spacer of the container that's containing it, and then add back in enough padding to align it.
Also, there's a border global variable you can use. I think the wizard design is using a border shadow so might want to check with the designers.

srambach: Also, width 100% is actually making it wider than it should be but I don't have a quick fix without looking more. This would be an issue if you ever right aligned the buttons or if they got too wide and ran off the screen.

mfrances17: i've added a negative margin so the border goes across the entire bottom, with no gap on the left

PR #442, opened 2021-03-15
partially solves issue #422

edewit:


src/clients/AdvancedTab.tsx
```diff
@@ -178,96 +176,95 @@ export const AdvancedTab = ({
...
-              labelIcon={
-                <HelpItem
-                  helpText="clients-help:notBefore"
-                  forLabel={t("notBefore")}
-                  forID="kc-not-before"
-                />
-              }
+      <>
+        <Text className="pf-u-py-lg">
+          <Trans i18nKey="clients-help:notBeforeIntro">
```
mfrances17, line 181: should use the "t" function here

mfrances17 (approved):
LGTM as a partial solve for the issues in 422

src/clients/AdvancedTab.tsx
edewit: I can't use the `t` function as I need a component in the translated text

mfrances17: ah, ok

PR #447, opened 2021-03-16
changed time select order + added success message

edewit:
fixes: #445

### DIFF
--- a/src/clients/initial-access/CreateInitialAccessToken.tsx
+++ b/src/clients/initial-access/CreateInitialAccessToken.tsx
@@ -51,6 +51,7 @@ export const CreateInitialAccessToken = () => {
           token={token}
           toggleDialog={() => {
             setToken("");
+            addAlert(t("tokenSaveSuccess"), AlertVariant.success);
             history.push(`/${realm}/clients/initialAccessToken`);
           }}
         />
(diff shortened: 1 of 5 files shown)


edewit:
now wraps the column title:
image

src/clients/initial-access/CreateInitialAccessToken.tsx
```diff
@@ -78,13 +79,14 @@ export const CreateInitialAccessToken = () => {
           >
             <Controller
               name="expiration"
-              defaultValue=""
+              defaultValue={86400}
```
mfrances17, line 82: expiration needs a min val of 1, as per the UX review... i was able to select blank and 0, and type a negative value

mfrances17 (changes requested):
Code looks good! One small issue, the rest seems to be working well to address the issues in 445

src/clients/initial-access/CreateInitialAccessToken.tsx
edewit: This is behavior we need as per old console

mfrances17: ok

mfrances17 (approved):

jenny-s51 (approved):

PR #448, opened 2021-03-17
Added rename group and adding members to a group

edewit:
## Motivation

This PR adds a way to add member to a group and also renaming

## Verification Steps

Go to a group detail page and click on the member tab then use the add members button.
Select a user to add to the group and click add.
Now the selected user are members of the selected group

From the actions menu on the detail page select rename group to change the group name

## Checklist:


mfrances17 (changes requested):
Code looks good. 

When testing add member(s) to group, was able to add combinations of 1, 2, and 3 users to the group, and remove them from the group in various combinations as well... no errors or behavioral issues.

When testing Rename group, I get uncaught type errors and the rename does not proceed. Note that this occurs when I have a specific group selected as well as when nothing is selected:

```
GroupsSection.js:120 Uncaught TypeError: Cannot read property 'name' of undefined
    at onClick (GroupsSection.js:120)
    at onClick (Popover-7fe436e5.js:3832)
    at HTMLUnknownElement.callCallback (index-e900dd09.js:1383)
    at Object.invokeGuardedCallbackDev (index-e900dd09.js:1432)
    at invokeGuardedCallback (index-e900dd09.js:1487)
    at invokeGuardedCallbackAndCatchFirstError (index-e900dd09.js:1501)
    at executeDispatch (index-e900dd09.js:1584)
    at executeDispatchesInOrder (index-e900dd09.js:1609)
    at executeDispatchesAndRelease (index-e900dd09.js:4473)
    at executeDispatchesAndReleaseTopLevel (index-e900dd09.js:4482)
    ```

**Comment by edewit** [comment]
@mfrances17 right removed the actions when no group is selected

**Review by mfrances17 — CHANGES_REQUESTED** [review]
The rename works when you are on a specific group's detail page, but not on the groups page. Is that intentional? I would expect to select a specific group and rename it from this screen (if only one were checked, otherwise the commands should be unselectable). 

no-rename

Also, menu commands should never disappear entirely in the wrong context, they should still appear but be disabled.

File: cypress/integration/group_test.spec.ts [hunk 1]
```diff
@@ -92,6 +92,17 @@ describe("Group test", () => {
...
 
+    it("Should add members", () => {
+      listingPage.goToItemDetails(groups[0]);
+      detailPage
+        .clickMembersTab()
+        .clickAddMembers()
+        .checkSelectableMembers(["user1", "user4"]);
+      detailPage.selectUsers(["user1"]).clickAdd();
+
+      masthead.checkNotificationMessage("1 user added to the group");
```
ssilvert, line 103: I haven't run this code yet because of the conflict, but it looks like the test only checks the notification.  It should also check that the user was added to the page, right?

ssilvert (changes requested):
I don't see a test for rename group.

ssilvert (changes requested):
When you add users to the group, you can only do it a page at a time.  If I select users on the first page and then go to a second page, all my users on the first page get unselected.

ssilvert (changes requested):
Found another issue:
1. Create a new realm
2. Go to old console and import 100 users.  (I'll email you the json) 
3. Got to Groups -> MyGroup -> Members
4. Click on Add Members
5. Add the 10 users on the first page
6. Go back to Add Members dialog again.
7. You will see "No users found"

Also, there was some odd behavior when playing around with the number of users to show per page on the "Add member" dialog.  Guessing it's probably related to the above issue.

As per our discussion, we need to include a cancel button on modals.  Checked on PatternFly, and it seems to be standard behavior: https://www.patternfly.org/v4/components/modal

ssilvert (approved):

PR #450, opened 2021-03-17
Users: Add groups tab and list groups 

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-864

## Brief Description

Add groups tab to users section and list joined groups in data table.

## Checklist:


edewit (approved):
:+1:

mfrances17 (changes requested):
i don't know what this message means when attempting to remove a group from the groups tab:
what

Also, you can execute a remove regardless of whether the group is selected, and in either case, it fails with a 404.

Add role is not working, search is not working... perhaps this PR only covers the listing of applicable groups for a specific user?

jenny-s51:
Thank you for your helpful comments @mfrances17 ! I've added the "Leave" functionality to this PR based on your feedback . And yes that is correct - the Add Role modal will be addressed in a separate PR and the search functionality will work once https://github.com/keycloak/keycloak-admin-ui/pull/433/files is merged.

ssilvert (changes requested):
The first time you go to the groups page, the list will be empty.  Nothing shows up until I check the "Direct membership" box.  Then the groups show up whether checked or not.

(BTW, we need help text for that "Direct membership" checkbox.  It's unclear what that is supposed to do.  I'll create an issue for that.)

ssilvert (changes requested):
Also, I don't see any new tests for the new functionality.

jenny-s51:
@ssilvert thank you for your feedback! I've updated this PR to list all the groups on the initial render. 

I'm currently working on a separate PR which will add the "Join Groups" functionality and will include robust cypress tests there, since a user will then be able to add groups from the new console rather than the old one.

This open PR here is meant to address the implementation of the groups tab and renders a data table if groups are added in the old console. What kind of tests would you like to see in this PR?

PR #452, opened 2021-03-19
fixed group navigation

edewit:


mfrances17 (changes requested):
This has no description, and I don't see a related issue. Barring steps to reproduce, can you at least explain what the issue was and how it was fixed in the description?

ssilvert (approved):
@mfrances17 @jenny-s51 Subgroups were broken when I switched us to hash routing.  This is Ed's fix.

PR #456, opened 2021-03-20
add Docker and docker-compose

nicolabeghin:
## Motivation
Adding `Dockerfile` and `docker-compose.yml` for ready-to-use local build for all the not-frontend-proficient folks like me 😄 

## Verification Steps
    docker-compose build
    docker-compose up -d 

Access http://localhost:8080/adminv2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,34 @@
-FROM node:12
+ARG DEFAULT_KEYCLOAK_ENDPOINT='http:\/\/localhost:8180'
 
+FROM node:12 as builder
+
+ARG DEFAULT_KEYCLOAK_ENDPOINT
+ARG KEYCLOAK_ENDPOINT
+
+# install updated yarn
 RUN rm -fr /usr/local/bin/yarn* &&\
     npm install -g yarn
 
 WORKDIR /app
 COPY . .
 
-RUN yarn
+# replace Keycloak endpoints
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN rm -fr /usr/local/bin/yarn* &&\
+    npm install -g yarn
+
+WORKDIR /app
+COPY . .
+
+RUN yarn
+
+CMD [ "yarn", "start" ]
```
edewit, line 11: Wouldn't it be nicer to do a `yarn build` and then use a simple http server to serve the static files? That way the site would be a lot faster as this is developer mode, or is that what you wanted?

nicolabeghin: @edewit totally agree, but tried to stick to "barebone" setup as per README for dev/sneak-peek mode (no PRD optimization at all).

edewit:
@nicolabeghin would be good to add these instructions to build and run in the README

Dockerfile
edewit, line 11: how about this:
```suggestion
RUN yarn build

FROM nginx
EXPOSE 80
COPY --from=builder /app/build /usr/share/nginx/html
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:12
```
edewit, line 1:
```suggestion
FROM node:12 as builder
```

nicolabeghin:
@edewit updated with build and nginx, plus some basic performance optimizations

edewit (approved):

PR #467, opened 2021-03-26
Users(groups): list top level groups, implement direct membership check

jenny-s51:
## Motivation

#466 

## Verification Steps

1. Go to the old admin console and create a group and another group with subgroups.
2. In the old admin console, go to the "Users" section, click on the "Groups" tab and add those groups to an existing user.
3. In the new admin console, go to the "Users" section and find the existing user.
4. Click on the User and go to the "Groups" tab.
5. Verify the joined groups show up in the table, and that the top-level groups are also visible.
6. Click on the "Direct membership" check and verify that the top-level groups disappear from the user view.

## Checklist:


edewit (approved):
:+1:

ssilvert (changes requested):
If I have group A/B/C and then I join group C, only A and C will be shown in the list.

ssilvert:
@jenny-s51 It is now showing indirect membership but it is not as per spec.  When you show the indirect group, it should have its checkbox disabled and the "Leave" link should not appear.

ssilvert (approved):

PR #472, opened 2021-03-29
adds fake pagination to datatable

edewit:
even when the backend doesn't support pagination we still show paginated results


jenny-s51 (commented):
@edewit pagination is looking good - I tested your changes on a few different components and have a quick question: in what cases would we need to use the `isPaginated` prop?

For example, in `AssociatedRolesTab.tsx` pagination works without this prop and breaks if `isPaginated` is set. But in `RolesList` the `isPaginated` must be set in order for pagination to work correctly. When would we want to set the `isPaginated` prop on a component?

edewit:
you use the `isPaginated` when the backend supports paginated results. Then for each new page the loader is called to get a page from the backend. When the backend doesn't support this or when you have aggregated data the data will be fetched in one go and then the component will make it look like there are pages

jenny-s51 (approved):

PR #479, opened 2021-03-30
fixed issues described in #468

edewit:
fixing: #468


xianli123:
Thanks. LGTM. But there are two parts that need to be updated. 

List review R2

A. If users select more than one client scopes, this checkbox should be mixed selected. 
B. We can adjust the spacing of the attributes to display the entire name, Display order. And we can put the "Display order" after the "Protocol".

edewit:
we use the `Table` component from patternfly for and it has an option to have select-able rows. The "select all" checkbox doesn't support the mixed selected state. Even if we would build our own html doesn't support a checkbox like this one would need to create it from the ground up.

ssilvert:
@xianli123 Please review again.

xianli123 (commented):
It has been reviewed and there are no new suggestions. Thanks.

xianli123 (approved):

PR #482, opened 2021-03-30
Allow four ways to set location of keycloak server.

ssilvert:
## Motivation
Need to be able to run new admin console from keycloak.org.

## Brief Description
Four ways to set location of Keycloak server:

1. http://anydomain/anycontext/?keyclaok-server=http://my-keycloak-server.com
2. Usual dev mode, which defaults server to http://localhost:8180
3. Demo mode, http://anydomain/adminv2/?demo=true, which defaults server to http://localhost:8080
4. Integrated into keycloak server, http://keycloak-server-domain/adminv2

## Verification Steps
Try out any or all of the modes above.

## Checklist:

### DIFF
--- a/src/context/auth/keycloak.ts
+++ b/src/context/auth/keycloak.ts
@@ -6,23 +6,43 @@ export default async function (): Promise<KcAdminClient> {
 
   const kcAdminClient = new KcAdminClient();
 
-  const authContext = "/auth";
-  const keycloakAuthUrl = window.location.origin + authContext;
-  const devMode = !window.location.pathname.startsWith("/adminv2");
   try {
     await kcAdminClient.init(
       { onLoad: "check-sso", pkceMethod: "S256" },
       {
-        url: devMode ? "http://localhost:8180/auth" : keycloakAuthUrl,
+        url: keycloakAuthUrl(),
         realm: realm,
         clientId: "security-admin-console-v2",
       }
     );
     kcAdminClient.setConfig({ realmName: realm });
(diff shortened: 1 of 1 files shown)


src/context/auth/keycloak.ts
```diff
@@ -6,23 +6,43 @@ export default async function (): Promise<KcAdminClient> {
...
+
+  const searchParams = new URLSearchParams(window.location.search);
+
+  // passed in as query param
+  const authUrlFromParam = searchParams.get("keycloak-server");
+  if (authUrlFromParam) return authUrlFromParam + authContext;
+
+  // dev mode
+  if (!window.location.pathname.startsWith("/adminv2"))
+    return "http://localhost:8180" + authContext;
```
edewit, line 41: Shouldn't dev mode be only `authContext` until the CORS issue is solved?

edewit (changes requested):

src/context/auth/keycloak.ts
ssilvert: @edewit The CORS issue is solved and merged yesterday: https://github.com/keycloak/keycloak/pull/7891

Did this not work for you?  I'll change it to plain authContext if it's not working with older keycloak servers.

ssilvert: @edewit You were right of course.  With a different port it's considered a different domain.

I changed it so it will continue to work in dev.  We can change it back once everyone is able to upgrade to Keycloak 13.

edewit (approved):
:+1:

PR #485, opened 2021-03-31
can't move groups to the root level

edewit:
disabled "Move" button on root level

fixing: #476

### DIFF
--- a/cypress/integration/group_test.spec.ts
+++ b/cypress/integration/group_test.spec.ts
@@ -94,6 +94,21 @@ describe("Group test", () => {
       sidebarPage.goToGroups();
       listingPage.deleteItem(targetGroupName);
     });
+
+    it("Should move group to root", async () => {
+      const groups = ["group1", "group2"];
+      await new AdminClient().createSubGroups(groups);
+      sidebarPage.goToGroups();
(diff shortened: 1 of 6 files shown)


ssilvert (changes requested):
Looks like this does fix an issue, but it's not what I meant.  I should have been more clear.

Say I have these groups:
/A
/A/B

I should be able to move B to the root so then I would have:
/A
/B

Also, I noticed that when you delete a group, the screen is not refreshed.

edewit:
Right that is why I disabled the `Move` button as you can't move group `B` to `/` as there is no API to do that you can only move group `B` to be a subgroup of another group

ssilvert:
> Right that is why I disabled the `Move` button as you can't move group `B` to `/` as there is no API to do that you can only move group `B` to be a subgroup of another group

I can do it in the old console.  If I have A/B I just click on B and say "cut".  Then I highlight "Groups", which is the root.  Then click on "paste".  Looks like it uses a different API when you want to move it to the root.

ssilvert (changes requested):
This looks good.  Just need a test case for it.

ssilvert (approved):

Issue #462, opened 2021-03-24
Implement functionality to disable kebabs/checkboxes for data list items


jenny-s51:
see #451 

> 3. Only the directly assigned roles can be removed, which means the checkboxes and the kababs of the inherited roles should be disabled. (Sorry I didn't include the annotation in the design. Just add it - https://marvelapp.com/prototype/f39d464/screen/72684631)

_Originally posted by @yih-wang in https://github.com/keycloak/keycloak-admin-ui/issues/451#issuecomment-804723743_

PR #486, opened 2021-03-31
added `isRowDisabled` function to disble rows

edewit:
fixing: #462


mfrances17 (changes requested):
I was able to pass a value of true to isRowDisabled prop of the associated roles data table, which did disable the checkboxes. However, the kababs were still selectable and I could click Remove. The design mentioned disabling the checkboxes and the kababs.

edewit:
Right, forgot about that, good call... will update that

mfrances17 (approved):
Yep, looks good - working as expected.

PR #490, opened 2021-03-31
User fed mappers layout and partial logic

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-541
https://issues.redhat.com/browse/APPDUX-707

## Brief Description
Adds mapper layout for new user fed mappers tab and forms for the 11 different mapper types, and fetches values for pre-existing mappers.

## Verification Steps
1. Create an LDAP provider if one doesn't already exist.
2. In the old admin console, add a mapper of each type for the LDAP provider.
3. In the new console, verify that the mappers are listed in the new Mappers tab of the LDAP provider.
4. Click each mapper and verify that the appropriate mapper information appears in its details, when comparing to the old console.

Note: This is part 1 of the user fed mapper implementation, was originally just for the layout, but i've added logic for fetching all the values for pre-existing mappers. Creating new, editing, saving, deleting, etc and other CRUD logic, as well as cypress tests, are all coming in part 2 scheduled for next sprint.

## Checklist:


edewit:
missing keys on empty mappers:
image

edewit (changes requested):
Other then that it looks good :+1:

mfrances17:
Thanks, good catch @edewit. It was on my list of strings to move to common and I must've missed it. Should be all set now.

jenny-s51 (changes requested):
Forms for the mapper types look great and match the forms in the old console! 

I noticed some of the cancel buttons on the mapper forms have different behaviors (e.g. some go back to the user fed card view, and some go back to the `settings` tab of the provider itself. What do you think should be the expected behavior @mfrances17 ? I think navigating them back to the `mappers` tab is a good idea.

Also I noticed a bug (seemingly related to pagination) which can be addressed in another issue. 

Steps to repro:
1. Make sure there are more than 10 mappers for the LDAP provider
2. Go to mappers list for LDAP provider in user fed
3. Either paginate to the next page, or set the max number of listed items to be more than 10
4. Click on any mapper and get "page not found"

Overall data fetching looks good 👍

mfrances17:
Thanks @jenny-s51, good finds. Both of those issues should be fixed in the latest push. The cancel button now returns the user to the mappers page, and the url-related issues aka 'page not found' errors should now be fixed (there were other ways to hit the error besides pagination, such as just navigating back using the browser back button instead of the breadcrumb). I was going to wait and fix these with the upcoming save/update logic but might as well address them now. Please take another look, thanks.

jenny-s51 (approved):
Looks great @mfrances17 ! 👍

Issue #396, opened 2021-02-23
[UX review] Realms list


yih-wang:
When there are more than 10 realms, it is supposed to show the 3 recently visited realms. I didn't see it in the current implementation. Is it postponed on purpose?
![image](https://user-images.githubusercontent.com/23331885/108847078-3710b080-761a-11eb-8331-f38ca288da3a.png)
@edewit

yih-wang:
The current realm will also be shown as the Recent realm. I think it's better not to be. The users won't want to click the current realm in the list so there isn't a need to put it at the top. 
![image](https://user-images.githubusercontent.com/23331885/119324593-97af9800-bcb2-11eb-8269-cceeae06d193.png)

By the way, can the dropdown be a bit higher? I'm expecting to see at least 5 realms when there are more than 5.

PR #502, opened 2021-04-06
added recently used realms to the top

edewit:
fixing: #396

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,6 +23,7 @@ import { SubGroups } from "./groups/SubGroupsContext";
 import { useRealm } from "./context/realm-context/RealmContext";
 import { useAdminClient, asyncStateFetch } from "./context/auth/AdminClient";
 import { ErrorRenderer } from "./components/error/ErrorRenderer";
+import { RecentUsed } from "./components/realm-selector/recent-used";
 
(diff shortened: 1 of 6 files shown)


ssilvert:
@edewit Not sure if this is related or not, but if I have six realms the "Create Realm" button stops working.

ssilvert:
> @edewit Not sure if this is related or not, but if I have six realms the "Create Realm" button stops working.

Looks like it is related to this PR.  This is working in master.

ssilvert (changes requested):
Small bug.  If you create a realm then you immediately go to that realm.  However, it is not considered a "Recent" realm.

So say I have Realms "A", "B", "C", ... "Y".  Then I create realm "Z".  Then I go to realm "A".  Even though realm "Z" is a most-recently used realm it will not show up in the list at all unless I search for it.

ssilvert:
> Small bug. If you create a realm then you immediately go to that realm. However, it is not considered a "Recent" realm.
> 
> So say I have Realms "A", "B", "C", ... "Y". Then I create realm "Z". Then I go to realm "A". Even though realm "Z" is a most-recently used realm it will not show up in the list at all unless I search for it.

Looks like this is just because the list has not been refreshed.  Needs to be fixed though.

ssilvert (approved):

Issue #317, opened 2021-01-21
Trying to save a field too long brings up a 500 error


Aboullos:
Steps
- Go to "Client Scope"
- Go to "Create client scope" or go to an item details
- Fill any text input with a long text
- Click "Save"

Expected result:
 - The error is handled and the notification to the users shows a message warning about the length of the text
 - Http response message warns about the error

Actual result:
 - Notification message unclear to user: "Could not update client scope: 'Error: Request failed with status code 500'"
 - Http response message is: {"error":"unknown_error"}

ssilvert:
This happens on lots of our forms.  We need to go through the whole app and do basic validation for lengths and numerics.

ssilvert:
Out of date

PR #506, opened 2021-04-06
added max length check

edewit:
fixing: #317

### DIFF
--- a/src/client-scopes/form/ClientScopeForm.tsx
+++ b/src/client-scopes/form/ClientScopeForm.tsx
@@ -155,9 +155,22 @@ export const ClientScopeForm = () => {
                   />
                 }
                 fieldId="kc-description"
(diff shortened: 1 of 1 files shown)


jenny-s51 (commented):
Validation on description looks good! However if the name of the client scope is too long I get an error saving. 

I also noticed the help text for `Display order` specifies the input value should be an integer, so it would be helpful to also validate this input type as a number.

What do you think @edewit ?

ssilvert:
I have the same feedback as @jenny-s51 on this one.  It's only fixed for the description field.

But we have this issue on other forms as well.   So somebody needs to go through all our forms and do basic validation of lengths and numerics.  I'm going to go ahead and merge this but keep the issue open and broaden its scope.

ssilvert (approved):

PR #501, opened 2021-04-06
disable when nothing selected and fix title

edewit:
fixing: #495

fixing: #493

PR #508, opened 2021-04-07
disable when nothing selected

edewit:
fixing: #495

fixing: #493


edewit:
instead of #501 to enable the CI checks

ssilvert (changes requested):
Now if I go to assign a service account role for a client called "foo", the header will say "Assign roles to account".  It should say "Assign roles to foo service account".

The Assign button is disabled until I check a box, but then if I uncheck the box it stays enabled.  

Lastly, you might as well go ahead and make the Cancel button be the link style since this is what we decided to do.

jenny-s51 (approved):
Looks good 👍

ssilvert (approved):

PR #512, opened 2021-04-08
Query realm list only once

edewit:
The realm list is queried multiple times this is an expensive request, this PR changes it so it's only called once.


ssilvert:
@edewit Please fix conflicts.

PR #513, opened 2021-04-08
Users(groups): Add modal for joining user groups

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-869

## Brief Description

Add functionality to join groups 

## Verification Steps

1. Go to Users section
2. Click on groups tab
3. Click "join group" button
4. Select a group and verify the new group shows up in the data list

## Checklist:


ssilvert:
@jenny-s51 The design allows you to select multiple groups at once using checkboxes.  Also, any group already joined should have a grayed out checkbox.

See https://marvelapp.com/prototype/f13gid4/screen/72210145

mfrances17 (changes requested):
The functionality to add a user to a group worked great. One thing I tried that should probably be covered is that I was able to add a user to the same group multiple times. I didn't get any errors when doing so, so that's good, but it should probably be disallowed to join a group that you are already a part of (either disable it or remove it from the Join groups modal.

jenny-s51:
@ssilvert following up on this per the group discussion during our team meeting this morning - I opened a separate issue to address functionality for adding multiple groups (#525), which will include the workaround we agreed to use (using a separate API call for each selected group, since the API currently doesn't support using just one to add multiple groups).

jenny-s51:
@mfrances17 thank you for your feedback! I've removed the existing joined groups from the modal for now so the user is not able to join those groups

mfrances17 (approved):
Looks good, verified that groups that have already been joined are no longer shown in the available groups to join list.

ssilvert (approved):

PR #514, opened 2021-04-09
Add scope tab to client scope detail page

edewit:
## Motivation
see design [marvel](https://marvelapp.com/prototype/5cf7cfd)

## Verification Steps

go to a client scope and assign roles

## Checklist:


ssilvert:
@edewit Please fix conflict.

mfrances17 (changes requested):
Code looks good! The ability to add roles worked well... one thing i noticed, some of the text seems off. 

1) When you unassign a role, the confirmation title is Remove mapping? and asks 'Are you sure you want to remove this mappings', if we want to keep 'mapping' it should be changed to 'Are you sure you want to remove this mapping?' (singular with question mark). I would think we'd want to change this to 'role' vs 'mapping' but up to you. 

2) Also, the button says Delete, it should probably be 'Remove' since you aren't deleting it.

3) The confirmation for the removal says 'The client has been deleted', that should be replaced with something like 'Role has been removed' or 'Role has been unassigned' etc

mfrances17 (commented):
I've run into an issue in which you can select the Hide inherited roles checkbox and it will toggle the inherited roles OR the assigned roles, but never both. The default list view should be both inherited and noninherited roles, i would think.

Another issue I noticed is that roles were automatically added by doing the above, that were never chosen from the list. I selected 'create-realm' and added it, but it is not listed in the Scope list, but the ones not chosen, 'uma_authorization', 'offline_access', 'admin', and 'myrole' appear.

edewit:
> I've run into an issue in which you can select the Hide inherited roles checkbox and it will toggle the inherited roles OR the assigned roles, but never both. The default list view should be both inherited and noninherited roles, i would think.

@mfrances17 I think the inherited roles always include the assigned roles so putting them both there would create duplicates?

Issue #428, opened 2021-03-10
[UX review] Client - Credentials review


xianli123:
Hi, @edewit I've closed the issue #424 and opened a new issue here for more details of the Credentials review. 

Please refer to the Marvel link related to the [Credentials design](https://marvelapp.com/prototype/695h94c/screen/72265686). 

1. Save strategy: 
- Whenever users change the **_Authenticator_**, the change will not take effect until it is saved. The **Save** button is clickable only if the authenticator or its related field changes, otherwise the **Save** button is disabled. After clicking the **Save** button, there will be a confirmation modal popup to remind users the authenticator will be changed.
  <img width="645" alt="Screen Shot 2021-03-10 at 4 18 51 PM" src="https://user-images.githubusercontent.com/53931519/110598193-70215700-81bc-11eb-9694-fa349d4c3b27.png">

2. Client ID and secret: 
- The authenticator is "Client ID and secret" by default. The user can click the **Regenerate** button to generate a new secret. 
  <img width="795" alt="Screen Shot 2021-03-10 at 4 23 49 PM" src="https://user-images.githubusercontent.com/53931519/110598856-3270fe00-81bd-11eb-9ce6-417c30051e6a.png">

- After clicking the **Regenerate** button, there will be a modal popup. The same thing happens to **Registration access token.**
  <img width="708" alt="Screen Shot 2021-03-10 at 4 25 27 PM" src="https://user-images.githubusercontent.com/53931519/110598960-57657100-81bd-11eb-94ae-f7cc4b663de2.png">

3. Signed JWT with client secret: 
- If the user changes the **_Authenticator_** to "**Signed JWT with client secret**", the **Save** button will be enabled, and the **Regenerate** button will be disabled. After clicking the **Save** button, there will be a confirmation modal popup.
  <img width="780" alt="Screen Shot 2021-03-10 at 4 30 10 PM" src="https://user-images.githubusercontent.com/53931519/110599598-00ac6700-81be-11eb-9b04-c7afa04b1354.png">

4. X509 certificate: 
- If the user changes the **_Authenticator_** to "**X509 certificate**", the user has to fill in the **_Subject DN_**, otherwise, the **Save** button is not clickable. After filling in the **_Subject DN_** and clicking the **Save** button, there will be a confirmation modal popup.
[...]

PR #522, opened 2021-04-13
fixing comment on issue:

edewit:
https://github.com/keycloak/keycloak-admin-ui/issues/428#issuecomment-818504696

### DIFF
--- a/src/clients/AdvancedTab.tsx
+++ b/src/clients/AdvancedTab.tsx
@@ -177,7 +177,7 @@ export const AdvancedTab = ({
   }
 
   return (
-    <PageSection variant="light">
+    <PageSection variant="light" className="pf-u-py-0">
       <ScrollForm sections={sections}>
         <>
           <Text className="pf-u-py-lg">
--- a/src/clients/ClientSettings.tsx
+++ b/src/clients/ClientSettings.tsx
@@ -40,7 +40,7 @@ export const ClientSettings = ({ save, reset }: ClientSettingsProps) => {
   return (
     <>
       <ScrollForm
-        className="pf-u-p-lg"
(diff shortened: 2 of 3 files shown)


mfrances17 (changes requested):
i'm not sure this meets what the issue had requested... it looks like there is a gap larger (height) than 24px between the tabs and the client authenticator section. also, noticed that the separator above client secret is not going all the way across anymore, probably not intended:

credentials-tab

edewit:
@mfrances17 with this padding on the top it's the same as the settings tab. If we make the padding on the top different on this page then others need to change too? @xianli123 ?

Any ideas on how to make the separator go the entire width?

xianli123:
hi @edewit @mfrances17 I've added some measurements, as the picture shows below. 

Client ID   Secret - 3 Copy

The padding on the top should be 24px. The same thing happens to the Settings page.
It’s better to let the separator go the entire width. WDYT?

xianli123:
This gap is still a bit wide, and it's not the same height as the Settings page. Can you fine-tune it? Thanks.
Screen Shot 2021-04-20 at 9 03 36 PM

edewit:
@xianli123 it's the same as on the side right, that was the idea e.g. x == y = 24px so I don't know what else to make of it
image

ssilvert (approved):
I'm going to go ahead and merge.  We can fine tune this later when we have more time.

xianli123:
> @xianli123 it's the same as on the side right, that was the idea e.g. x == y = 24px so I don't know what else to make of it
> image

OK, I may make some mistakes. I will review the new UI, and if there are some updates, I will let you know.

Issue #521, opened 2021-04-13
Realm client details crashes app


ssilvert:
If you try to go to one of the realm clients such as "master-realm", you will get `Cannot read property 'charAt' of undefined` and several errors in the console.

PR #523, opened 2021-04-13
protocol can be not set

edewit:
fixing: #521


mfrances17 (approved):
LGTM, not seeing the same charAt issue with this fix

Issue #509, opened 2021-04-07
[UX review] Events - User events


xianli123:
Hi @edewit, there are some fine-tuning of the **Users events**, please reference the pictures below.

> ![User events](https://user-images.githubusercontent.com/53931519/113823903-bf55aa00-97b1-11eb-9f4b-8895477c7b65.png)

- A. Header: "Event configs" is an anchor. Click it will enter the Event configs tab inside Realm settings. 
- B. We can display the time rather than the period of time. The same thing goes with **_Admin events_**. The list should be collapsible as the picture shows.
> ![User events #2 - 2](https://user-images.githubusercontent.com/53931519/113824854-d779f900-97b2-11eb-98bb-d452e04bb00c.png)

- C. **User ID** is an anchor. After clicking it, users can view the user's details on the Users detail page.
- D. We can use the different labels to distinguish the Error type and the regular type. Hover on the Error label, the error report will pop up.
> ![User events #2 - 1](https://user-images.githubusercontent.com/53931519/113825100-26279300-97b3-11eb-8b35-cab2e4821102.png)

edewit:
@xianli123 I could make the collapsible rows, but I don't have the information you want to put there

edewit:
same goes for the popup on the status, I don't have any extra information why the error occurred

xianli123:
In the collapsible list, if you want to get the displayed info, please refer to the Details part inside the old console as shown in the picture below.

> ![UserEvents](https://user-images.githubusercontent.com/53931519/114671638-e161a600-9d36-11eb-9cab-be6ddddfb09f.png)

As for the Error report, there is a piece of error information in the details part.

> ![Error](https://user-images.githubusercontent.com/53931519/114671824-15d56200-9d37-11eb-9ccd-0580d0297853.png)

PR #529, opened 2021-04-14
fixing UXD review issue

edewit:
fixing: #509


jenny-s51 (commented):
User details tab looks good and matches the info in the old console. For example when I log in and log out I see the data listed under user events. 

Under event type I'm seeing the general event type with the green checkmark only. How would you suggest testing the error case?

edewit:
@jenny-s51 you get an error by trying to log in with a invalid password

jenny-s51 (approved):
Thank you @edewit - LGTM

PR #535, opened 2021-04-15
Add/edit/delete LDAP mappers

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-900

## Brief Description
Add ability to Create, Update, and Delete LDAP mappers.

Notes:
- You can't create either of the role-based mappers yet.
- Several of the group fields are not updating properly (shared form with roles).
- No cypress tests yet.

These will be addressed in upcoming PRs.
 
## Verification Steps
1. Select an existing User federation > LDAP provider.
2. Go to the mappers tab.
3. Verify that you can delete an existing mapper.
4. Verify that you can add a new mapper (one of the 9 non-role-based mappers) and that the mapper is populated in the mapper list, and that the values you chose are saved.
5. Verify that you can edit an existing mapper and your changes are saved when you navigate away and back.

## Checklist:


src/route-config.ts
```diff
@@ -220,21 +220,21 @@ export const routes: RoutesFn = (t: TFunction) => [
     access: "view-realm",
   },
   {
-    path: "/:realm/user-federation/ldap/:id/:tab?",
+    path: "/:realm/user-federation/ldap/new",
```
edewit, line 223: This broke your ldap create as new is no longer a parameter for the router

edewit:
The Mapping tab need to not be there on the new page I think. And the mapping table is missing the type column

src/route-config.ts
mfrances17: > The Mapping tab need to not be there on the new page I think. And the mapping table is missing the type column

I've fixed both of these issues... please re-review. Also, thanks for the commits to help fix the refresh.

edewit (approved):

src/user-federation/ldap/mappers/LdapMapperDetails.tsx
```diff
@@ -0,0 +1,394 @@
...
+                    required
+                    onToggle={() =>
+                      setIsMapperDropdownOpen(!isMapperDropdownOpen)
+                    }
+                    isOpen={isMapperDropdownOpen}
+                    onSelect={(_, value) => {
+                      onChange(value as string);
+                      setIsMapperDropdownOpen(false);
+                    }}
+                    selections={value}
```
jenny-s51, line 235:
```suggestion
                    placeholderText="Select a mapper type"
                    selections={value}
```

jenny-s51 (changes requested):

jenny-s51 (changes requested):
Add/delete functionality works great @mfrances17 ! 

I created a group-ldap-mapper type. When I edited the form and then returned to the mapper, several fields did not get updated with the new values, such as the "Ignore missing groups" switch for example.

Where is the `config` you are reading from to set the `name` attribute? I searched the repo but I the search didn't return any results with these values (i.e. `"config.always-read-value-from-ldap"`).

mfrances17:
> I created a group-ldap-mapper type. When I edited the form and then returned to the mapper, several fields did not get updated with the new values, such as the "Ignore missing groups" switch for example.
@jenny-s51 did you get an error or did you get a success message and the fields were just not updated? i see now that the specific switch you call out isn't saving, were there others? i do know that some basic validation logic needs to be added to a bunch of the fields, i was going to add those later, but in those cases you would've gotten an error on submit.

> Where is the `config` you are reading from to set the `name` attribute? I searched the repo but the search didn't return any results with these values (i.e. `"config.always-read-value-from-ldap"`).
I'm not sure what you mean, is there a related error? always-read-value-from-ldap is in LdapMapperuserAttribute and it seems to be working fine

jenny-s51:
@mfrances17 just tested again - I'm always getting a save success for the group-ldap-mapper when I update any field (provided that the field input is valid; it would be helpful to use the error message from the server in the error alerts here, like we do in other places throughout Keycloak, but validation logic can be done in another PR as you mentioned).

Yes, the following fields do not update if I hit save and then navigate back to the form:
- Group object classes
- Ignore missing groups (switch)
- Membership LDAP attribute
- LDAP filter
- User groups retrieve strategy
- Drop non-existing groups during sync (switch)
- Groups path

mfrances17:
@jenny-s51 i've looked into it and i can't figure out why these group fields aren't saving properly. 

at this point i'd like to propose we disregard the groups mapper from this PR along with the already disregarded roles (they share the same form) and i will take care of them in the next PR. the rest of the mappers are working correctly, i'd like to get the code in if possible.

jenny-s51 (approved):
Sounds like a good plan @mfrances17 - LGTM

PR #551, opened 2021-04-27
add option to add borders on scroll panel

edewit:
Removes the borders from the client detail page scroll form and only adds them on the advanced tab.


jenny-s51 (approved):
Nice addition - LGTM 👍

PR #560, opened 2021-04-30
Fixing UXD design review

edewit:
* Removed divider
* Adding type field
* Hide "Consent screen text" field when display consent screen is off
* Fixed cancel button

fixing: #470

### DIFF
--- a/src/client-scopes/ClientScopesSection.tsx
+++ b/src/client-scopes/ClientScopesSection.tsx
@@ -12,7 +12,6 @@ import {
   ToolbarItem,
 } from "@patternfly/react-core";
 import { cellWidth } from "@patternfly/react-table";
-import type ClientScopeRepresentation from "keycloak-admin/lib/defs/clientScopeRepresentation";
 
 import { useAdminClient } from "../context/auth/AdminClient";
 import { ViewHeader } from "../components/view-header/ViewHeader";
--- a/src/client-scopes/ClientScopesSection.tsx
+++ b/src/client-scopes/ClientScopesSection.tsx
@@ -24,57 +24,13 @@ import {
   CellDropdown,
   ClientScope,
   AllClientScopes,
-  AllClientScopeType,
+  ClientScopeDefaultOptionalType,
+  changeScope,
+  removeScope,
 } from "../components/client-scope/ClientScopeTypes";
-import type KeycloakAdminClient from "keycloak-admin";
 import { ChangeTypeDialog } from "./ChangeTypeDialog";
 
 import "./client-scope.css";
-
-type ClientScopeDefaultOptionalType = ClientScopeRepresentation & {
-  type: AllClientScopeType;
-};
-
-const castAdminClient = (adminClient: KeycloakAdminClient) =>
-  (adminClient.clientScopes as unknown) as {
-    [index: string]: Function;
-  };
-
-const changeScope = async (
-  adminClient: KeycloakAdminClient,
-  clientScope: ClientScopeDefaultOptionalType,
-  changeTo: AllClientScopeType
-) => {
-  await removeScope(adminClient, clientScope);
-  await addScope(adminClient, clientScope, changeTo);
-};
-
-const removeScope = async (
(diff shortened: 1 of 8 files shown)


ssilvert:
@edewit I think this one needs a rebase to get Cypress to run?

ssilvert:
@xianli123 @edewit We need to talk about the Type field.

The wireframe has four types as shown here: https://marvelapp.com/prototype/5cf7cfd/screen/72040352
On the wireframe it shows "Default", "Always", "Optional", and "Required".

The current list screen has three types, "None", "Default", and "Optional".

In the UXD design review we have two types on the Settings tab, "Default" and "Optional".

Plus, in this PR, if you change the Type field, it is not saved.  It does save if you change it from the list screen.

Also, how does this relate to the old console?

xianli123:
@ssilvert I list some answers to your question.

> On the wireframe it shows "Default", "Always", "Optional", and "Required".
The current list screen has three types, "None", "Default", and "Optional".

In the new design, there are should be 4 types, but the backend doesn't support "Always" and "Required" now. Therefore, we just keep "Default" and "Optional" on the list. Regarding "None", it has the same meaning as "Available" in "Client scopes" inside the Client detail page. 
Screen Shot 2021-05-18 at 11 07 03 AM



> Plus, in this PR, if you change the Type field, it is not saved. It does save if you change it from the list screen.

I think the changes should be saved automatically in the list.



> Also, how does this relate to the old console?

We can refer to the "Client scopes" tab of one specific client in the old console, as shown in the picture below. Specifically, there are 3 types namely Default, Optional, None. "None" means available in the old console.
Screen Shot 2021-05-18 at 11 17 47 AM

@edewit Do you have any other ideas?

edewit:
No @xianli123 I implemented it, just like your explanation. @ssilvert saving the type is now fixed

mfrances17 (commented):
LGTM, tested all of the uxd requests in 470 and they all work. 

I'm quickly re-running the jobs for the PR, for some reason the build is failing and calling out clientScopeRepresentation as being defined but not used, when it is clearly being used. Will approve once it completes successfully.

ssilvert:
@xianli123 Thanks for the outstanding explanation.  I know we went through all that stuff ages ago but I had long forgotten all the details.  @andymunro needs to bookmark that explanation as it will be very useful for the documentation.

ssilvert:
@mfrances17 @jenny-s51 Please don't merge this yet.  

After @edewit fixes the conflicts I want to make sure all the type values are reflected so that you can set them in the new or old console and everything shows up properly in both.

xianli123:
@ssilvert @andymunro My pleasure. If you have any questions, please feel free to let me know.

ssilvert:
> @ssilvert I listed some answers here to your question.
> 
> > On the wireframe it shows "Default", "Always", "Optional", and "Required".
> > The current list screen has three types, "None", "Default", and "Optional".
> 
> In the new design, there are should be 4 types, but the backend doesn't support "Always" and "Required" now. Therefore, we just keep "Default" and "Optional" on the list. Regarding "None", it has the same meaning as "Available" in "Client scopes" inside the Client details page.
> Screen Shot 2021-05-18 at 11 07 03 AM
> 
> > Plus, in this PR, if you change the Type field, it is not saved. It does save if you change it from the list screen.
> 
> I think the changes should be saved automatically in the list.
> 
> > Also, how does this relate to the old console?
> 
> We can refer to the "Client scopes" tab of one specific client in the old console, as shown in the picture below. Specifically, there are 3 types namely Default, Optional, None. "None" means available in the old console.
> Screen Shot 2021-05-18 at 11 17 47 AM
> 
> @edewit Do you have any other ideas?

@andymunro To be clear, the Type only affects what happens to the client scope for a newly created client.  After client creation, the available default and optional scopes can be changed on a per-client basis. 

BTW, I've verified that it's all working as expected now with Type properly persisted.  Changes in the new console do get reflected in the old console.

ssilvert (approved):

Issue #436, opened 2021-03-12
[UX review] User federation - LDAP


yih-wang:
### **High priority**

[Empty state]

1. When hovering on the cards, the mouse pointer doesn't change to a hand.

[Creation page]

2. The old console provides ldap as the Console display name by default. It's a good feature to keep in the new console.
![image](https://user-images.githubusercontent.com/23331885/115397921-2e52ea00-a219-11eb-9442-34e0018e736c.png)
4. Change the value of Vendor should make some differences to "Username LDAP attribute", "RDN LDAP attribute", "UUID LDAP attribute" and "User Object Classes" in the _LDAP searching and updating_ section. I don't see this in the new console, not sure if there's a discussion happened before. Just bring this issue for your reference.
![image](https://user-images.githubusercontent.com/23331885/110935145-ff1fa200-8369-11eb-9f23-909969396207.png)
6. Clicking the Test button doesn't give me any feedback now. A success/error notification is supposed to display.
![image](https://user-images.githubusercontent.com/23331885/110936864-7d7d4380-836c-11eb-90ce-a8034ba4cb80.png)
7. When the bind type is none, the "Bind DN" and "Bind credentials" are not supposed to show. And the Test button only works for test connection. 
![image](https://user-images.githubusercontent.com/23331885/110936329-b1a43480-836b-11eb-8400-3697b18bfcad.png)
8. Following the last point, I think we can add some help info to the Test button. We used to have two separate buttons to test connection and authentication. But now, there is only one Test button to test both. And it is only used to test connection as mentioned in the last point. Maybe change the name of it to "Test connection and authentication" and "Test connection"?
13. The Periodic Full Sync and the Periodic Changed Users Sync are supposed to have toggles control them.

### **Low priority**

[Creation page]

1. The asterisks and the help icons don't keep in the same rows with some of the labels, I remember @jenny-s51 got a same comment during the App Service meeting. Maybe we can solve it as a global issue.
![image](https://user-images.githubusercontent.com/23331885/110934510-0eeab680-8369-11eb-9bbd-ff6dc4926623.png)
5. The Connection pooling toggle used to trigger a set of settings like the screen shows, but I don't see it in the new console. Not sure whether this has been discussed as well.
![image](https://user-images.githubusercontent.com/23331885/110935413-5887d100-836a-11eb-82f8-9bfa0d71a411.png)
9. I noticed that there is an extra button available in the creation page. Is it supposed to be added to the new console?
[...]

PR #569, opened 2021-04-30
Show hand cursor when hovering over User Fed empty state cards

mfrances17:
## Motivation
Partially fixes https://github.com/keycloak/keycloak-admin-ui/issues/436.

## Brief Description
Adds a hand cursor icon when hovering over the User federation empty state cards. Also adds a default value of 'ldap' to the console name field when creating a new LDAP provider.

## Verification Steps
1. Go to User Federation.
2. If there are existing LDAP/Kerberos providers configured, delete them from the card view.
3. Verify that the cursor changes to a hand icon when hovering over either of the cards in empty state view.
4. Click the LDAP card and verify that the default console name is listed as 'ldap'.

## Checklist:

## Additional Notes
None

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -19,10 +19,14 @@ label.pf-c-form__label {
 }
 
 .pf-c-select__toggle:before {
-  border-top: var(--pf-c-select__toggle--before--BorderTopWidth) solid var(--pf-c-select__toggle--before--BorderTopColor);
-  border-right: var(--pf-c-select__toggle--before--BorderRightWidth) solid var(--pf-c-select__toggle--before--BorderRightColor);
-  border-bottom: var(--pf-c-select__toggle--before--BorderBottomWidth) solid var(--pf-c-select__toggle--before--BorderBottomColor);
-  border-left: var(--pf-c-select__toggle--before--BorderLeftWidth) solid var(--pf-c-select__toggle--before--BorderLeftColor);
+  border-top: var(--pf-c-select__toggle--before--BorderTopWidth) solid
+    var(--pf-c-select__toggle--before--BorderTopColor);
+  border-right: var(--pf-c-select__toggle--before--BorderRightWidth) solid
+    var(--pf-c-select__toggle--before--BorderRightColor);
+  border-bottom: var(--pf-c-select__toggle--before--BorderBottomWidth) solid
+    var(--pf-c-select__toggle--before--BorderBottomColor);
+  border-left: var(--pf-c-select__toggle--before--BorderLeftWidth) solid
+    var(--pf-c-select__toggle--before--BorderLeftColor);
 }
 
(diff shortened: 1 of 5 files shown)


edewit (commented):
Isn't this a patternfly bug? I think we might be using Card in the wrong way? Or would it be better to put a `<a` inside it maybe instead?

edewit:
Also this should also be applied to the `IdentityProviderSection` as it uses similar layout

mfrances17:
> Isn't this a patternfly bug? I think we might be using Card in the wrong way? Or would it be better to put a `<a` inside it maybe instead?

I don't think i'd consider this a bug. In most cases i've used cards, they have clickable components inside like a link or a menu. They are also used for sections of the UI that are not necessarily clickable. So i've only added this for the empty state cards because they have no internal clickable items and are meant to be clicked.

ssilvert:
> > Isn't this a patternfly bug? I think we might be using Card in the wrong way? Or would it be better to put a `<a` inside it maybe instead?
> 
> I don't think i'd consider this a bug. In most cases i've used cards, they have clickable components inside like a link or a menu. They are also used for sections of the UI that are not necessarily clickable. So i've only added this for the empty state cards because they have no internal clickable items and are meant to be clicked.

I agree it's probably not a bug.

Have you tried putting a `<a` inside?  I'm always a fan of avoiding custom css.

mfrances17:
> Have you tried putting a `<a` inside? I'm always a fan of avoiding custom css.

@ssilvert @edewit Disagree. This is a perfect use case for CSS, as we have done in other parts of the UI. However, Erik's correct, the cards are used elsewhere so I will move the custom style to the component css e.g. keycloak-card.css.

mfrances17:
> Also this should also be applied to the `IdentityProviderSection` as it uses similar layout
IdentityProvidersSection uses Card directly, not the KeycloakCard component. I will be adding the custom style to the KeycloakCard component.

edewit:
it works fine and the cursor looks nice but in the generated html there is no hint that this could be clicked:

```html
<div class="pf-c-card__title">
   <div class="pf-l-split pf-m-gutter">
      <div class="pf-l-split__item">
         <svg fill="currentColor" height="2em" width="2em" viewBox="0 0 448 512" aria-hidden="true" role="img" style="vertical-align: -0.25em;"></path></svg>
      </div>
      <div class="pf-l-split__item pf-m-fill">Add LDAP providers</div>
   </div>
</div>
```
I think people that use screen readers will not pickup on the fact that you can click this.

mfrances17:
> I think people that use screen readers will not pickup on the fact that you can click this.
Hmmm, not sure but in this PR all i do is make a CSS change that alters the cursor on hover. I'm not adding the link - the link already preexisted via the card.

edewit (approved):

PR #576, opened 2021-05-03
Fixes with realm list

edewit:
realm list was queried multiple times and switching no longer worked this fixes that


jenny-s51 (commented):
Nice fixes @edewit - realm switching works great.

When I type something into the realm selector search box, the filtering works, but the "recent" realms show first and the searched realm is hidden by the "Create Realm" button underneath and a user cannot see it until they scroll through the recent realms first.

Screen Shot 2021-05-03 at 11 04 00 AM

Maybe we can hide the recent used realms if a user enters a search input? WDYT @edewit?

ssilvert:
> Nice fixes @edewit - realm switching works great.
> 
> When I type something into the realm selector search box, the filtering works, though the "recent" realms show first and the searched realm is hidden by the "Create Realm" button underneath and a user cannot see it until they scroll through the recent realms first.
> 
> Screen Shot 2021-05-03 at 11 04 00 AM
> 
> Maybe we can hide the recent used realms if a user enters a search input? WDYT @edewit?

@jenny-s51 Please add a new issue for that.  I think it's beyond the scope of this PR.

ssilvert (approved):

PR #579, opened 2021-05-03
Use NPM as the default package manager

jonkoops:
## Motivation
Makes it easier for new contributors to get started and reduces additional steps in the CI scripts.

## Brief Description
Replaces all yarn commands with their NPM equivalent, a new lockfile was generated using NPM.

## Verification Steps
- Run normal installation steps

## Checklist:


jonkoops:
Looks like the `build` task is failing because of some dependencies around WebPack that got moved around. I'll try to resolve it in a separate PR.

ssilvert:
@jonkoops This is the kind of decision you make at the beginning of a project and the team chose yarn over npm.  I honestly don't remember the reasons, but it's just the wrong time to change something so fundamental to our daily work.  

Perhaps when the first version is complete we could change if others decide they prefer npm.

jonkoops:
@ssilvert Sure no problem, perhaps we can discuss this in the future. The reason I added this is because I am slowly converting the setup so it makes use of Snowpack v3 and I wanted to clean up the project a little to match what is generated by Snowpack out of the box.

ssilvert:
@jonkoops Are you considering a PR to upgrade to Snowpack v3?  If so, we need to discuss as a team. 

We definitely want to do that eventually but I'm not sure if we would want to do that now.  In the regular Keycloak project we discuss major changes in the keycloak-dev mailing list, as I think you already know.

However, for this project, stuff like this that doesn't need input from the full Keycloak team can be discussed as GitHub issues.

Issue #577, opened 2021-05-03
Realm selector: update search filter to list only filtered realms


jenny-s51:
> When I type something into the realm selector search box, the filtering works, though the "recent" realms show first and the searched realm is hidden by the "Create Realm" button underneath and a user cannot see it until they scroll through the recent realms first.
> 
> <img alt="Screen Shot 2021-05-03 at 11 04 00 AM" width="308" src="https://user-images.githubusercontent.com/32821331/116894190-ca102d00-abff-11eb-9575-1002a86dae78.png">
> 

We should hide all the recent used realms if a user enters a search input that doesn't match any of the recent realms.

_Originally posted by @jenny-s51 in https://github.com/keycloak/keycloak-admin-ui/issues/576#issuecomment-831383240_

edewit:
fixed in #580

PR #580, opened 2021-05-04
filter recent used realms as well as others

edewit:
fixing: #577

### DIFF
--- a/src/components/realm-selector/RealmSelector.tsx
+++ b/src/components/realm-selector/RealmSelector.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useContext, useEffect, ReactElement } from "react";
+import React, { useState, useContext, ReactElement, useMemo } from "react";
 import { useHistory } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 
@@ -33,7 +33,6 @@ export const RealmSelector = () => {
   const { whoAmI } = useContext(WhoAmIContext);
   const [open, setOpen] = useState(false);
   const [search, setSearch] = useState("");
-  const [filteredItems, setFilteredItems] = useState<RecentUsedRealm[]>();
   const history = useHistory();
   const { t } = useTranslation("common");
   const recentUsed = new RecentUsed();
@@ -48,6 +47,16 @@ export const RealmSelector = () => {
           return { name: r.realm!, used: false };
         })
     );
+
+  const filteredItems = useMemo(() => {
+    if (search === "") {
+      return undefined;
+    }
+
+    return all.filter((r) =>
+      r.name.toLowerCase().includes(search.toLowerCase())
+    );
+  }, [search, all]);
 
   const RealmText = ({ value }: { value: string }) => (
     <Split className="keycloak__realm_selector__list-item-split">
@@ -70,26 +79,11 @@ export const RealmSelector = () => {
     </Button>
   );
 
-  const onFilter = () => {
-    if (search === "") {
-      setFilteredItems(undefined);
-    } else {
-      const filtered = all.filter(
-        (r) => r.name.toLowerCase().indexOf(search.toLowerCase()) !== -1
-      );
-      setFilteredItems(filtered);
-    }
-  };
-
   const selectRealm = (realm: string) => {
     setRealm(realm);
     setOpen(!open);
     history.push(`/${realm}/`);
   };
-
-  useEffect(() => {
-    onFilter();
-  }, [search]);
 
   const dropdownItems = realms.map((r) => (
     <DropdownItem
@@ -138,7 +132,6 @@ export const RealmSelector = () => {
           }}
           searchInputValue={search}
           onSearchInputChange={(value) => setSearch(value)}
-          onSearchButtonClick={() => onFilter()}
           className="keycloak__realm_selector__context_selector"
         >
           {(filteredItems || all).map((item) => (
--- a/src/components/realm-selector/RealmSelector.tsx
+++ b/src/components/realm-selector/RealmSelector.tsx
@@ -16,7 +16,6 @@ import {
 } from "@patternfly/react-core";
 import { CheckIcon } from "@patternfly/react-icons";
 
-import type RealmRepresentation from "keycloak-admin/lib/defs/realmRepresentation";
 import { toUpperCase } from "../../util";
 import { useRealm } from "../../context/realm-context/RealmContext";
 import { WhoAmIContext } from "../../context/whoami/WhoAmI";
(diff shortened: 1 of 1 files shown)


src/components/realm-selector/RealmSelector.tsx
```diff
@@ -16,23 +16,38 @@ import {
...
+    .map((name) => {
+      return { name, used: true };
+    })
+    .concat(
+      realms
+        .filter((r) => !recentUsed.used.includes(r.realm!))
+        .map((r) => {
+          return { name: r.realm!, used: false };
+        })
+    );
```
jonkoops, line 50: This will create a new instance and also do a possibly expensive transformation on every render. To prevent this you can memoize the value.

```ts
// This will prevent new instances of 'RecentUsed' from being constructed on every render.
// Instead only a single instance will be created an memoized beteen render cycles.
const recentUsed = useMemo(() => new RecentUsed(), []);

// Here 'all' will only be recalculated if one of it's dependencies change.
const all = useMemo<RecentUsedRealm[]>(() => recentUsed.used
  .map((name) => ({ name, used: true }))
  .concat(realms
    .filter((r) => !recentUsed.used.includes(r.realm!))
    .map((r) => ({ name: r.realm!, used: false }))
  ),
  [recentUsed.used, realms]
)
```

src/components/realm-selector/RealmSelector.tsx
```diff
@@ -16,23 +16,38 @@ import {
...
+  used: boolean;
+};
+
 export const RealmSelector = () => {
   const { realm, setRealm, realms } = useRealm();
   const { whoAmI } = useContext(WhoAmIContext);
   const [open, setOpen] = useState(false);
   const [search, setSearch] = useState("");
-  const [filteredItems, setFilteredItems] = useState<RealmRepresentation[]>();
+  const [filteredItems, setFilteredItems] = useState<RecentUsedRealm[]>();
```
jonkoops, line 36: Instead of using a side-effect when `search` change it would make more sense to make `filteredItems` a memoized value based on on `search`, so basically a computed property.

For example, if we remove the `onFilter` function we can do this:

```ts
const filteredItems = useMemo(() => {
  if (search === "") {
    return undefined;
  }

  return all.filter((r) => r.name.toLowerCase().includes(search.toLowerCase()))
}, [search, all])
```

This relies on the assumption that the `onSearchButtonClick` event is not actually really needed as `onSearchInputChange` already updates the `search` value which subsequently triggers the side-effect to update the results.

src/components/realm-selector/RealmSelector.tsx
```diff
@@ -126,18 +141,12 @@ export const RealmSelector = () => {
           onSearchButtonClick={() => onFilter()}
           className="keycloak__realm_selector__context_selector"
         >
-          {recentUsed.used.map((realm) => (
-            <ContextSelectorItem key={realm}>
-              <RealmText value={realm} /> <Label>{t("recent")}</Label>
+          {(filteredItems || all).map((item) => (
```
jonkoops, line 144: Note that you can also use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here to expicitly handle `undefined` and `null` values instead of 'falsy' values.

```suggestion
          {(filteredItems ?? all).map((item) => (
```

jonkoops (changes requested):

src/components/realm-selector/RealmSelector.tsx
edewit: you suggested that somewhere else as well I don't see the value in it as this list of items can only be undefined or an array of strings

src/components/realm-selector/RealmSelector.tsx
edewit: Why would that be better?

src/components/realm-selector/RealmSelector.tsx
jonkoops: Yeah, this is just a personal preference.

src/components/realm-selector/RealmSelector.tsx
jonkoops: In my opinion it is better because is immediately clear that `filteredItems` is a derived value from `search` and `all`. Whereas in the original code this is handled a little less transparently because you first have to set a different value, and then wait for a side-effect to be triggered, which will then set's the actual state you want.

Besides that all the logic is isolated in a single block of code, instead of spread out through the file in different places.

Also if for some reason in the future `search` would get a value that is not empty by default (for example coming from a query parameter) the following would happen.

1. The value of `search` is set to a value like `Hello World`
2. The component renders with `filteredItems` set to `undefined`
3. The `useEffect` triggers
4. The value of `filteredItems` is set to the actual content we want
5. This causes the component to rerender again since state has changed

This would mean that the component re-renders for no reason besides that we are doing this in a `useEffect`. This is due to fact that a `useEffect` is triggered **after** a render takes place (see the [React documentation](https://reactjs.org/docs/hooks-effect.html#example-using-hooks)). 

You can see this happen in this example: https://stackblitz.com/edit/react-ts-kfllpr
Wheras with a `useMemo` it only renders once: https://stackblitz.com/edit/react-ts-kbq7rk

src/components/realm-selector/RealmSelector.tsx
jenny-s51: This is a good idea.

jenny-s51 (commented):
Looks and works great @edewit 👍 I like @jonkoops's suggestion with the `useMemo` hook

src/components/realm-selector/RealmSelector.tsx
jenny-s51: > This will create a new instance and also do a possibly expensive transformation on every render. To prevent this you can memoize the value.
> 
> ```ts
> // This will prevent new instances of 'RecentUsed' from being constructed on every render.
> // Instead only a single instance will be created and memoized beteen render cycles.
> const recentUsed = useMemo(() => new RecentUsed(), []);
> 
> // Here 'all' will only be recalculated if one of it's dependencies change.
> const all = useMemo<RecentUsedRealm[]>(() => recentUsed.used
>   .map((name) => ({ name, used: true }))
>   .concat(realms
>     .filter((r) => !recentUsed.used.includes(r.realm!))
>     .map((r) => ({ name: r.realm!, used: false }))
>   ),
>   [recentUsed.used, realms]
> )
> ```

This is a good idea.

jenny-s51 (approved):
Looks great!

Issue #453, opened 2021-03-19
Client role details broken


ssilvert:
Go to Clients -> Client details -> Roles
Click on any role and you get `Cannot read property 'map' of null`

edewit:
ahh the new `UsersInRoleTab` doesn't work for client roles

PR #582, opened 2021-05-04
fixed undefined

edewit:
fixing: #453

### DIFF
--- a/src/realm-roles/UsersInRoleTab.tsx
+++ b/src/realm-roles/UsersInRoleTab.tsx
@@ -27,7 +27,7 @@ export const UsersInRoleTab = () => {
       first: first!,
       max: max!,
     });
-    return usersWithRole;
+    return usersWithRole || [];
   };
 
   const { enabled } = useContext(HelpContext);


src/realm-roles/UsersInRoleTab.tsx
```diff
@@ -27,7 +27,7 @@ export const UsersInRoleTab = () => {
       first: first!,
       max: max!,
     });
-    return usersWithRole;
+    return usersWithRole || [];
```
jonkoops, line 30: Note that you can also use the nullish coalescing operator here to expicitly handle `undefined` and `null` values instead of 'falsy' values.

```suggestion
    return usersWithRole ?? [];
```

jonkoops (approved):

src/realm-roles/UsersInRoleTab.tsx
edewit: but in this case it would be the same right? As it's either a `RoleRepresentation` or `undefined`. `undefined` in this case just means nothing was found.

jonkoops: The result would be the same, I prefer using `??` over `||` in cases where nullish values are checked. But not a blocker or anything. Might be something to make a linting rule for.

edewit: The difference is that `0 || 'test'` would be `'test'` and `0 ?? 'test'` would be `0` right. But here because we are using typescript we know that this value can't be "nullish" it can only be `RoleRepresentation[] | undefined` so that means it doesn't matter. That is the great thing about having types it's easier to do the right thing.

jonkoops: Exactly! TypeScript is awesome in this regard. Just a personal preference really, so pay it no mind.

mfrances17 (approved):
LGTM

PR #584, opened 2021-05-04
Extend ts config from Snowpack and fix type errors

jonkoops:
## Motivation
`@snowpack/app-scripts-react` provides a base configuration for TypeScript that can be extended. This configuration provides some specific compiler flags such as `isolatedModules` and `importsNotUsedAsValues` which help prevent errors when compiling with `esbuild`.

For more information see the [base configuration](https://github.com/snowpackjs/snowpack/blob/main/create-snowpack-app/app-scripts-react/tsconfig.base.json).

## Brief Description
The ts config in the project now extends the base config provided by `@snowpack/app-scripts-react`. Compilation errors resulting from this change have been fixed (mostly by supplementing import statements with `import type`).

## Verification Steps
The CI should pass, application still runs normally.

## Checklist:


edewit (approved):
We don't use a "per-file" transpilers atm, but it doesn't hurt to support that

jonkoops:
@edewit can you give this one a merge? :)

mfrances17 (approved):
LGTM, conflicts need to be resolved otherwise gtg

jonkoops:
I've rebased the PR and resolved the conflicts and updated the new code.

jonkoops:
> We don't use a "per-file" transpilers atm, but it doesn't hurt to support that

Isn't Snowpack a per-file transpiler, as it uses esbuild under the hood? According to the [esbuild documentation](https://esbuild.github.io/content-types/#isolated-modules):

> Even when transpiling a single module, the TypeScript compiler actually still parses imported files so it can tell whether an imported name is a type or a value. However, tools like esbuild and Babel (and the TypeScript compiler's `transpileModule` API) compile each file in isolation so they can't tell if an imported name is a type or a value.
>
> Because of this, you should enable the [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) TypeScript configuration option if you use TypeScript with esbuild. This option prevents you from using features which could cause mis-compilation in environments like esbuild where each file is compiled independently without tracing type references across files. For example, it prevents you from re-exporting types from another module using `export {T} from './types'` (you need to use `export type {T} from './types'` instead).

PR #588, opened 2021-05-06
Fixed bug

edewit:
`Console display name` had a default value of `ldap` making the tests fail
also added a small refactor to the tests putting declare and assign in the same line

### DIFF
--- a/src/user-federation/ldap/LdapSettingsGeneral.tsx
+++ b/src/user-federation/ldap/LdapSettingsGeneral.tsx
@@ -83,7 +83,7 @@ export const LdapSettingsGeneral = ({
             type="text"
             id="kc-console-display-name"
             name="name"
-            defaultValue=""
+            defaultValue="ldap"
             ref={form.register({
               required: {
                 value: true,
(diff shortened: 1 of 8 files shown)


src/user-federation/ldap/LdapSettingsGeneral.tsx
```diff
@@ -83,7 +83,7 @@ export const LdapSettingsGeneral = ({
             type="text"
             id="kc-console-display-name"
             name="name"
-            defaultValue="ldap"
```
mfrances17, line 86: @edewit The default value of 'ldap' was just recently added a few days ago as a must-fix request by UX (see item 2): https://github.com/keycloak/keycloak-admin-ui/issues/436
Apparently that's what the old console UI used to do as well.

mfrances17 (changes requested):

src/user-federation/ldap/LdapSettingsGeneral.tsx
edewit: ahh okay i thought it was an error, fixed it now by clearing the field first

mfrances17 (approved):
LGTM

Issue #589, opened 2021-05-06
Update global checkbox styles throughout app


jenny-s51:
Checkboxes should be larger on data tables across Keycloak admin console.

Example mock:

https://marvelapp.com/prototype/7i7ja65/screen/75693208

ssilvert:
@jenny-s51 Can this issue be closed?

jenny-s51:
@ssilvert yes

PR #590, opened 2021-05-06
Users: update checkbox styling and add disabled checkboxes/row functionality to modal

jenny-s51:
## Motivation

Towards #462 and #589 

JIRA:
https://issues.redhat.com/browse/APPDUX-957
https://issues.redhat.com/browse/APPDUX-956

## Verification Steps

1. Go to Users section.
2. Go to the groups tab.
3. Add groups.
4. Open the "Add groups" modal again and verify that the group you just added is not selectable.
5. Verify that the groups without subgroups do not have an angle icon at the end of the data list item and are not selectable.

## Checklist:


mfrances17 (changes requested):
Following the testing steps worked well:

- i was able to join groups as a user, and if i had already joined a group, it was no longer selectable
- if a group had a subgroup, there was a > that allowed me to select subgroups, and the > did not appear for any groups that did not have subgroups

One issue I saw, is that if you join a subgroup of a group, it will be listed in the group membership column successfully. However if you go back and list the groups, you can still choose its parent group and add it, with a success message, but the parent group is never added to the group membership column. As opposed to doing it in a different sequence, you can add a parent group successfully, and then go back and add its child successfully, and both will display in the membership column. You just can't add the child first and then the parent, or the parent will fail to be listed even though it says successful.

jenny-s51:
@mfrances17 thank you for bringing up the subgroups anomaly - @ssilvert and I had a meeting about the parent/child group adding sequence and we discovered the anomaly too. In the old console it currently does the same thing due to the way the call is made to the API (adding parent -> child works and shows both, but adding child -> parent never lists the parent separately). This PR was only meant to update the checkbox styling and add disabled rows/checkboxes in the modal - are you okay with merging?

@ssilvert @edewit what do you think is the best route to take with the groups listing? Is there something that needs to be updated in the back end to get the groups to show correctly (regardless of the order they are added)?

mfrances17:
@jenny-s51 yes absolutely, i'll approve this PR based on styling changes, as it was intended.

we might want to capture the group/subgroup anomaly in a separate github issue, so it is not forgotten. although if it works like this in the old admin console, it should probably be lower priority as we've always kind of gone with the 'make it work like the old console' mantra.

mfrances17 (approved):
Approving based on the styling changes, as was intended by the PR

mfrances17:
This is approved, and only the notorious group test is failing (which should now be fixed on CI), i'm moving forward with the merge.

PR #593, opened 2021-05-07
fix groups test

jenny-s51:
# Motivation
Fixes the groups cypress tests.

# Brief Description
Allows the cypress tests to work without failures.

# Verification Steps
Run the identity providers cypress tests locally, or wait for CI to process them.


edewit (commented):
@jenny-s51 is it really the problem that we can't use `getId` it does the same as `get("data-testid="` if the function is there it should work right? I mean I can't think why it would help not to have this helper function

PR #602, opened 2021-05-10
fixing issues described in 595

edewit:
fixing: #595

### DIFF
--- a/src/identity-providers/add/DiscoverySettings.tsx
+++ b/src/identity-providers/add/DiscoverySettings.tsx
@@ -2,6 +2,7 @@ import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
 import { useFormContext, useWatch } from "react-hook-form";
 import {
+  Divider,
   ExpandableSection,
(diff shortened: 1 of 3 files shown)


mfrances17 (changes requested):
could be user error, but i don't see a visible option to show/hide metadata option when creating an openID Connect provider:
no-metadata

also, when viewing settings of an existing openid connect provider, looks like an extra divider is being shown:
extra-divider

edewit:
@mfrances17 this option is available after you have used an url to load this information put in `http://localhost:8180/auth/realms/master/.well-known/openid-configuration` as a discovery endpoint to see that

mfrances17:
> @mfrances17 this option is available after you have used an url to load this information put in `http://localhost:8180/auth/realms/master/.well-known/openid-configuration` as a discovery endpoint to see that

Ah, ok... yes that worked great. Thanks for the info.

mfrances17 (approved):
LGTM, code is good and it looks like all of the issues were addressed in my testing.

PR #605, opened 2021-05-12
add `/` after baseUrl

edewit:
## Motivation
Base url is now invalid because it doesn't have a slash at the end resulting in Urls with `authreamls`

## Verification Steps

go to the clients section and see that the urls are valid or the realm settings


mfrances17 (approved):
LGTM, verified that the links in the Clients Home URL column are now formatted correctly and work as expected (e.g. the links now appear as **server/auth/realms/...** instead of **server/authrealms/...**, **server/auth/admin/...** as opposed to **server/authadmin/...** etc

PR #618, opened 2021-05-20
Introduces new Realm Events tab

edewit:
## Brief Description

Realm event tab as described by [marvel](https://marvelapp.com/prototype/6ac0aja/screen/72967364) it's still missing the
event listener tab, but has all the other screens

## Verification Steps

Go to realm settings there you'll find the new event tab

## Checklist:


src/realm-settings/help.json
```diff
@@ -10,6 +10,12 @@
     "loginTheme": "Select theme for login, OTP, grant, registration and forgot password pages.",
     "accountTheme": "Select theme for user account management pages.",
     "adminConsoleTheme": "Select theme for admin console.",
-    "emailTheme": "Select theme for emails that are sent by the server."
+    "emailTheme": "Select theme for emails that are sent by the server.",
+    "save-user-events": "If enabled, login events are saved to the database, which makes events available to the admin and account management consoles.",
+    "save-admin-events": "If enabled, admin events are saved to the database, which makes events available to the admin console.",
+    "expiration": "Sets the expiration for events. Expired events are periodically deleted from the database.",
+    "admin-clearEvents": "Deletes all events in the database.",
```
mfrances17, line 17: the strings are the same, why not just have one 'clearEvents' string?

src/realm-settings/messages.json
```diff
@@ -14,6 +14,15 @@
     "general": "General",
     "login": "Login",
     "themes": "Themes",
+    "events": "Events",
+    "userEventsConfig": "User events config",
```
mfrances17, line 18: could wait for the doc review, but most likely we will probably want to spell out configuration instead of config

mfrances17 (changes requested):
Looks good - a couple of potential string improvements, plus noticed the following when testing:

We should probably only enable Save if the form is dirty. So e.g. right now in the user events config, after i click Save i can keep clicking it w/o making any changes.

Should Revert revert all changes even after a proper Save, as it is currently doing? I thought revert was more like an undo of changes before a save is performed, but once you save, your data should still be saved.

Full sentences in the confirm dialog like 'If "Save events is disabled, subsequent...' and 'If you clear all events of this realm, ...' should end in a period.

When i delete one of the Event saved types, the following notification appears:
'Successfully saved configuration'
That's probably not the intended notification message.

The 'User events settings' and 'Admin events settings' tabs seem to be spaced tightly under the main tab row. Personally i think there should be more space between them but as always go with the UX designers/Marvel spec.

src/realm-settings/help.json
```diff
@@ -14,8 +14,8 @@
     "save-user-events": "If enabled, login events are saved to the database, which makes events available to the admin and account management consoles.",
     "save-admin-events": "If enabled, admin events are saved to the database, which makes events available to the admin console.",
     "expiration": "Sets the expiration for events. Expired events are periodically deleted from the database.",
-    "admin-clearEvents": "Deletes all events in the database.",
-    "user-clearEvents": "Deletes all events in the database.",
+    "admin-clearEvents": "Deletes all user events in the database.",
```
mfrances17, line 17: it looks like you got these backwards :) quick fix

mfrances17 (approved):
Most of the issues I brought up in my original review have been addressed, I'm ok with getting this in.

Issue #607, opened 2021-05-13
[UX Review] Realm settings - General


yih-wang:
1. The Endpoint link name should be "OpenID Endpoint Configuration" instead of "Open Endpoint Configuration "
![image](https://user-images.githubusercontent.com/23331885/118091685-b2098c00-b3fd-11eb-8a68-2f15a6434418.png)

@edewit

yih-wang:
Click Save doesn't give me any feedback

PR #633, opened 2021-05-25
fixes realm save

edewit:
currently save on realm screen didn't work as the email settings where required. Fixed this by changing email to be it's own section. Also email section was using the wrong attributes therefor the changes where not reflected in the old console

### DIFF
--- a/src/realm-settings/KeysListTab.tsx
+++ b/src/realm-settings/KeysListTab.tsx
@@ -2,42 +2,51 @@ import React, { useState } from "react";
 import { useHistory, useRouteMatch } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import { Button, ButtonVariant, PageSection } from "@patternfly/react-core";
+import { cellWidth } from "@patternfly/react-table";
+
 import type { KeyMetadataRepresentation } from "keycloak-admin/lib/defs/keyMetadataRepresentation";
+import type ComponentRepresentation from "keycloak-admin/lib/defs/componentRepresentation";
 import { ListEmptyState } from "../components/list-empty-state/ListEmptyState";
 import { KeycloakDataTable } from "../components/table-toolbar/KeycloakDataTable";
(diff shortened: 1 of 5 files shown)


edewit:
fixing comment: https://github.com/keycloak/keycloak-admin-ui/issues/607#issuecomment-846941758

src/realm-settings/KeysListTab.tsx
```diff
@@ -2,42 +2,51 @@ import React, { useState } from "react";
...
+    const keysMetaData = await adminClient.realms.getKeys({
+      realm: realmName,
+    });
+    const keys = keysMetaData.keys;
 
-  React.useEffect(() => {
-    refresh();
-  }, [keys]);
+    return keys?.map((key) => {
+      const provider = realmComponents.find(
```
jenny-s51, line 44: For some reason `realmComponents` is always `undefined` if called in the loader - the providers column of the table is empty when provider is initialized this way. Splitting up into `KeysListTab` into two components resolved this issue and the provider is rendered for each row

jenny-s51 (changes requested):
Save/email form updates look great @edewit ! How did you know to use the `smptServer` object? It looked like the old console was also using `attributes`.

Also, a small thing - if the KeysListTab is consolidated into one component, props passed into the loader are `undefined` when accessed, and the `provider` column is empty for each row if the KeysList. This is why it was split up into two components initially. 

What do you think is the best thing to do here? Should we split it up again or is there a way we can just use the one component?

src/realm-settings/KeysListTab.tsx
edewit: right it's a timing issue the table gets rendered when the realmComponent array is still empty. Fixed it by making the render depended on the realmComponent array being loaded.

jenny-s51 (approved):
Thank you for clarifying - updates look awesome 👍

PR #634, opened 2021-05-25
fixed type of save button

edewit:
fixing: #632

### DIFF
--- a/src/clients/advanced/SaveReset.tsx
+++ b/src/clients/advanced/SaveReset.tsx
@@ -9,14 +9,14 @@ type SaveResetProps = ActionGroupProps & {
 };
 
 export const SaveReset = ({ name, save, reset, ...rest }: SaveResetProps) => {
-  const { t } = useTranslation();
+  const { t } = useTranslation("common");
   return (
     <ActionGroup {...rest}>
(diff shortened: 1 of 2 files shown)


src/clients/advanced/SaveReset.tsx
```diff
@@ -9,14 +9,14 @@ type SaveResetProps = ActionGroupProps & {
 };
 
 export const SaveReset = ({ name, save, reset, ...rest }: SaveResetProps) => {
-  const { t } = useTranslation();
+  const { t } = useTranslation("common");
   return (
     <ActionGroup {...rest}>
-      <Button data-testid={name + "Save"} variant="tertiary" onClick={save}>
```
mfrances17, line 15: Was there a special request to use the tertiary variant here? I would think we'd want this to be primary (the default).

edewit: right that's why this PR changes it to be "default"

mfrances17: @edewit DOH! sorry, i read the diff wrong... all set, thanks

mfrances17 (approved):
LGTM

PR #642, opened 2021-05-25
Realm settings(keys): toggle provider modal for all provider types

jenny-s51:
## Verification Steps

1. Go to Realm settings -> Keys -> Providers tab.
2. Create a provider for each of the options listed.
3. Verify that the provider shows in the table.

## Checklist:


src/realm-settings/AddProviderModal.tsx
```diff
@@ -0,0 +1,759 @@
...
+                      key={`email-theme-${idx}`}
+                      value={p}
+                    ></SelectOption>
+                  ))}
+                </Select>
+              )}
+            />
+          </FormGroup>
+        )}
+        {providerType === "hmac-generated" && (
```
edewit, line 504: maybe it would be better to split this up into multiple smaller components based on the provider type?

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -97,8 +112,21 @@ export const KeysTabInner = ({ components }: KeysTabInnerProps) => {
...
+  };
+
   return (
     <>
+      {isCreateModalOpen && (
+        <AddProviderModal
+          handleModalToggle={handleModalToggle}
+          providerType={defaultConsoleDisplayName}
+          refresh={refresh}
+          open={isCreateModalOpen}
```
edewit, line 126: `open` is always true because if it's not it's not rendered

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@snowpack/app-scripts-react/tsconfig.base.json",
   "compilerOptions": {
+    "jsx": "react",
```
edewit, line 4: why is this?

edewit (changes requested):
I get this warning in the console on the rsa provider
image

src/realm-settings/KeysProvidersTab.tsx
jenny-s51: good point - fixed

tsconfig.json
jenny-s51: don't need this anymore

src/realm-settings/AddProviderModal.tsx
jenny-s51: That makes sense! I split up the modals into separate components

edewit (approved):

Issue #628, opened 2021-05-24
[UX Review] Realm creation


yih-wang:
1. When I create a realm by importing JSON/ paste the JSON code, the Create button isn't enabled.
![image](https://user-images.githubusercontent.com/23331885/119323379-363af980-bcb1-11eb-8cac-324c87dd045e.png)
2. I tried to paste and edit the code in the import component. But I can only type one character each time, then this component loses focus.
![image](https://user-images.githubusercontent.com/23331885/119323573-6f736980-bcb1-11eb-88de-6d55fbc896f0.png)
3. Following the last issue, when I import the realm by pasting the code. The realm id won't be filled automatically. 
4. The create button is only disabled when the initial state. But I suppose it should be disabled whenever the required field hasn't been filled.
5. I tried to create a disabled realm (turn off the enable toggle on the creation page). But it's still enabled after it was created.

@jenny-s51 @edewit I

PR #643, opened 2021-05-26
fixes issue around importing files

edewit:
fixes: #628

### DIFF
--- a/src/clients/import/ImportForm.tsx
+++ b/src/clients/import/ImportForm.tsx
@@ -32,7 +32,7 @@ export const ImportForm = () => {
 
   const { addAlert } = useAlerts();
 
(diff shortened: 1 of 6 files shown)


ssilvert:
@yih-wang Can you review this please?  Trying to get old PR's merged.

yih-wang (changes requested):
For the 2nd issue - 

> I tried to paste and edit the code in the import component. But I can only type one character each time, then this component loses focus.

This function still cannot be used normally. Each time I type a character, the cursor will jump to the end of the code.

yih-wang (approved):

yih-wang:
Thanks for the updates @edewit. Looks good to me!

PR #644, opened 2021-05-26
makes client scope test more reliable

edewit:
Added a wait for the loading to be done just like on the group test


mfrances17 (approved):
LGTM and test seems to be passing on CI so 👍

Issue #640, opened 2021-05-25
[UX Review] Users


yih-wang:
[Creation]
1. The Enable switch should be ON by default.
![image](https://user-images.githubusercontent.com/23331885/119492661-e97b1f80-bd91-11eb-9298-9a380a47139b.png)

2. The required user actions dropdown doesn't work well. The close icon doesn't work. And each time I click "Show more"/"Show less"/the close icon in the chip, the dropdown will be opened/closed.

[Join group modal]

3. Can we have pagination in the Group modal?
![image](https://user-images.githubusercontent.com/23331885/119493380-b6855b80-bd92-11eb-98dd-5f55432bed0a.png)

4. The checkboxes and group names should have some space.
![image](https://user-images.githubusercontent.com/23331885/119493453-c8ff9500-bd92-11eb-91d1-962f37db58eb.png)

5. The search box doesn't work.

6. Is it possible to hide the breadcrumb in the first level? Like this - https://marvelapp.com/prototype/7i7ja65/screen/78142144

7. We should show the path of the group in the chips. Or the groups with the same name will be confused. Reference this screen - https://marvelapp.com/prototype/7i7ja65/screen/78142146

yih-wang:
I'll create a separate issue to track the updates of the Join groups feature. Other issues look good.

PR #648, opened 2021-05-27
Fixing UXD review for user

edewit:
fixing: #640


ssilvert:
@yih-wang Please review.

yih-wang (changes requested):
The issue of Required actions is still there - each time I click "Show more"/"Show less"/the close icon in the chip, the dropdown will be opened/closed - it shouldn't be influenced. Reference the [PF component](https://www.patternfly.org/v4/components/select#multiple). 

Also, a new option "Update User locale" is added to the menu in the latest Keycloak version. Do I need to raise this request in a separate issue?
 image

For the "Join groups" modal, I think there's still something that needs to be discussed. And I'm wondering if it is reusing the component of the "Move to" group modal. It may make more sense to open a new issue tracking this so that it won't block this pr.

edewit:
@yih-wang This is what the patternfly component does has the same behavior as the reference page that you link

ssilvert:
I'm merging this now.

@yih-wang Please open issues for anything remaining that needs attention.

Issue #629, opened 2021-05-24
[UX Review] Groups


yih-wang:
[Empty state]
1. Is it possible to remove the toolbar in the Empty state? As there are no groups yet, the toolbar doesn't have any use here.
![image](https://user-images.githubusercontent.com/23331885/119337208-21fef880-bcc1-11eb-922a-c51804e81bac.png)
2. Maybe we can disable the Actions when there are no groups. The search group function doesn't make sense at the time.
@edewit

yih-wang:
[Search globally]

1. A breadcrumb is missing so I can't find a way to return to the group list.
![image](https://user-images.githubusercontent.com/23331885/119338950-33e19b00-bcc3-11eb-864f-a0ff82306267.png)

2. I'm doubting this page is forgot to be updated so its style is obviously not aligned with all other pages (it's in the old style we used to apply). Can we adjust the background and table the same with other pages?
![image](https://user-images.githubusercontent.com/23331885/119339749-24af1d00-bcc4-11eb-98a7-5d3ea52344f8.png)

3. The divider line is too far away from the toolbar, is it possible to narrow the space between them?
![image](https://user-images.githubusercontent.com/23331885/119342276-6c837380-bcc7-11eb-88a4-be0b1075a799.png)

yih-wang:
The header here should be the Group name, not always the "Groups"

![image](https://user-images.githubusercontent.com/23331885/119343163-a739db80-bcc8-11eb-86fe-a406df1fb406.png)

edewit:
> The paginations are not aligned with each other. Also I found it's a generic issue around the console. Can we fix it by changing the template?

What do you mean by this? I think it's correct to show that it's showing items 1 till 2 in this case?

edewit:
> The header here should be the Group name, not always the "Groups"

"Groups" in this case is the root it's not a specific group, if it's unclear it could also be called "Group list" maybe?

yih-wang:
> > The paginations are not aligned with each other. Also I found it's a generic issue around the console. Can we fix it by changing the template?
> 
> What do you mean by this? I think it's correct to show that it's showing items 1 till 2 in this case?

Here I mean the top pagination and the bottom pagination are not aligned. The bottom pagination has a right margin but the top one doesn't.

yih-wang:
> > The header here should be the Group name, not always the "Groups"
> 
[...]

PR #655, opened 2021-06-01
fixing issue 629

edewit:
fixing: #629


ssilvert:
@yih-wang Please review.

yih-wang (changes requested):
Here are the issues that haven't been addressed 

[Issue 1 - Groups list]
> When I search in the list and there is no matched result. The default empty state displays and ask the user to create a group, which doesn't make sense here. We can show a simple no result screen to users. Reference the Roles list screen I paste below.
image

[Issue 2 - Global search]
> A breadcrumb is missing so I can't find a way to return to the group list.
image

The design has been updated in the original link, you can reference [this screen](https://marvelapp.com/prototype/228971ig/screen/79872064)

[Issue 3 - Global search]
> This page hasn't been updated so its style is obviously not aligned with all other pages (it's in the old style we used to apply). Can we adjust the background and table to the same as other pages?
image

The design has been updated in the original link, you can reference [this screen](https://marvelapp.com/prototype/228971ig/screen/79872065)

yih-wang:
I'll open a separate issue for the "Move to" function so don't cover the issues of it in this thread.

edewit:
@yih-wang so this one is good?

ssilvert:
@yih-wang I'm going to go ahead and merge this.  Please create issues for any remaining work.

ssilvert (approved):

PR #660, opened 2021-06-02
fixing issue 416

edewit:
@jenny-s51 already did some of the things in this issue...


mfrances17 (changes requested):
Changes look good. Noticed a couple of things that should be fixed:

If there are multiple role names in the table, if you check the row of a role name and then select 'Remove' from a different row that is not checked, it will delete the checked role name. Best way to fix this would probably be to not enable the Remove command unless its row is checked.

Very odd text on this confirm modal, seems like a typo:
what

mfrances17 (approved):
LGTM, both of the issues i brought up in my review have been addressed. Nice work.

Issue #636, opened 2021-05-25
[UX Review] Identity provider


yih-wang:
[Manage order]
1. The providers that are disabled shouldn't be included in the modal.

@edewit

yih-wang:
[Add Keycloak OIDC]

The configuration should be the same with the Add OIDC, not Add Social provider.

yih-wang:
[Add Github provider]

1. There is an additional "/" in the redirect URI
![image](https://user-images.githubusercontent.com/23331885/120294897-3cf0ee80-c2f9-11eb-8e59-51c0859b0d5d.png)

yih-wang:
[Github provider details - Advanced]
1. Default scopes should be changed to Scopes
2. Post login flow - seems the old console doesn't include "clients" flow in the options but the new console does. Also, the old console has an empty option, which is also the default option. Maybe we should have a "Choose..." option in the new console to reflect that?
![image](https://user-images.githubusercontent.com/23331885/120298017-43349a00-c2fc-11eb-8399-cb4b5020688a.png)
3. Change the default value of the "Sync mode" to "legacy" instead of "Import".

yih-wang:
[github provider details save/reload]
1. The Save button should be in the primary button style.
2. The Revert button doesn't work.

yih-wang:
[Github provider details - General]

1. The display order is broken...
![image](https://user-images.githubusercontent.com/23331885/120299089-43816500-c2fd-11eb-8171-97c4e45885cd.png)

PR #663, opened 2021-06-02
fixes issue

edewit:
fixes: #636


ssilvert:
@yih-wang Please review.

yih-wang (changes requested):
[Creation - Keycloak OIDC]
1. The title of the Keycloak OIDC creation page should be "Add Keycloak OpenID Connect provider" instead.
image

[Creation - Social providers]
1. When creating social providers, the leaf (last node) of the breadcrumb should be the same as the header, e.g. "Add Github provider" instead of "Provider"
image

[Detail page]
1. The header should be the Provider name/Alias, e.g. "Github" instead of "Add Github provider"
image
2. For the client secret, I just noticed that the credentials are both encrypted no matter show or hide it using the eye icon... And the old console does it in the same way. Do you think it is because the safety reason? If so, maybe we could only keep the eye icon on the creation page, and remove the one in the detail page as it doesn't have any use...
3. For the post login flow, we should allow users to still choose the empty option ("Choose...") after they switch to others. The tooltip says "Leave this empty if you need no any additional authenticators to be triggered after login with this identity provider." Or does it make more sense to change it from "Choose..." to "None"? 
4. Change the default value of the "Sync mode" to "legacy" instead of "Import".
5. The Revert button will set all the fields to the default values now. It's not correct. It should clear the unsaved changes.

[Detail page - specifically to social provider]
1. The display order field on the social provider page hasn't been corrected. The "-" should be disabled when the value is 0.
image

[Detail page - specifically to OIDC provider]
1. I got two divider lines here, which seems not correct...
image

edewit:
> For the client secret, I just noticed that the credentials are both encrypted no matter show or hide it using the eye icon... And the old console does it in the same way. Do you think it is because the safety reason? If so, maybe we could only keep the eye icon on the creation page, and remove the one in the detail page as it doesn't have any use...

Yes that is something the server does the password is not saved as clear text for security reasons. I'll change it so that the eye icon is only there in the create page

ssilvert:
@edewit I went through Haley's comments.  Looks like Revert still doesn't work.  

Also, Post login flow and Sync mode have defaults set instead of being blank like the old console.  If you have verified that those are the proper defaults then I think this is good.

Let's try to get this merged before the end of the week.  If there isn't time to fix the rest then let's just create issues.

yih-wang:
> Also, Post login flow and Sync mode have defaults set instead of being blank like the old console. If you have verified that those are the proper defaults then I think this is good.

@ssilvert If the post login flow is not empty by default, should we still provide an empty option to users? The tooltip has the instructions like "Leave this empty if you need no any additional authenticators to be triggered after login with this identity provider." If no, the tooltip needs to be updated.

By the way, is there a way for me to check the default value of a field in code? For example, check which file or search what keywords. I'd like to know how the code is implemented instead of how it is displayed on the interface. It will give me a better understanding of the settings.

yih-wang:
@edewit I'll create separate issues for the remaining things that need to be fixed. The breadcrumbs and headers kind of mix up between different pages. Separate them could make things easier.

yih-wang:
Looks good to me now, thanks!

ssilvert:
> > Also, Post login flow and Sync mode have defaults set instead of being blank like the old console. If you have verified that those are the proper defaults then I think this is good.
> 
> @ssilvert If the post login flow is not empty by default, should we still provide an empty option to users? The tooltip has the instructions like "Leave this empty if you need no any additional authenticators to be triggered after login with this identity provider." If no, the tooltip needs to be updated.
> 
I think we can assume that the tooltip is correct as it was written by the original author of the feature.

So we do need an "empty" option.

yih-wang:
> I think we can assume that the tooltip is correct as it was written by the original author of the feature.
> 
> So we do need an "empty" option.

A "None" option is added now and set as the default value. Does it make sense? I'm not sure whether it should be the default value.

ssilvert:
> > I think we can assume that the tooltip is correct as it was written by the original author of the feature.
> > So we do need an "empty" option.
> 
> A "None" option is added now and set as the default value. Does it make sense? I'm not sure whether it should be the default value.

Unless someone told you otherwise, we should assume it should be just like the old console.  "None" is the default for Post Login Flow.  "Import" is the default for Sync Mode.

Also, looks like "None" should not be an option for Sync Mode.

ssilvert:
@yih-wang I'm going to go ahead and merge this as it is well past the deadline.  If you still see problems then please log an issue.

PR #678, opened 2021-06-07
changed location due to new snowpack postcss plugin

edewit:
location of fonts is now different because of new postcss plugin

https://github.com/keycloak/keycloak-admin-ui/pull/627


edewit:
`yarn install` and then start to see that the font are loaded

mfrances17 (approved):
LGTM - rebuilt node_modules with snowpack/plugin-postcss 1.4.0 and the font issue is fixed.

PR #691, opened 2021-06-09
Consents UI fixes

### LINKED ISSUE #668: [UX Review] User - Consents
Hi @jenny-s51, I've reviewed the "Consents" tab, there are some finetuning. Please check them.

> Screen Shot 2021-06-04 at 3 07 56 PM

1. We can remove this redundant line.
2. There is no distance between the tab bar and the border.
3. We can add a search bar here in order to help users search for some info.
4. If there are too many “Granted client scopes" here to show, we can add a label to show the number of the hidden client scopes, as shown in the picture below. 

> Screen Shot 2021-06-04 at 3 38 58 PM

5. Each row should have a kebab icon. Click the kebab icon can revoke the consent.

> Screen Shot 2021-06-04 at 3 39 34 PM

6. There is no distance between the table and the border.

jenny-s51:
## Motivation

Closes #668 

## Verification Steps

1. Go to client scopes -> create a new client scope.
2. Set the “Type” to “default” and make sure “Display on consent screen” is enabled.
3. Repeat for however many client scopes you want to create/grant.
4. Go to Clients and make sure both security-admin-console and security-admin-console-v2 have “Consent required” set to “on” under Login Settings
5. Go to Clients -> Client Scopes and then add the new client scopes as “Default”
6. Log out of both the new admin console and the old one
7. Log back into both
8. It should ask you to verify the new consents -> click accept/yes
9. Go to Users -> admin -> consents and you should see them listed there

## Checklist:

### DIFF
--- a/src/user/UserConsents.tsx
+++ b/src/user/UserConsents.tsx
@@ -1,7 +1,7 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 import { useTranslation } from "react-i18next";
-import { PageSection } from "@patternfly/react-core";
+import { Button, Label, PageSection } from "@patternfly/react-core";
 import { ListEmptyState } from "../components/list-empty-state/ListEmptyState";
 import { KeycloakDataTable } from "../components/table-toolbar/KeycloakDataTable";
 import { emptyFormatter } from "../util";
@@ -27,10 +27,31 @@ export const UserConsents = () => {
     return alphabetize(consents);
   };
 
+  const [labelClicked, setLabelClicked] = useState(false);
+
+   useEffect(() => {
+    console.log(labelClicked)
+  }, [labelClicked])
+
   const clientScopesRenderer = ({
     grantedClientScopes,
   }: UserConsentRepresentation) => {
-    return <>{grantedClientScopes!.join(", ")}</>;
+    if (grantedClientScopes!.length <= 5 && labelClicked) {
+      return <>{grantedClientScopes!.join(", ")}</>;
+    } else
+      return (
+        <>
+          {grantedClientScopes!.slice(1, 6).join(", ")}{" "}
+          <Label
+            color="blue"
+            style={{ cursor: "pointer" }}
+            variant="outline"
+            onClick={() => setLabelClicked(true)}
+          >
+            {grantedClientScopes!.length - 5 + " more"}
+          </Label>
+        </>
+      );
   };
 
   const createdRenderer = ({ createDate }: UserConsentRepresentation) => {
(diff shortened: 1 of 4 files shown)


mfrances17 (changes requested):
Everything looks like it was addressed, well done. Thanks for adding the expanded verification steps, it helped a lot. One small change to handle the string better than using the concatenation.

src/user/UserConsents.tsx
```diff
@@ -43,52 +70,83 @@ export const UserConsents = () => {
     return <>{moment(lastUpdatedDate).format("MM/DD/YY hh:MM A")}</>;
   };
 
+  const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
+    titleKey: "users:revokeClientScopesTitle",
+    messageKey: t("users:revokeClientScopes") + selectedClient?.clientId + "?",
```
mfrances17, line 75: Recommend doing something like this instead of concatenating the string here:
```
    messageKey: t("users:revokeClientScopes", {
      clientId: selectedClient?.clientId,
    }),
```

src/user/messages.json
```diff
@@ -60,6 +60,11 @@
     "noConsents": "No consents",
     "noConsentsText": "The consents will only be recorded when users try to access a client that is configured to require consent. In that case, users will get a consent page which asks them to grant access to the client.",
     "whoWillAppearLinkText": "Who will appear in this group list?",
-    "whoWillAppearPopoverText": "Groups are hierarchical. When you select Direct Membership, you see only the child group that the user joined. Ancestor groups are not included."
+    "whoWillAppearPopoverText": "Groups are hierarchical. When you select Direct Membership, you see only the child group that the user joined. Ancestor groups are not included.",
+    "revoke": "Revoke",
+    "revokeClientScopesTitle": "Revoke all granted client scopes?",
+    "revokeClientScopes": "Are you sure you want to revoke all granted client scopes for ",
```
mfrances17, line 66: And change the message to:
```
    "revokeClientScopes": "Are you sure you want to revoke all granted client scopes for {{clientId}}?",
```

src/user/UserConsents.tsx
ssilvert: > Recommend doing something like this instead of concatenating the string here:
> 
> ```
>     messageKey: t("users:revokeClientScopes", {
>       clientId: selectedClient?.clientId,
>     }),
> ```

Good catch @mfrances17.  This is actually not just a recommendation.  It's a must-do.  In different languages, the placement of the param within the sentence can be different, which means that you can't translate it correctly if that placement is hard-coded.  So concatenation of localized strings should almost always be avoided.

Also concur with @mfrances17 overall.  I tried it out and it looks really good.

jenny-s51: @ssilvert That makes sense, thank you for your helpful clarification on this

jenny-s51:
@mfrances17 @ssilvert Thank you for your helpful feedback and reviews! I've addressed the localization instance per your review comments

PR #696, opened 2021-06-11
Test email connection

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-936

## Verification Steps

1. From your local version of https://github.com/keycloak/keycloak, `cd` to `testsuite/utils` 
2. Run `mvn exec:java -Pmail-server` 
3. Go to Realm settings -> email
4. Enter `localhost` in the "Host" form field.
5. Enter `3025` in the "Port" form field.
6. Click the "Test Connection" button on the bottom of the page.
7. Verify the network call is successful and the UI displays a success alert.

## Checklist:


edewit (commented):
Do we really need to fill out this email address all the time? The old console seemed to take one of the form values? What also seems strange is that you need to hit save before you can click test connection, shouldn't it do that automatically?

mfrances17 (changes requested):
I was finally able to get the email server up, but when i enter a host, port, and email in the new admin UI, it fails with 500 errors. When i enter the same information in the old admin UI running on 8180, the email is sent successfully.

edewit (approved):

mfrances17 (changes requested):
I followed the verification steps and it still failed to send the email. HOWEVER, I saw on the keycloak server an EmailException that there wasn't a valid address. So I looked and the From field is marked as required (with a *) but you can still click Test connection with that field being empty. Once I entered a valid email address, it succeeded.

So, long story short - add some validation for a valid email address in the From field (or even just checking that it is not empty, but checking for valid email would be a plus) so the Test connection cannot be clicked unless it is poplulated.

mfrances17 (changes requested):
Apologies... just noticed another issue. When the email tab is first loaded, under Connection & Authentication, Authentication defaults to Disabled but the Username and Password fields are displayed by default (and they have a required *, which makes it a more obvious bug). If you Enable then Disable authentication, the fields are hidden, as expected. So these fields needs to do the same check on load.

mfrances17 (changes requested):
Email test still fails... change it to this and it will pass:
```
  it("Go to email tab", () => {
    sidebarPage.goToRealmSettings();
    cy.getId("rs-email-tab").click();

    cy.wait(1000);

    realmSettingsPage.addSenderEmail("example@example.com");

    realmSettingsPage.toggleCheck(realmSettingsPage.enableSslCheck);
    realmSettingsPage.toggleCheck(realmSettingsPage.enableStartTlsCheck);

    realmSettingsPage.save(realmSettingsPage.emailSaveBtn);

    realmSettingsPage.fillHostField("localhost");
    cy.getId(realmSettingsPage.testConnectionButton).click();

    masthead.checkNotificationMessage("Error! Failed to send email.");
  });
```

mfrances17 (approved):
LGTM, well done

Issue #667, opened 2021-06-04
[UX Review] Client - Service account roles


xianli123:
hi @edewit, in this issue, I will leave some comments about the "Service account roles". Please check the comments below.

**[Table]**

> <img width="1151" alt="Screen Shot 2021-06-04 at 2 07 00 PM" src="https://user-images.githubusercontent.com/53931519/120754471-4d92a600-c53f-11eb-88a7-f5cccf779c14.png">
1. There is a question mark. After clicking the question mark, there will be a piece of help info: "Disabled clients cannot initiate a login or have obtained access tokens."   
2. Please change "Service account" to "Service account roles".    
3. There should be a checkbox in front of each role and the checkboxes of Inherited roles are disabled.    
4. Each direct role is followed by a kebab. Click the kebab icon and then click the "Unassign" to unassign the role. Please note that if the roles are inherited roles, the kebab should be disabled.

xianli123:
**[Assign modal]**

> <img width="1145" alt="Screen Shot 2021-06-04 at 2 34 49 PM" src="https://user-images.githubusercontent.com/53931519/120757470-2ccc4f80-c543-11eb-8596-01d466d0c728.png">

1. Put the filter bar and the search bar together. If possible, we can fix the top filter and search bar. 
2. We can put a check box here. 

Please refer to the screenshot below.

<img width="656" alt="Screen Shot 2021-06-04 at 2 44 28 PM" src="https://user-images.githubusercontent.com/53931519/120757705-77e66280-c543-11eb-8613-28da096539c8.png">

edewit:
> There should be a checkbox in front of each role and the checkboxes of Inherited roles are disabled.
These only appear when you check "Hide inherited roles" as the inherited roles can not be removed.

xianli123:
I think we should show the check box even though "Hide inherited role" is not marked. We can use the disabled checkbox to indicate that this role is an inherited role as shown in the screenshot below. What do you think?

> <img width="517" alt="Screen Shot 2021-06-10 at 5 17 36 PM" src="https://user-images.githubusercontent.com/53931519/121499876-5e945900-ca10-11eb-8799-2d115a4e1335.png">

PR #708, opened 2021-06-16
fixing issue

edewit:
fixes: #667

### DIFF
--- a/src/client-scopes/form/ClientScopeForm.tsx
+++ b/src/client-scopes/form/ClientScopeForm.tsx
@@ -19,7 +19,11 @@ import { convertFormValuesToObject } from "../../util";
 import { MapperList } from "../details/MapperList";
 import { ScopeForm } from "../details/ScopeForm";
 import { useConfirmDialog } from "../../components/confirm-dialog/ConfirmDialog";
-import { RoleMapping, Row } from "../../components/role-mapping/RoleMapping";
+import {
+  mapRoles,
+  RoleMapping,
+  Row,
+} from "../../components/role-mapping/RoleMapping";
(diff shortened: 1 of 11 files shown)


mfrances17 (changes requested):
Is this meant as a partial fix to #667? If so, maybe it should be renamed. I don't see fixes for 1 (question mark/help), 2 (rename tab), or 4 (add kabobs), I only see a fix for 3 which is to deal with checkboxes and inherited roles, which works in my testing.

Let me know if only number 3 was supposed to be covered in this review, and i'll approve. Also, one thing i'm not sure can be changed - if you select the select all rows checkbox aka the one to the left of the Name column header, the checkboxes for disabled roles (the inherited roles) are checked. But i think this is a visual thing only as i had 6 checked but the count in the confirm modal was 5. It may just be an unfortunate visual anomaly, not sure if it can be easily fixed or not, or if it's worth fixing because again i don't think the selection is actually happening.

edewit:
ehh... #708 is the number of this PR it fixes #667

mfrances17:
Sorry that was a typo on my part... I was referencing issue https://github.com/keycloak/keycloak-admin-ui/issues/667. I'll edit the original.

edewit:
Added help on Enable Switch: 
image

edewit:
@mfrances17 this PR now fixes all of the issues in #667

mfrances17:
@edewit looks good, all issues are addressed. great job. i'm approving as-is. for the future we should think about creating a keycloak component that has the enabled and question mark help built-in, and also the ? should be closer to the Enabled text since it almost looks like it belongs with the Action menu since it is closer to that, and it should be a shade of gray like the other help toggles instead of black. But i think those should be handled as separate issues.

PR #709, opened 2021-06-16
fixing issue

edewit:
fixing: #680

### DIFF
--- a/src/components/group/GroupPath.tsx
+++ b/src/components/group/GroupPath.tsx
@@ -1,0 +1,48 @@
+import React from "react";
+import { Tooltip } from "@patternfly/react-core";
+import type { TableTextProps } from "@patternfly/react-table";
+
+import type GroupRepresentation from "keycloak-admin/lib/defs/groupRepresentation";
+
+type GroupPathProps = TableTextProps & {
+  group: GroupRepresentation;
+};
+
+const MAX_LENGTH = 20;
+const PART = 10;
+
+const truncatePath = (path?: string) => {
+  if (path && path.length >= MAX_LENGTH) {
+    return (
+      path.substr(0, PART) +
+      "..." +
+      path.substr(path.length - PART, path.length)
+    );
+  }
+  return path;
+};
+
+export const GroupPath = ({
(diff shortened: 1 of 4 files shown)


mfrances17 (changes requested):
Great work... was comparing side-by-side to issue #680 when i tested:

Issues 1 and 2 - now work as expected.
Issue 3 - regarding dealing with a long string in the membership column, I see an extra prepended '/' at the beginning of the tooltip string as well as it not matching up to what is displayed (there is an appended /test-erik at the end) - see cap:
membership-popup
Issue 4 - i believe there was an ask to add a kabob to all rows, as in other data tables, but i still don't see them when testing

edewit:
@mfrances17 how did you get this group name, I can't seem to get it to look like that

mfrances17:
These are the steps i think i used to test it last week:

1. Groups > Create group and named it a long name e.g. ten/twenty/thirty/forty/fifty/sixty/seventy/eighty
2. Users > Add user and named it test-user-1.
3. Groups > select group in step 1.
4. Members tab > click Add member, select username from step 2, and click Add. 

As of this morning's testing, I'm not able to reproduce the issue. The Membership column does seem to be displaying the tooltip correctly for the truncated group. I'm not sure what I did differently, if anything. Also, the double // at the start of the path is due to adding one at the beginning of your group name, if you don't name it with a leading / then it appears correctly.

mfrances17:
@edewit Actually, I think I figured out how to reproduce the discrepancy, took me a while:

Do steps 1-4 above and then:

5. Groups > select Group in step 1.
6. Child groups > Create group and name it my-child-group.
7. Click my-child-group > select Members tab > click Add member.
8. Select user in step 2 (test-user-1) and click Add.

Notice that the child group is not listed in the truncated Membership column but is listed in the hover popup:
popup-not-same

edewit:
right this is due to the fact that it are 2 groups listed there, I added a "," to make that more clear:
image

mfrances17 (approved):
The separation makes it much, much clearer. LGTM!

PR #723, opened 2021-06-23
fixed default value warning

edewit:
This PR fixes the default value warnings that coccured in the console on the realm settings


src/realm-settings/ThemesTab.tsx
```diff
@@ -244,7 +244,7 @@ export const RealmSettingsThemesTab = ({
             <Controller
               name="internationalizationEnabled"
               control={control}
-              defaultValue={internationalizationEnabled}
+              defaultValue={false}
               render={({ onChange, value }) => (
```
mfrances17, line 248: LGTM, default value warnings no longer appear

mfrances17 (approved):

PR #725, opened 2021-06-25
updated admin client with new endpoints

edewit:
the new admin client supports these endpoints now
 * test smtp settings - email
 * localisation support

To test see email setting under realm and localisation


jenny-s51 (approved):
These look and work great @edewit - thank you for getting these new endpoints into the server

PR #735, opened 2021-06-28
removed member count column

edewit:
The member count was a very intensive query at best and not do-able for other storage providers. So we decided to remove it.


mfrances17 (changes requested):
Code looks good, in testing the Members column is now gone as expected. One thing I noticed - you still want to make sure the kabob in each row is right-aligned, as in other data tables:

kabob
:

edewit:
Right align:
image

mfrances17 (approved):
LGTM!

PR #741, opened 2021-06-30
Fixed these 2 issues:

edewit:
* removed double loading indicator
* fix order of alerts where the newest is on the top


mfrances17 (changes requested):
The new implementation of the alerts is not working as expected. The alert notifications do time out and go away as they used to, but now when you do anything, the entire list of notifications reappear. For example, on an empty screen e.g. no notifications present, when i  add a new group, instead of 'Group created', i received a stack of 5 notifications for the one action (it looks like the new notifications is moved to the top but the old notifications are displayed again after they've timed out long ago). The old notifications should not be re-displayed.

mfrances17 (approved):
Works great now - LGTM!

jenny-s51 (approved):
👍

yih-wang (approved):
LGTM

PR #786, opened 2021-07-05
sessions: added sessions type dropdown

agagancarczyk:
## Checklist:


agagancarczyk:
Implemented as per attached 
Screenshot 2021-07-05 at 16 57 08

src/sessions/SessionsSection.tsx
```diff
@@ -53,6 +93,25 @@ export const SessionsSection = () => {
...
+              onToggle={() => setFilterDropdownOpen(!filterDropdownOpen)}
+              toggleIcon={<FilterIcon />}
+              onSelect={(_, value) => {
+                setFilterType(value as string);
+                refresh();
+                setFilterDropdownOpen(false);
+              }}
+              selections={filterType}
+            >
+              {options}
```
jonkoops, line 109: Since `options` is never used anywhere else, perhaps it can be in-lined here? This would also eliminate the need to give these children a `key` prop.

src/sessions/SessionsSection.tsx
```diff
@@ -21,7 +29,15 @@ const Clients = (row: UserSessionRepresentation) => {
 };
 
 export const SessionsSection = () => {
+  const { t } = useTranslation("sessions");
   const adminClient = useAdminClient();
+  const [filterDropdownOpen, setFilterDropdownOpen] = useState(false);
+  const [filterType, setFilterType] = useState("All session types");
+  const [key, setKey] = useState(0);
```
jonkoops, line 36: `key` is never used here, meaning that if `refresh()` is triggered, nothing will happen.

src/sessions/SessionsSection.tsx
```diff
@@ -21,7 +29,15 @@ const Clients = (row: UserSessionRepresentation) => {
 };
 
 export const SessionsSection = () => {
+  const { t } = useTranslation("sessions");
   const adminClient = useAdminClient();
+  const [filterDropdownOpen, setFilterDropdownOpen] = useState(false);
+  const [filterType, setFilterType] = useState("All session types");
```
jonkoops, line 35: Should this initial value not be replaced with `t("sessions:All session types")`? As the initial value might be mismatched if the user is using a different locale.

```suggestion
  const [filterType, setFilterType] = useState(t("sessions:All session types"));
```

src/sessions/SessionsSection.tsx
```diff
@@ -53,6 +98,25 @@ export const SessionsSection = () => {
           loader={loader}
           ariaLabelKey="session:title"
           searchPlaceholderKey="sessions:searchForSession"
+          searchTypeComponent={
+            <Select
+              width={200}
```
jonkoops, line 95: I am not aware if we are optimizing the layout for mobile as well, but this size squishes the other controls on smaller viewports.

image

src/sessions/SessionsSection.tsx
```diff
@@ -53,6 +90,25 @@ export const SessionsSection = () => {
...
+              data-testid="filter-session-type-select"
+              isOpen={filterDropdownOpen}
+              className="filter-session-type-select"
+              variant={SelectVariant.single}
+              onToggle={() => setFilterDropdownOpen(!filterDropdownOpen)}
```
jonkoops, line 100: It looks like `onToggle` provides us with the [new expanded state](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/Select.tsx#L110), so we do not have to determine it ourselves.
```suggestion
              onToggle={(isExpanded) => setFilterDropdownOpen(isExpanded)}
```

src/sessions/SessionsSection.tsx
```diff
@@ -53,6 +90,25 @@ export const SessionsSection = () => {
...
+              toggleIcon={<FilterIcon />}
+              onSelect={(_, value) => {
+                setFilterType(value as string);
```
jonkoops, line 103: Instead of type-casting value here we can also call `toString()` instead, as this is a method that any `SelectOptionObject` [must implement](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/SelectOption.tsx#L12).
```suggestion
                setFilterType(value.toString());
```

jonkoops (dismissed):

src/sessions/SessionsSection.tsx
agagancarczyk: I have just checked other screens and we don't seem to optimize the layout for mobile. Perhaps we can discuss this with the team.

jonkoops: Yeah, I am curious what the team's opinion is. If we are not going for responsive, we might consider locking the min viewport size.

src/sessions/SessionsSection.tsx
agagancarczyk: I added the key back in as it's going to be needed to fully implement this filter. I followed the logic in KeysListTab for consistency but it can be refactored once fully implemented.

src/sessions/SessionsSection.tsx
agagancarczyk: Added the key back here.

src/sessions/SessionsSection.tsx
agagancarczyk: This is not fully implemented yet, and I followed the logic in KeysListTab, but I'm happy to change to useState(t("sessions:All session types")); for now.

src/sessions/SessionsSection.tsx
agagancarczyk: That's a good call!

src/sessions/SessionsSection.tsx
agagancarczyk: Done

src/sessions/SessionsSection.tsx
agagancarczyk: Yes, we could do that.

edewit (approved):
:+1:

src/sessions/SessionsSection.tsx
agagancarczyk: Done

src/sessions/SessionsSection.tsx
```diff
@@ -21,7 +29,17 @@ const Clients = (row: UserSessionRepresentation) => {
 };
 
 export const SessionsSection = () => {
+  const { t } = useTranslation("sessions");
   const adminClient = useAdminClient();
+  const [filterDropdownOpen, setFilterDropdownOpen] = useState(false);
+  const [filterType, setFilterType] = useState(
+    t("sessionsType.allSessions").toString()
+  );
```
jonkoops, line 37: There is no need to call to `toString()` here if the value is returned from a factory function.
```suggestion
  const [filterType, setFilterType] = useState(() => t("sessionsType.allSessions"));
```

src/sessions/SessionsSection.tsx
```diff
@@ -50,9 +68,54 @@ export const SessionsSection = () => {
...
+              toggleIcon={<FilterIcon />}
+              onSelect={(_, value) => {
+                setFilterType(value.toString());
+                refresh();
+                setFilterDropdownOpen(false);
+              }}
+              selections={filterType}
+            >
+              <SelectOption
+                key={0}
```
jonkoops, line 92: These `key` props can be removed from the children as they are now static in order.

Issue #834, opened 2021-07-12
Route feedback on AESProviderDetails PR


jenny-s51:
This issue will address @jonkoops 's comment: https://github.com/keycloak/keycloak-admin-ui/pull/729#discussion_r667804207

jenny-s51:
Closing - will be resolved by @jonkoops 's work on #878

PR #835, opened 2021-07-12
Route changes

jenny-s51:
## Motivation

#834 

## Checklist:

- [ ] Type checking has been performed via 'npm run check-types'

### DIFF
--- a/src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
+++ b/src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
@@ -24,6 +24,7 @@ import { FormAccess } from "../../../components/form-access/FormAccess";
 import { ViewHeader } from "../../../components/view-header/ViewHeader";
 import { convertToFormValues } from "../../../util";
 import { useAlerts } from "../../../components/alert/Alerts";
+import { AESProviderTypeRoute } from "./AESProviderTypeRoute";
 
 type AESGeneratedFormProps = {
   handleModalToggle?: () => void;
@@ -346,9 +347,8 @@ export const AESGeneratedForm = ({
 
 export const AESGeneratedSettings = () => {
   const { t } = useTranslation("realm-settings");
-  const providerId = useRouteMatch<MatchParams>(
-    "/:realm/realm-settings/keys/:id?/:providerType?/settings"
-  )?.params.providerType;
+  const providerId = useRouteMatch<MatchParams>(AESProviderTypeRoute.path)
+    ?.params.providerType;
   return (
     <>
       <ViewHeader titleKey={t("editProvider")} subKey={providerId} />
--- a/src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
+++ b/src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
@@ -51,9 +51,7 @@ export const AESGeneratedForm = ({
 
   const { id } = useParams<{ id: string }>();
 
-  const providerId =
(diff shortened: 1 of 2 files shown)


jonkoops:
@jenny-s51 could you rebase this PR on `master`? It looks like there are some merge conflicts.

src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+          {t("common:cancel")}
+        </Button>
+      </ActionGroup>
+    </FormAccess>
+  );
+};
+
+export const AESGeneratedSettings = () => {
+  const { t } = useTranslation("realm-settings");
+  const providerId = useRouteMatch<MatchParams>(AESProviderTypeRoute.path)
```
jonkoops, line 350: Since this match is using the path from `AESProviderTypeRoute` we should also use the corresponding type from that file.

```suggestion
  const providerId = useRouteMatch<AESProviderTypeRouteParams>(AESProviderTypeRoute.path)
```

But since this should be the only route that is triggered anyways I believe we can also use `useParams` here.
```suggestion
  const params = useParams<AESProviderTypeRouteParams>()
```

src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+        }
+      >
+        <Controller
+          name="config.secretSize"
+          control={form.control}
+          defaultValue={["16"]}
+          render={({ onChange, value }) => (
+            <Select
+              toggleId="kc-aes-keysize"
+              onToggle={() => setIsKeySizeDropdownOpen(!isKeySizeDropdownOpen)}
```
jonkoops, line 305: `onToggle` provides state so we do not have to manage it ourselves:

```suggestion
              onToggle={(isExpanded) => setIsKeySizeDropdownOpen(isExpanded)}
```

src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+            forID="kc-active"
+          />
+        }
+      >
+        <Controller
+          name="config.active"
+          control={form.control}
+          defaultValue={["true"]}
+          render={({ onChange, value }) => {
+            return (
```
jonkoops, line 268: This can be an arrow function so there is no need for a `return` statement.

src/realm-settings/key-providers/aes-generated/AESGeneratedForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+        await adminClient.components.update(
+          { id },
+          {
+            ...component,
+            parentId: component.parentId,
+            providerId: providerType,
+            providerType: "org.keycloak.keys.KeyProvider",
+          }
+        );
+        addAlert(t("saveProviderSuccess"), AlertVariant.success);
```
jonkoops, line 70: From what I can tell this code is never executed as it's on a route that has no `id` parameter.

jonkoops (changes requested):

jenny-s51:
@jonkoops Thank you for your helpful comments and feedback! Rebased - looks like the cyclical dependency (AESProviderTypeRoute is used in AESGeneratedSettings and vice versa) is still resulting in an error and the console is currently not loading properly if we try routing this way. What do you think would be the best direction to go here?

jonkoops:
Closing this, as we want to come up with a better solution for all route related code.

PR #839, opened 2021-07-13
Upgrade to Snowpack 3

jonkoops:

PR #840, opened 2021-07-13
Run as theme

ssilvert:
## Motivation
Allow new admin console to run as a theme on Keycloak server.

## Brief Description
When you build the app, it will no longer go into `/build`.  Instead, it will be built into `/build/src/main/resources/admin/resources`.  The reason for this change is that we now have a Maven `pom.xml` whose home is `/build`.  This Maven build will facilitate integration in the main Keycloak server build.

## Verification Steps
_Note: The steps below are for verification of Keycloak server integration.  Everything should work normally in development mode._
1.  `npm run build`
2. `cd build`
3. `mvn clean install`
4. copy `/build/target/classes/admin` directory to `<keycloak install>/themes/keycloak.v2`.  Alternatively, you can unzip keycloak-`admin-ui-0.0.1-SNAPSHOT.jar` into that directory.
5. Until we make changes to the Keycloak server, you will need to edit `<keycloak install>/themes/keycloak.v2/admin/index.ftl`.  You will need to hard-code the value for loginRealm.  So most of the time, it will be sufficient to replace `"${loginRealm}"` with `"master"`.
6. Go to `Realm Settings --> Themes` and set Admin Console Theme to `keycloak.v2`.
7. Go directly to Keycloak server with a URL such as [http://localhost:8180/auth/admin/master/console/](http://localhost:8180/auth/admin/master/console/)
8. Step 7 gets you to the console logging into master realm.  But you can also act as an administrator who is restricted to a single realm.  For that, you will need to set the name of the single realm, such as `"test"` in step 5.  Then create a user with proper permissions for the realm and log in with a URL such as [http://localhost:8180/auth/admin/test/console/](http://localhost:8180/auth/admin/test/console/).  See [https://www.keycloak.org/docs/6.0/server_admin/#_per_realm_admin_permissions](https://www.keycloak.org/docs/6.0/server_admin/#_per_realm_admin_permissions)

_You can also test step 8 in development mode.  Follow the same instructions, except the URL will be [http://localhost:8080/?realm=test](http://localhost:8080/?realm=test)_

## Checklist:

- [X] Unit tests have been created/updated
- [X] Type checking has been performed via 'npm run check-types'

## Additional Notes
One of the notes in the README.md does not apply yet.  `-Dprofile.feature.newadmin=enabled` does not exist in the Keycloak server yet.  So you do not need to enable it.


.gitignore
```diff
@@ -128,8 +128,10 @@ sketch
 
 # snowpack
 web_modules/
-build/
-.build/
+build/src
+.build/src
```
jonkoops, line 132: As far as I can tell we never generate a `.build` directory, I think we can remove these entries.

build/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+&lt;head&gt;
+  &lt;script type="text/javascript"&gt;
+    var loginRealm = "${loginRealm}";
+    var authServerUrl = "${authServerUrl}";
+    var authUrl = "${authUrl}";
+    var consoleBaseUrl = "${consoleBaseUrl}";
+    var resourceUrl = "${resourceUrl}";
+    var masterRealm = "${masterRealm}";
+    var resourceVersion = "${resourceVersion}";
+  &lt;/script&gt;
```
jonkoops, line 88: Instead of putting a bunch of vars on the global scope I think it's better if we create a file in the application which can read these variables as JSON by retrieving the element from the DOM. I wrote a blog post on how to do this to [Dockerize front-end applications](https://dev.to/jonkoops/building-portable-front-end-applications-with-docker-2cpm), from which I think we can take some pointers.

build/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+            <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <properties></properties>
+
+    <dependencyManagement>
+        <dependencies></dependencies>
+    </dependencyManagement>
```
jonkoops, line 38: Do we need these elements if they are empty?

build/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+                    </execution>
+                </executions>
+                <configuration>
+                    <file>target/classes/admin/resources/index.html</file>
+                    <outputFile>target/classes/admin/index.ftl</outputFile>
+                    <regex>false</regex>
+                    <replacements>
+                        <replacement>
+                            <token>src="/_dist_</token>
+                            <value>src="${resourceUrl}/_dist_</value>
```
jonkoops, line 70: Be aware that in the Snowpack 3 PR (https://github.com/keycloak/keycloak-admin-ui/pull/839) this directory no longer exists, and all is served from the directory root.

src/context/whoami/WhoAmI.tsx
```diff
@@ -4,12 +4,10 @@ import i18n from "../../i18n";
...
 import type { AccessType } from "keycloak-admin/lib/defs/whoAmIRepresentation";
 import { useAdminClient, useFetch } from "../auth/AdminClient";
+import { homeRealm } from "../../util";
 
 export class WhoAmI {
-  constructor(
-    private homeRealm?: string | undefined,
-    private me?: WhoAmIRepresentation | undefined
-  ) {
+  constructor(private me?: WhoAmIRepresentation | undefined) {
```
jonkoops, line 10: `me` is already marked as optional, so it's type does not have to be overloaded into `undefined`.
```suggestion
  constructor(private me?: WhoAmIRepresentation) {
```

jonkoops (changes requested):

build/pom.xml
ssilvert: Note that Maven is creating index.ftl from index.html.

build/pom.xml
ssilvert: Nope.  Need to remove those.

build/pom.xml
ssilvert: Yes.  We will need to update pom.xml in your PR.

build/pom.xml
edewit: Instead of having all these `var`s as separate global variables we could create one and have these as properties:

```javascript
const settings = {
    loginRealm: "${loginRealm}",
    authServerUrl: "${authServerUrl}",
    authUrl: "${authUrl}",
    consoleBaseUrl: "${consoleBaseUrl}",
    resourceUrl: "${resourceUrl}",
    masterRealm: "${masterRealm}",
    resourceVersion: "${resourceVersion}",
}
```
this would be a slight improvement, but I must say I like the suggestion of @jonkoops with the json. That is so much cleaner then a global variable

build/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+                        <replacement>
+                            <token>href="./</token>
+                            <value>href="${resourceUrl}/</value>
+                        </replacement>
+                        <replacement>
+                            <token>&lt;head&gt;</token>
+                            <value>
+&lt;head&gt;
+  &lt;script type="text/javascript"&gt;
+    var loginRealm = "${loginRealm}";
```
edewit, line 75: This seems to not be in the freemarker config

build/pom.xml
ssilvert: I agree Jon's way would be better, but it's not possible.  Those values have to be embedded into index.ftl so they can be passed in by Freemarker.

build/pom.xml
ssilvert: @edewit Right.  I've added it on the Keycloak side.  For now, you need to hard-code this by editing index.ftl.  See instructions above.

build/pom.xml
edewit: @ssilvert yes Jon's way would be possible it would look like this in `index.ftl`:

```html
<script id="settings" type="application/json">
   {
     loginRealm: "${loginRealm}",
     authServerUrl: "${authServerUrl}",
     authUrl: "${authUrl}",
     consoleBaseUrl: "${consoleBaseUrl}",
     resourceUrl: "${resourceUrl}",
     masterRealm: "${masterRealm}",
     resourceVersion: "${resourceVersion}",
  }
</script>
```
And then you could use it like this:
```typescript
getSettings() {
  const element = document.getElementById('settings');

  try {
    return JSON.parse(element.textContent)
  } catch (error) {
    console.error('Unable to parse settings!');
  }

  return {}
}
```
By doing it like this there aren't any global variables

ssilvert: @edewit @jonkoops Yes. I misunderstood what he was wanting to do.  I like that a lot.

jonkoops (approved):

PR #842, opened 2021-07-14
Remove unused Keycloak environment variables

jonkoops:


jonkoops:
@edewit wants to take a look if we can get the Docker setup working, because at the moment it is not functional at all.

PR #844, opened 2021-07-14
Use a Snowpack plugin to check TypeScript types

jonkoops:
## Motivation
This allows types to be checked during development without running `npm run check-types` separately and reduces the need to run this task on the CI.


edewit (approved):
Why didn't we have this before :slightly_smiling_face:

jonkoops:
> Why didn't we have this before slightly_smiling_face

This didn't exist when the project was started, thankfully now it does!

PR #851, opened 2021-07-14
Parse environment information as JSON

jonkoops:
## Motivation
Prevents having to use global variables for environment variables injected by running the application as a Keycloak theme.

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -68,18 +68,22 @@
                             <value>href="${resourceUrl}/</value>
                         </replacement>
                         <replacement>
-                            <token>&lt;head&gt;</token>
-                            <value>
-&lt;head&gt;
-  &lt;script type="text/javascript"&gt;
-    var loginRealm = "${loginRealm}";
-    var authServerUrl = "${authServerUrl}";
-    var authUrl = "${authUrl}";
-    var consoleBaseUrl = "${consoleBaseUrl}";
-    var resourceUrl = "${resourceUrl}";
-    var masterRealm = "${masterRealm}";
-    var resourceVersion = "${resourceVersion}";
-  &lt;/script&gt;
+                            <token><![CDATA[</body>]]></token>
+                            <value xml:space="preserve">
+<![CDATA[
+  <script id="environment" type="application/json">
+    {
+      "loginRealm": "${loginRealm}",
+      "authServerUrl": "${authServerUrl}",
+      "authUrl": "${authUrl}",
+      "consoleBaseUrl": "${consoleBaseUrl}",
+      "resourceUrl": "${resourceUrl}",
+      "masterRealm": "${masterRealm}",
+      "resourceVersion": "${resourceVersion}"
+    }
+  </script>
+</body>
+]]>
 </value>
                         </replacement>
                     </replacements>
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -79,7 +79,8 @@
       "consoleBaseUrl": "${consoleBaseUrl}",
       "resourceUrl": "${resourceUrl}",
       "masterRealm": "${masterRealm}",
-      "resourceVersion": "${resourceVersion}"
+      "resourceVersion": "${resourceVersion}",
+      "isRunningAsTheme": true
     }
   </script>
 </body>
--- a/src/context/whoami/WhoAmI.tsx
+++ b/src/context/whoami/WhoAmI.tsx
@@ -4,7 +4,6 @@ import i18n from "../../i18n";
 import type WhoAmIRepresentation from "keycloak-admin/lib/defs/whoAmIRepresentation";
 import type { AccessType } from "keycloak-admin/lib/defs/whoAmIRepresentation";
 import { useAdminClient, useFetch } from "../auth/AdminClient";
-import { homeRealm } from "../../util";
 
 export class WhoAmI {
   constructor(private me?: WhoAmIRepresentation) {
@@ -25,13 +24,6 @@ export class WhoAmI {
     if (this.me === undefined) return "";
 
     return this.me.userId;
-  }
-
-  /**
-   * Return the realm I am signed in to.
-   */
-  public getHomeRealm(): string {
-    return homeRealm();
   }
 
   public canCreateRealm(): boolean {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -5,6 +5,7 @@ export type Environment = {
   authServerUrl: string;
   /** The URL to the path of the auth server where client requests can be sent. */
   authUrl: string;
+  /** The URL to the base of the admin console. */
   consoleBaseUrl: string;
   resourceUrl: string;
   /** The name of the master realm. */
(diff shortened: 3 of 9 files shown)


src/environment.ts
```diff
@@ -0,0 +1,66 @@
...
+  /** The URL to the path of the auth server where client requests can be sent. */
+  authUrl: string;
+  consoleBaseUrl: string;
+  resourceUrl: string;
+  /** The name of the master realm. */
+  masterRealm: string;
+  /** The version number of the auth sever (e.g. 14.0.0). */
+  resourceVersion: string;
+  /** Indicates if the application is running as a Keycloak theme. */
+  isRunningAsTheme: boolean;
```
jonkoops, line 15: @ssilvert Can you tell me what the purpose is of the fields I have not commented so that I can add that information. Also let me know if the comments I have are correct.

src/environment.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// TODO: See if we can drop this by switching away from hash based routing.
```
jonkoops, line 18: Is this something we can resolve, or do we want to keep this around for development purposes?

src/environment.ts
```diff
@@ -0,0 +1,66 @@
...
+// The default environment, used during development.
+const defaultEnvironment: Environment = {
+  loginRealm: realm,
+  authServerUrl: "TODO",
+  authUrl: "http://localhost:8180/auth",
+  consoleBaseUrl: "TODO",
+  resourceUrl: ".",
+  masterRealm: "master",
+  resourceVersion: "TODO",
+  isRunningAsTheme: false,
```
jonkoops, line 31: @ssilvert Can you tell me if these are the correct defaults for development, and what the values of the missing ones should be?

src/environment.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  // Attempt to parse the contents as JSON and return its value.
+  try {
+    return JSON.parse(element.textContent);
+  } catch (error) {
+    console.error("Unable to parse environment variables.");
+  }
+
+  // Otherwise, return an empty record.
+  return {};
```
edewit, line 65: We can simplify this a little bit and always print the error:
```suggestion
  const element = document.getElementById(ENV_ELEMENT_ID);

  try {
    return JSON.parse(element?.textContent || "");
  } catch (error) {
    console.error("Unable to parse environment variables.");
  }

  return {};
```

jonkoops: If we do that the error would also occur if you are developing locally since the element for the environment would not be present in the DOM.

src/environment.ts
ssilvert: This should clarify.  These are the values that are passed in the old console.  We don't use all those values yet, but I thought we should go ahead and add them to the environment.  I believe resourceVersion is some sort of hash of the version string.

        var authServerUrl = 'http://localhost:8180/auth';
        var authUrl = 'http://localhost:8180/auth';
        var consoleBaseUrl = '/auth/admin/master/console/';
        var resourceUrl = '/auth/resources/z1vp3/admin/keycloak';
        var masterRealm = 'master';
        var resourceVersion = 'z1vp3';

src/environment.ts
ssilvert: Those look fine.  I don't think we are using the TODO's yet.

src/environment.ts
```diff
@@ -0,0 +1,66 @@
...
+  ...defaultEnvironment,
+  ...getInjectedEnvironment(),
+  isRunningAsTheme: document.getElementById(ENV_ELEMENT_ID) !== null,
+};
+
+export default environment;
+
+/**
+ * Extracts the environment variables that are passed if the application is running as a Keycloak theme.
+ * These variables are injected by Keycloak into the `index.html` as a script tag, the contents of which can be parsed as JSON.
```
ssilvert, line 47:
```suggestion
 * These variables are injected by Keycloak into the `index.ftl` as a script tag, the contents of which can be parsed as JSON.
```

build/pom.xml
```diff
@@ -68,18 +68,22 @@
...
+<![CDATA[
+  <script id="environment" type="application/json">
+    {
+      "loginRealm": "${loginRealm}",
+      "authServerUrl": "${authServerUrl}",
+      "authUrl": "${authUrl}",
+      "consoleBaseUrl": "${consoleBaseUrl}",
+      "resourceUrl": "${resourceUrl}",
+      "masterRealm": "${masterRealm}",
+      "resourceVersion": "${resourceVersion}"
```
ssilvert, line 82: Might be simpler to just hard code isRunningAsTheme here.

jonkoops: Nice one! I went ahead and made the change.

src/environment.ts
jonkoops: Ok, then I will just leave it like this for now.

src/environment.ts
jonkoops: Went ahead and documented this behaviour instead of having a todo

src/environment.ts
jonkoops: I've added these as defaults where it makes sense.

build/pom.xml
```diff
@@ -68,18 +68,23 @@
...
+  <script id="environment" type="application/json">
+    {
+      "loginRealm": "${loginRealm}",
+      "authServerUrl": "${authServerUrl}",
+      "authUrl": "${authUrl}",
+      "consoleBaseUrl": "${consoleBaseUrl}",
+      "resourceUrl": "${resourceUrl}",
+      "masterRealm": "${masterRealm}",
+      "resourceVersion": "${resourceVersion}",
+      "isRunningAsTheme": true
```
edewit, line 83: always true? I mean when one of these is set it means it's "running as a theme" right?

src/context/whoami/WhoAmI.tsx
```diff
@@ -27,13 +26,6 @@ export class WhoAmI {
     return this.me.userId;
   }
 
-  /**
-   * Return the realm I am signed in to.
-   */
-  public getHomeRealm(): string {
-    return homeRealm();
-  }
-
```
edewit, line 36: are you sure we don't need this anymore?

src/environment.ts
```diff
@@ -0,0 +1,64 @@
...
+  // If the element cannot be found, return an empty record.
+  if (!element?.textContent) {
+    return {};
+  }
+
+  // Attempt to parse the contents as JSON and return its value.
+  try {
+    return JSON.parse(element.textContent);
+  } catch (error) {
+    console.error("Unable to parse environment variables.");
```
edewit, line 59: guess if you have the "environment" element but it doesn't contain JSON we might not want to just log it and move on?

build/pom.xml
jonkoops: Yes, and in the case of the result from the Maven build this should always be `true`. During development this will not be injected, so it will take the value of the default environment (so `false`).

src/context/whoami/WhoAmI.tsx
jonkoops: Yes, this is only used in the RealmProvider, and there I just refer to the environment variable directly.

src/environment.ts
jonkoops: I have no strong opinion about this, but I think it's nice that we keep it around so that the user can be informed if something is wrong. I would maybe even take it one step further and just throw an exception directly.

edewit (approved):

PR #854, opened 2021-07-15
Place Maven build into dedicated directory

jonkoops:
## Motivation
Keeps the build directory clean and gives Java devs a single point of entry if they do not want to deal with pesky JavaScript tooling.


edewit (approved):
I like it, makes more sense then it living in a bit obscure `build` folder, let's wait what @ssilvert thinks though before merging

PR #856, opened 2021-07-15
Update development Keycloak server to version 14

jonkoops:
## Motivation
Allows us to run the Admin UI as a Keycloak theme locally by updating the latest version of Keycloak and enabling the required feature flags.


jonkoops:
This change seems to fail the Cypress tests, looks like upstream changes from Keycloak are responsible for this.

jonkoops:
Tests should be fixed from the change in #861

edewit (approved):

PR #859, opened 2021-07-15
Convert scripts to ESM

jonkoops:
## Motivation
More modern syntax and should prevent linting issues when editing these files.


import.mjs
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+import { readFile } from "node:fs/promises";
+import KcAdminClient from "keycloak-admin";
+
+const consoleClientConfig = JSON.parse(
+  await readFile(new URL("./security-admin-console-v2.json", import.meta.url))
+);
+
+const adminClient = new KcAdminClient.default({
```
jonkoops, line 9: @edewit Looks like the admin client is not yet shipped as ESM compatible. Might be something to look into.

mfrances17 (approved):
LGTM - well done, the export statements etc that were being flagged as pre-commit issues are no longer flagged by intellisense in VSCode, so this should be great going forward.

PR #861, opened 2021-07-15
Comment out localization tab.

ssilvert:
## Motivation
Tests for https://github.com/keycloak/keycloak-admin-ui/pull/856 are failing because you can't go to the Realm Settings for any realm except master.  The reason is that we are getting a NPE on WildFly 14+.  I believe that there is a bug in WildFly.  It is probably not initializing the localization when you create a new realm.

However, I see some possible issues with the implementation on our side as well.  Our localization tab assumes that you should always be able to create bundle keys and values for English.  I believe this should not be the case.

The error is actually caused when you try to make the REST call to admin/realms/mynewrealm/localization/en.  But the old console does not make this call until you explicitly create the locale for en.  So you don't see the error in the old console.

This PR should allow us to continue until we decide what to do.

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -52,7 +52,7 @@ describe("Realm settings", () => {
     return this;
   };
 
-  const addBundle = () => {
+  /*const addBundle = () => {
     const localizationUrl = `/auth/admin/realms/${realmName}/localization/en`;
     cy.intercept(localizationUrl).as("localizationFetch");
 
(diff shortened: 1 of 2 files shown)


jonkoops (approved):

mfrances17 (approved):
LGTM... verified that commenting out the L10N tab allowed me to get to realm settings for a new realm and i didn't get the blocking error.  

Recommend that you create an associated issue for the investigation/work to be completed as a result of this, so we can eventually get the L10N tab back.

ssilvert:
@jonkoops @mfrances17 Thanks guys.  Wasn't expecting anyone to see it before I got back home this evening and here it is with two reviews!

It did still have one failing test so I'm going to re-run.  Will create an issue to track.

edewit:
So the main issue with this seem that the new design removed the create localisation button:
image
I think we can solve it by introducing that

PR #866, opened 2021-07-16
Clean up and modernize ESlint configuration

jonkoops:
## Motivation
Our existing ESlint configuration has a lot of unrelated settings that can be provided by extending the base configs instead, this cleans up the configuration. We're now also extending a stricter configuration for TypeScript, which can catch possible type-errors early on (warnings only for now).


mfrances17 (commented):
Great stuff. I am absolutely thrilled to see that intellisense is finally working for this project in VSCode, i had attempted to implement this at one point but gave up. So i love this side of these changes; i can simply right-click lint issues in my IDE in real-time as opposed to running the scripts.

That said, the majority of files i opened have flagged issues, so the lint additions are a bit of a double-edge sword. No doubt we should fix them eventually, but this could get frustrating as we will need to address these pre-existing issues each time we add new code. Maybe there should be a sweep performed to fix all these flagged pre-existing issues with this PR so we will only get flagged on new violations aka the ones we would be introducing with our commits.

.eslintrc.js
```diff
@@ -0,0 +1,24 @@
...
+    "plugin:@typescript-eslint/recommended",
+    "plugin:prettier/recommended",
+  ],
+  settings: {
+    react: {
+      version: "detect",
+    },
+  },
+  rules: {
+    // TODO: These rules should be re-enabled and the resulting linting errors should be fixed.
```
mfrances17, line 19: this TODO comment can probably be removed unless you will be commenting out a selection of tests

.eslintrc.js
```diff
@@ -0,0 +1,24 @@
...
+    "@typescript-eslint/no-empty-function": "warn",
+    "@typescript-eslint/ban-types": "warn",
+    "@typescript-eslint/no-non-null-asserted-optional-chain": "warn",
+  },
```
mfrances17, line 23: will the source files with warnings, which i assume correlate to the 'Unchanged files with check annotations' (this feature is new to me, i haven't seen that before), be addressed as part of this PR? seems like we will be adding a lot of noise to future PRs for pre-existing violations

mfrances17 (commented):
Some general questions... i guess we should discuss as a group how we plan to address all of the pending issues that will be introduced by these lint additions

jenny-s51 (commented):
@jonkoops super cool - this looks like an excellent addition to the codebase and these new eslint warnings will be awesome to have. 

I agree with @mfrances17 - we should discuss as a team how we'll tackle the existing warnings once this gets merged and come up with a plan for how we want to resolve them. 

Otherwise looks great, thank you for making this update!

jonkoops:
@mfrances17 @jenny-s51 Yes, I agree. Let's discuss it during the team so we can see what the best approach is for this.

.eslintrc.js
jonkoops: These have been removed after restoring the original configuration.

.eslintrc.js
jonkoops: These warnings should no longer be showing up as we now extend the same configs as before.

jonkoops:
I've updated the PR with the suggestions from our team meeting, should be good for another round of reviews.

edewit (approved):

mfrances17 (commented):
LGTM, not seeing the same high numbers of warnings i was originally concerned with

PR #882, opened 2021-07-20
Introduce 'useRequiredContext' utility function

jonkoops:
Introduce a `useRequiredContext` function to assert the value from `useContext` is non-nullable, and throw an exception if it is not.

Other notable changes:
- All contexts now have a default value of `undefined`.
- Use dedicated hooks to retrieve values of contexts instead of direct access through `useContext` calls.
- All hooks calls proxy through `useRequiredContext` to ensure the context is always present.

### DIFF
--- a/src/utils/useRequiredContext.ts
+++ b/src/utils/useRequiredContext.ts
@@ -1,0 +1,29 @@
+import { useContext } from "react";
+import type { Context } from "react";
+
+/**
+ * Passes the call to `useContext` and throw an exception if the resolved value is either `null` or `undefined`.
+ * Can be used for contexts that are required and should always have a non nullable value.
+ *
+ * @param context The context to pass to `useContext`
+ * @returns
+ */
+export default function useRequiredContext<T>(
+  context: Context<T>
+): NonNullable<T> {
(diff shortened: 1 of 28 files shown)


src/utils/useRequiredContext.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (context.displayName) {
+    throw new Error(
+      `No provider found for the '${context.displayName}' context, make sure it is included in your component hierarchy.`
+    );
+  }
+
+  throw new Error(
+    `No provider found for an unknown context, make sure it is included in your component hierarchy.`
+  );
```
edewit, line 28:
```suggestion
 throw new Error(
      `No provider found for the '${context.displayName || "unknown"}' context, make sure it is included in your component hierarchy.`
    );
```

edewit (approved):
can't believe there was a `useRealm` function but it wasn't used.

src/utils/useRequiredContext.ts
jonkoops: Ah yes, that would simplify it a bit. I'll add this suggestion!

PR #883, opened 2021-07-20
Defer breadcrumbs translations on route definitions

jonkoops:
## Motivation
Allows us to access the route definitions without first having to call a function to pass the translations.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -74,13 +74,9 @@ export const App = () => {
             onReset={() => window.location.reload()}
           >
             <Switch>
-              {routes(() => {}).map((route, i) => (
+              {routes.map((route, i) => (
                 <Route
-                  exact={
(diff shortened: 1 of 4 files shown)


jenny-s51 (commented):
Overall looks great @jonkoops! A small thing I noticed - it looks like this PR introduces the id and provider type into breadcrumbs of key providers, though they should not be there (see #884).

I spent some time investigating the code and removing the id and provider type is not possible when modifying `EditProviderCrumb` directly. You also mentioned you'd like to open some more work on top of this so happy to approve if you were planning to address that in a following PR, thanks for making these changes 👍

jonkoops:
@jenny-s51 Turns out this is a bug in https://github.com/icd2k3/use-react-router-breadcrumbs/issues/30. I've pushed a commit to resolve the issue.

edewit (approved):

PR #887, opened 2021-07-21
Initial version of the authentication section

edewit:
Just wanted to add this because it's getting kinda big, it's not done yet, but it's a start

Missing are adding flows and executions.


package.json
```diff
@@ -28,6 +28,7 @@
     "@patternfly/react-core": "4.135.0",
     "@patternfly/react-icons": "4.11.0",
     "@patternfly/react-table": "4.29.0",
+    "dagre": "^0.8.5",
```
jonkoops, line 31: Do we want to introduce this dependency? From the [README of the project](https://github.com/dagrejs/dagre):

> # Important!
> 
> **This project does not have a maintainer or active project members. There won’t be any support or attention to pull requests. Please do not contact previous maintainers unless you are qualified and have the resources to make a serious commitment to fully take over ownership of the project.**

src/authentication/AuthenticationSection.tsx
```diff
@@ -198,6 +213,17 @@ export const AuthenticationSection = () => {
               loader={loader}
               ariaLabelKey="authentication:title"
               searchPlaceholderKey="authentication:searchForEvent"
+              toolbarItem={
+                <ToolbarItem>
+                  <Button
+                    onClick={() =>
+                      history.push(`/${realm}/authentication/create`)
+                    }
+                  >
```
jonkoops, line 222: This `Button` can be a React Router `Link` by providing the `component` prop.
```suggestion
                  <Button component={Link} to={`/${realm}/authentication/create`}>
```

package.json
edewit: well it works we could also not use it...

jonkoops: As discussed during the team meet, let's create an issue for this and keep the dependency for now.

src/authentication/AuthenticationSection.tsx
```diff
@@ -163,9 +168,22 @@ export const AuthenticationSection = () => {
...
     <>
-      <Link to={`${url}/${id}`} key={`link-{id}`}>
+      <Link
+        to={toFlow({
+          realm,
+          id: id!,
+          usedBy: usedBy.type || "notInUse",
+          builtIn: builtIn ? "builtIn" : undefined,
+        })}
+        key={`link-{id}`}
```
jonkoops, line 185: This template string is missing an expression.

```suggestion
        key={`link-${id}`}
```

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+        await adminClient.authenticationManagement.delExecution({ id });
+        const result =
+          await adminClient.authenticationManagement.addExecutionToFlow({
+            flow: change.parent?.displayName! || flow?.alias!,
+            provider: ex.providerId!,
+          });
+        id = result.id!;
+      }
+      const c = change as IndexChange;
+      const times = c.newIndex - c.oldIndex;
```
jonkoops, line 82: If I remove this cast the code still works, so I think it's not needed.

```suggestion
      const times = change.newIndex - change.oldIndex;
```

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+      for (let index = 0; index < Math.abs(times); index++) {
+        if (times > 0) {
+          await adminClient.authenticationManagement.lowerPriorityExecution({
+            id,
+          });
+        } else {
+          await adminClient.authenticationManagement.raisePriorityExecution({
+            id,
+          });
```
jonkoops, line 91: These requests will be waiting for each request to finish before firing another. I am not aware of how this API works, but if it doesn't care about the requests completing in order this could be parallelized using `Promise.all()`.

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+        }),
+        AlertVariant.danger
+      );
+    }
+  };
+
+  return (
+    <>
+      <ViewHeader
+        titleKey={toUpperCase(flow?.alias! || "")}
```
jonkoops, line 129: Since `alias` is already asserted as non-null the default fallback can be removed here.
```suggestion
        titleKey={toUpperCase(flow?.alias!)}
```

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+                  </Label>
+                ),
+                id: "builtIn",
+              }
+            : {},
+        ]}
+      />
+      <PageSection variant="light">
+        {executionList?.expandableList &&
+          executionList.expandableList.length > 0 && (
```
jonkoops, line 149: To reduce the size of this statement you could use `length` as an implicit truthy or falsy value by letting JavaScript cast it for you:
```suggestion
        {executionList?.expandableList?.length && (
```

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+            <Toolbar id="toolbar">
+              <ToolbarContent>
+                <ToggleGroup>
+                  <ToggleGroupItem
+                    icon={<TableIcon />}
+                    aria-label="copy icon button"
```
jonkoops, line 155: This label should be a translated human-readable string.

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    buttonId="third"
+                    isSelected={tableView}
+                    onChange={() => setTableView(true)}
+                  />
+                  <ToggleGroupItem
+                    icon={<i className="fas fa-project-diagram"></i>}
+                    aria-label="undo icon button"
```
jonkoops, line 162: This label should be a translated human-readable string.

src/authentication/FlowDetails.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    isSelected={!tableView}
+                    onChange={() => setTableView(false)}
+                  />
+                </ToggleGroup>
+              </ToolbarContent>
+            </Toolbar>
+          )}
+        {tableView &&
+          executionList?.expandableList &&
+          executionList.expandableList.length > 0 && (
```
jonkoops, line 173: To reduce the size of this statement you could use `length` as an implicit truthy or falsy value by letting JavaScript cast it for you:

```suggestion
        {tableView && executionList?.expandableList?.length && (
```

jonkoops (changes requested):

src/authentication/__tests__/ExecutionList.test.tsx
```diff
@@ -0,0 +1,150 @@
+import { ExecutionList, IndexChange, LevelChange } from "../execution-model";
```
jonkoops, line 1: We're using a mix of tests in `__tests__` directories and `*.test.* files living next to their source files. I personally prefer the latter if we need to settle on a convention here.

Also, tests :partying_face: Yay!

src/authentication/__tests__/ExecutionList.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    { id: "1", index: 0, level: 0 },
+    { id: "2", index: 1, level: 0 },
+    { id: "3", index: 0, level: 1 },
+    { id: "4", index: 1, level: 1 },
+    { id: "5", index: 0, level: 2 },
+    { id: "6", index: 1, level: 2 },
+    { id: "7", index: 2, level: 0 },
+  ]);
+
+  it("Move 1 down to the end", () => {
```
jonkoops, line 14: I prefer writing my test cases like a sentence, for example:

```
it moves 1 down to the end
```

or

```suggestion
  it("moves 1 down to the end", () => {
```

But this is a personal preference :man_shrugging:

src/realm-settings/messages.json
```diff
@@ -551,9 +551,5 @@
     "FAIL": "Fail import",
     "SKIP": "Skip",
     "OVERWRITE": "Overwrite"
-  },
-  "onDragStart": "Dragging started for item {{id}}",
-  "onDragMove": "Dragging item {{id}}",
-  "onDragCancel": "Dragging cancelled. List is unchanged.",
-  "onDragFinish": "Dragging finished {{list}}"
+  }
```
jonkoops, line 554: This change now only removes a comma. However it should be removed as we no longer support JSON translation files.

src/authentication/FlowDetails.tsx
edewit: it's only faster when one of these calls fails, even though this code looks synchrones it's not, it is syntactic sugar that makes it easier to write promises

package.json
edewit: created issue: https://github.com/keycloak/keycloak-admin-ui/issues/919

src/components/view-header/ViewHeader.tsx
```diff
@@ -40,11 +44,16 @@ export type ViewHeaderProps = {
   helpTextKey?: string;
 };
 
+export type ViewHeaderBadge = {
+  id?: string;
+  text?: string | ReactNode;
```
jonkoops, line 49: No need to provide `string` as a type here, as a string is already a type under `ReactNode`.
```suggestion
  text?: ReactNode;
```

src/authentication/FlowDetails.tsx
edewit: yeah that is a part of old code I think `LevelChange` has the same properties as `IndexChange` now

jonkoops (changes requested):

src/realm-settings/messages.json
jonkoops: Fixed this in #920

mfrances17 (commented):
Code looks good. I was able to create a new flow and dup existing ones. There's a lot of changes here and I realize this is a WIP, but it would be nice to have test steps within the PR so as to know what is expected to be working and how to test it appropriately. For example, I was not able to select a new or existing flow, i get 'Cannot read property at 'charAt' of undefined' errors. So I'm not sure if that should be working yet or not. Based on the code, I would expect more of that to be working.

src/authentication/form/FlowType.tsx
```diff
@@ -0,0 +1,62 @@
...
+      fieldId="flowType"
+    >
+      <Controller
+        name="providerId"
+        defaultValue={types[0]}
+        control={control}
+        render={({ onChange, value }) => (
+          <Select
+            toggleId="flowType"
+            onToggle={() => setOpen(!open)}
```
jonkoops, line 40: We don't need to manage this state ourselves. We can use the value provided by the select instead.
```suggestion
            onToggle={(isExpanded) => setOpen(isExpanded)}
```

src/authentication/form/FlowType.tsx
```diff
@@ -0,0 +1,62 @@
...
+            onSelect={(_, value) => {
+              onChange(value as string);
```
jonkoops, line 42: A `SelectOptionObject` need to implement `toString()` so we do not have to type-cast this value.
```suggestion
              onChange(value.toString());
```

src/realm-settings/UserRegistration.tsx
```diff
@@ -0,0 +1,65 @@
...
+          save={() => Promise.resolve()}
+          onHideRolesToggle={() => {}}
+        />
+      </Tab>
+      <Tab
+        id="evaluate"
+        eventKey={20}
+        title={<TabTitleText>{t("defaultGroups")}</TabTitleText>}
+      >
+        {/* <EvaluateScopes clientId={clientId} protocol={client!.protocol!} /> */}
```
jonkoops, line 61: Why is this code commented out?

src/realm-settings/UserRegistration.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from "react";
```
jonkoops, line 1: This file seems like a work-in-progress. I assume the rest of the functionality will be implemented in another PR?

src/realm-settings/UserRegistration.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const adminClient = useAdminClient();
+  const { realm: realmName } = useRealm();
+
+  const loader = async () => {
+    const realm = await adminClient.realms.findOne({ realm: realmName });
+    const assignedRoles = realm.defaultRoles!.map(toRole);
+
+    const clients = await adminClient.clients.find();
+    const clientRoles = (
+      await Promise.all(
```
jonkoops, line 24: I might be mistaken, but it looks like nothing is async here, perhaps we can get rid of the `async`/`await`?

src/authentication/components/FlowRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const FlowRow = ({
+  execution,
+  onRowClick,
+  onRowChange,
+}: FlowRowProps) => {
+  const { t } = useTranslation("authentication");
+  const hasSubList =
+    execution.executionList && execution.executionList.length > 0;
```
jonkoops, line 35: I believe you could fall back on JavaScript casting here.
```suggestion
  const hasSubList = !!execution.executionList.length;
```

src/authentication/components/FlowRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <>
+      <DataListItem
+        className="keycloak__authentication__flow-item"
+        id={execution.id}
+        isExpanded={!execution.isCollapsed}
+      >
+        <DataListItemRow
+          key={execution.id}
```
jonkoops, line 44: I don't think a `key` is required at this level.

src/authentication/components/FlowRow.tsx
```diff
@@ -0,0 +1,106 @@
...
+          id={execution.id}
+          className="keycloak__authentication__flow-row"
+          aria-level={execution.level}
+        >
+          <DataListControl>
+            <DataListDragButton
+              aria-label={t("common:reorder")}
+              aria-labelledby={execution.displayName}
```
jonkoops, line 52: I don't believe it's valid to use both of these attributes.

jonkoops (changes requested):

src/realm-settings/UserRegistration.tsx
edewit: yes it wasn't suppose to be here... I'll remove it

src/authentication/components/FlowRow.tsx
edewit: Are you sure? In the [example](https://www.patternfly.org/v4/components/data-list/#datalistdragbutton) they set both of these

jonkoops: Yes, the ARIA conventions used in Patternfly are wrong. According to [Google Developers](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships), which states the following:

> Importantly, `aria-labelledby` overrides **all** other name sources for an element. So, for example, if an element has both an `aria-labelledby` and an `aria-label`, or an `aria-labelledby` and a native HTML `label`, the `aria-labelledby` label always takes precedence.

src/authentication/components/FlowRequirementDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+    </SelectOption>
+  ));
+
+  return (
+    <>
+      {flow.requirementChoices && flow.requirementChoices?.length > 1 && (
+        <Select
+          className="keycloak__authentication__requirement-dropdown"
+          variant={SelectVariant.single}
+          onToggle={() => setOpen(!open)}
```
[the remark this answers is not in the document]
edewit: wouldn't this `isExpanded` always be `!open` and we still need `open` right?

src/authentication/components/FlowRequirementDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const options = flow.requirementChoices!.map((option, index) => (
+    <SelectOption key={index} value={option}>
+      {t(`requirements.${option}`)}
+    </SelectOption>
+  ));
+
+  return (
+    <>
+      {flow.requirementChoices && flow.requirementChoices?.length > 1 && (
```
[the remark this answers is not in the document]
edewit: Can't do that as I need it to be bigger then 1 not exactly 1. The expression `flow.requirementChoices?.length` will be true for a length of one. I can how ever remove the ? as it can't be undefined

edewit:
@mfrances17 I introduced a bug sorry. Added some tests and it's ready to review again

src/authentication/components/FlowRequirementDropdown.tsx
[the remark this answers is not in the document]
jonkoops: Ahhh, yes. That makes sense!

jonkoops (approved):

PR #888, opened 2021-07-21
Use new routing conventions for client routes

jonkoops:
## Motivation
Aligns the routing for client routes with the conventions as discussed in the team meeting.

### DIFF
--- a/src/clients/initial-access/CreateInitialAccessToken.tsx
+++ b/src/clients/initial-access/CreateInitialAccessToken.tsx
@@ -15,11 +15,12 @@ import { FormAccess } from "../../components/form-access/FormAccess";
 import { ViewHeader } from "../../components/view-header/ViewHeader";
 import { HelpItem } from "../../components/help-enabler/HelpItem";
 import { TimeSelector } from "../../components/time-selector/TimeSelector";
-import { useHistory } from "react-router-dom";
+import { Link, useHistory } from "react-router-dom";
 import { useRealm } from "../../context/realm-context/RealmContext";
 import { useAdminClient } from "../../context/auth/AdminClient";
 import { useAlerts } from "../../components/alert/Alerts";
 import { AccessTokenDialog } from "./AccessTokenDialog";
+import { toClients } from "../routes/Clients";
 
 export const CreateInitialAccessToken = () => {
   const { t } = useTranslation("clients");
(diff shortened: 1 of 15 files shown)


src/clients/initial-access/CreateInitialAccessToken.tsx
```diff
@@ -129,9 +130,9 @@ export const CreateInitialAccessToken = () => {
             <Button
               data-testid="cancel"
               variant="link"
-              onClick={() =>
-                history.push(`/${realm}/clients/initialAccessToken`)
-              }
+              component={Link}
+              // @ts-ignore
```
edewit, line 134: why?

src/clients/initial-access/InitialAccessTokenList.tsx
```diff
@@ -57,7 +56,8 @@ export const InitialAccessTokenList = () => {
         loader={loader}
         toolbarItem={
           <>
-            <Button onClick={() => history.push(`${url}/create`)}>
+            {/* @ts-ignore */}
```
edewit, line 59: this should be right?

edewit (changes requested):

src/clients/initial-access/CreateInitialAccessToken.tsx
jonkoops: Because the type definitions of Patternfly do not change the accepted props when the `component` props is passed, causing it to believe the `to` prop is invalid.

jonkoops: More info here: https://github.com/patternfly/patternfly-react/issues/5491

src/clients/initial-access/InitialAccessTokenList.tsx
jonkoops: See my other reply.

src/clients/initial-access/CreateInitialAccessToken.tsx
edewit: but this `@ts-ignore` is on the `to` attribute

edewit: why not do:

```
<Link to={toClients({ realm, tab: "initialAccessToken" })}>
     <Button
              data-testid="cancel"
              variant="link">
          {t("common:cancel")}
     </Button>
 </Link>
```

edewit (approved):

PR #890, opened 2021-07-21
Use new routing conventions for realm routes

jonkoops:
## Motivation
Aligns the routing for client realm with the conventions as discussed in the team meeting.


edewit (approved):
:tada: so much nicer

PR #895, opened 2021-07-22
Fixed docker build for testing purposes

edewit:
Now the docker build uses the "run as theme" option making the docker build super easy.

Also removed the loginRealm property as keycloak 14 will not yet send that. We can change it back when it's in a release keycloak version


Dockerfile
```diff
@@ -1,32 +1,11 @@
...
-
-ARG DEFAULT_KEYCLOAK_ENDPOINT
-ARG KEYCLOAK_ENDPOINT
+FROM maven:3.6.0-jdk-11-slim as builder
 
 WORKDIR /app
 COPY . .
 
-# replace Keycloak endpoints
-RUN sed -i "s/${DEFAULT_KEYCLOAK_ENDPOINT}/${KEYCLOAK_ENDPOINT}/g" import.mjs &&\
```
jonkoops, line 12: Nice cleanup!

jonkoops (approved):

keycloak-theme/pom.xml
```diff
@@ -137,7 +137,7 @@
 <![CDATA[
   <script id="environment" type="application/json">
     {
-      "loginRealm": "${loginRealm}",
+      "loginRealm": "master",
```
edewit, line 140: @ssilvert is this okay for now we can change it back when keycloak 15 is out

PR #896, opened 2021-07-22
Sessions(actions dropdown): Adds actions dropdown - "revocation" and "log out all sessions"

jenny-s51:
## Motivation

Towards https://issues.redhat.com/browse/APPDUX-1067

## Brief Description

Adds "revocation" functionality to Actions dropdown of sessions page

## Verification Steps

1. Go to Sessions
2. Click on "Actions" menu in the top right and select "Revocation"
3. Click "Set to now".
4. Sign back in
5. Repeat steps 1-2
6. Verify that "Not before" value is correct
7. Click "Clear"
8. Click on "Actions" menu in the top right and select "Revocation"
9. Verify that "Not before" value is set to none.

* Push functionality will be confirmed by @ssilvert per Zulip thread

## Checklist:

### DIFF
--- a/cypress/support/pages/admin_console/manage/sessions/SessionsPage.ts
+++ b/cypress/support/pages/admin_console/manage/sessions/SessionsPage.ts
@@ -1,23 +1,19 @@
 export default class SessionsPage {
-  sessionTypeDrpDwn: string;
-  sessionTypeList: string;
-  allSessionTypesOption: string;
-  regularSSOOption: string;
-  offlineOption: string;
-  directGrantOption: string;
-  serviceAccountOption: string;
-  selectedType: string;
-
-  constructor() {
-    this.sessionTypeDrpDwn = ".pf-c-select__toggle";
-    this.sessionTypeList = ".pf-c-select__toggle + ul";
-    this.allSessionTypesOption = "all-sessions-option";
-    this.regularSSOOption = "regular-sso-option";
-    this.offlineOption = "offline-option";
-    this.directGrantOption = "direct-grant-option";
-    this.serviceAccountOption = "service-account-option";
-    this.selectedType = ".pf-c-select__toggle-text";
-  }
+  sessionTypeDrpDwn = ".pf-c-select__toggle";
+  sessionTypeList = ".pf-c-select__toggle + ul";
+  allSessionTypesOption = "all-sessions-option";
+  regularSSOOption = "regular-sso-option";
+  offlineOption = "offline-option";
+  directGrantOption = "direct-grant-option";
+  serviceAccountOption = "service-account-option";
+  selectedType = ".pf-c-select__toggle-text";
+  revocationActionItem = "revocation";
+  setToNowButton = "set-to-now-button";
+  actionDropdown = "action-dropdown";
+  clearNotBeforeButton = "clear-not-before-button";
+  notBeforeInput = "not-before-input";
+  logoutAll = "logout-all";
+  logoutAllConfirm = "logout-all-confirm-button";
 
   shouldDisplay() {
     cy.get(this.sessionTypeDrpDwn).should("exist");
@@ -33,30 +29,60 @@ export default class SessionsPage {
   selectAllSessionsType() {
     cy.get(this.sessionTypeDrpDwn).click();
     cy.getId(this.allSessionTypesOption).click();
-    cy.get(this.selectedType).should('have.text', 'All session types');
+    cy.get(this.selectedType).should("have.text", "All session types");
   }
 
   selectRegularSSO() {
     cy.get(this.sessionTypeDrpDwn).click();
     cy.getId(this.regularSSOOption).click();
-    cy.get(this.selectedType).should('have.text', 'Regular SSO');
+    cy.get(this.selectedType).should("have.text", "Regular SSO");
   }
 
   selectOffline() {
     cy.get(this.sessionTypeDrpDwn).click();
     cy.getId(this.offlineOption).click();
-    cy.get(this.selectedType).should('have.text', 'Offline');
+    cy.get(this.selectedType).should("have.text", "Offline");
   }
 
   selectDirectGrant() {
     cy.get(this.sessionTypeDrpDwn).click();
     cy.getId(this.directGrantOption).click();
-    cy.get(this.selectedType).should('have.text', 'Direct grant');
+    cy.get(this.selectedType).should("have.text", "Direct grant");
   }
 
   selectServiceAccount() {
     cy.get(this.sessionTypeDrpDwn).click();
     cy.getId(this.serviceAccountOption).click();
-    cy.get(this.selectedType).should('have.text', 'Service account');
+    cy.get(this.selectedType).should("have.text", "Service account");
+  }
+
+  setToNow() {
+    cy.getId(this.actionDropdown).click();
+    cy.getId(this.revocationActionItem).click();
+    cy.getId(this.setToNowButton).click();
+  }
+
+  checkNotBeforeValueExists() {
+    cy.getId(this.actionDropdown).click();
+    cy.getId(this.revocationActionItem).click();
+    cy.getId(this.notBeforeInput).should("not.have.value", "None");
+  }
+
+  clearNotBefore() {
+    cy.getId(this.actionDropdown).click();
+    cy.getId(this.revocationActionItem).click();
+    cy.getId(this.clearNotBeforeButton).click();
+  }
+
+  checkNotBeforeCleared() {
+    cy.getId(this.actionDropdown).click();
+    cy.getId(this.revocationActionItem).click();
+    cy.getId(this.notBeforeInput).should("have.value", "None");
+  }
+
+  logoutAllSessions() {
+    cy.getId(this.actionDropdown).click();
+    cy.getId(this.logoutAll).click();
+    cy.getId(this.logoutAllConfirm).click();
   }
 }
--- a/src/components/view-header/ViewHeader.tsx
+++ b/src/components/view-header/ViewHeader.tsx
@@ -32,6 +32,8 @@ import { HelpItem } from "../help-enabler/HelpItem";
 export type ViewHeaderProps = {
   titleKey: string;
   badges?: ViewHeaderBadge[];
+  badgeId?: string;
+  isDropdownDisabled?: boolean;
   subKey?: string | ReactNode;
   actionsDropdownId?: string;
   subKeyLinkProps?: FormattedLinkProps;
@@ -54,6 +56,7 @@ export const ViewHeader = ({
   actionsDropdownId,
   titleKey,
   badges,
+  isDropdownDisabled,
   subKey,
   subKeyLinkProps,
   dropdownItems,
@@ -138,6 +141,7 @@ export const ViewHeader = ({
                       position={DropdownPosition.right}
                       toggle={
                         <DropdownToggle
+                          isDisabled={isDropdownDisabled}
                           id={actionsDropdownId}
                           onToggle={onDropdownToggle}
                         >
--- a/src/components/view-header/ViewHeader.tsx
+++ b/src/components/view-header/ViewHeader.tsx
@@ -32,7 +32,6 @@ import { HelpItem } from "../help-enabler/HelpItem";
 export type ViewHeaderProps = {
   titleKey: string;
   badges?: ViewHeaderBadge[];
-  badgeId?: string;
   isDropdownDisabled?: boolean;
   subKey?: string | ReactNode;
   actionsDropdownId?: string;
--- a/src/context/whoami/WhoAmI.tsx
+++ b/src/context/whoami/WhoAmI.tsx
@@ -21,6 +21,10 @@ export class WhoAmI {
     return this.me.displayName;
   }
 
+  public getRealm() {
+    return this.me?.realm ?? "";
+  }
+
   public getUserId(): string {
     if (this.me === undefined) return "";
 
--- a/src/sessions/LogoutAllSessionsModal.tsx
+++ b/src/sessions/LogoutAllSessionsModal.tsx
@@ -1,0 +1,73 @@
+import React from "react";
+import {
+  AlertVariant,
+  Button,
+  ButtonVariant,
+  Modal,
+  ModalVariant,
+  TextContent,
+} from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+import { useAdminClient } from "../context/auth/AdminClient";
+import { useRealm } from "../context/realm-context/RealmContext";
+import { useAlerts } from "../components/alert/Alerts";
+
+type RevocationModalProps = {
+  handleModalToggle: () => void;
+};
+
+export const LogoutAllSessionsModal = ({
+  handleModalToggle,
+}: RevocationModalProps) => {
+  const { t } = useTranslation("sessions");
+  const { addAlert } = useAlerts();
+
+  const { realm: realmName } = useRealm();
+  const adminClient = useAdminClient();
+
+  const logoutAllSessions = async () => {
+    try {
+      await adminClient.realms.logoutAll({ realm: realmName });
+      adminClient.keycloak.logout({ redirectUri: "" });
+    } catch (error) {
+      addAlert(t("logoutAllSessionsError", { error }), AlertVariant.danger);
+    }
+  };
+
+  return (
+    <Modal
+      variant={ModalVariant.small}
+      title={t("signOutAllActiveSessionsQuestion")}
+      isOpen={true}
+      onClose={handleModalToggle}
+      actions={[
+        <Button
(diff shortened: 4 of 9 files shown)


src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  useFetch(
+    async () => {
+      const realm = await adminClient.realms.findOne({ realm: realmName });
+
+      return { realm };
+    },
+    ({ realm }) => {
+      setRealm(realm);
+    },
```
edewit, line 60: Can be written a bit simpler:

```suggestion
  useFetch(
    () => adminClient.realms.findOne({ realm: realmName }),
    (realm) => {
      setRealm(realm);
    },
```

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const setToNow = async () => {
+    await adminClient.realms.update(
+      { realm: realmName },
+      {
+        realm: realmName,
+        notBefore: Number(moment().unix()),
+      }
+    );
+    adminClient.keycloak.logout({ redirectUri: "" });
```
edewit, line 72: why would this be necessary?

edewit (commented):
Looking good so far, why is it a draft?

src/sessions/RevocationModal.tsx
jenny-s51: @edewit This is currently how it looks like it's done in the old admin console - it logs the user out and the notBefore time is set successfully - otherwise the call results in a network error if you try setting the `notBefore` and then toggling the modal again to clear it

jenny-s51:
@edewit does this call look okay to you? https://github.com/keycloak/keycloak-admin-ui/compare/master...jenny-s51:sessionsActionsMenu?expand=1#diff-8cb2449f432ea56b8ea79baaf2d01514e4ec7482e1ae615d2569cfaf19f35b15R96-R99

I've declared `id` and `realm` values here but they are not being sent in the request payload (sent as `{}`)

edewit (approved):

edewit:
You don't need `realm` as it will use the "current" realm set on the client

src/sessions/messages.ts
```diff
@@ -14,5 +14,19 @@ export default {
...
     },
+    revocation: "Revocation",
+    revocationDescription:
+      "This is a way to revoke all active sessions and access tokens. Not before means you can revoke any tokens issued before the date.",
+    notBefore: "Not before",
+    notBeforeSuccess: 'Success! "Not before" set for realm',
+    notBeforeError: 'Error clearing "Not Before" for realm: {{error}}',
+    notBeforeClearedSuccess: 'Success! "Not Before" cleared for realm.',
+    signOutAllActiveSessions: "Sign out all active sessions",
+    pushError:
```
jonkoops, line 25: `pushError` is never used as far as I can tell.

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+import type RealmRepresentation from "keycloak-admin/lib/defs/realmRepresentation";
+import { useAdminClient, useFetch } from "../context/auth/AdminClient";
+import moment from "moment";
+import { useRealm } from "../context/realm-context/RealmContext";
+import type ClientRepresentation from "keycloak-admin/lib/defs/clientRepresentation";
+import { useAlerts } from "../components/alert/Alerts";
+import type GlobalRequestResult from "keycloak-admin/lib/defs/globalRequestResult";
+
+type RevocationModalProps = {
+  id?: string;
```
jonkoops, line 28: Looks like the `id` prop is never used or passed, so I believe it can be removed.

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+        t("clients:" + prefixKey + "Success", {
+          successNodes: result.successRequests,
+        }),
+        AlertVariant.success
+      );
+    }
+  };
+
+  const setToNow = async () => {
+    await adminClient.realms.update(
```
jonkoops, line 91: Should this perhaps be wrapped in a `try` block so we can show an error message if something goes wrong?

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+      { realm: realmName },
+      {
+        realm: realmName,
+        notBefore: Number(moment().unix()),
```
jonkoops, line 95: Would `Date.now()` not give the same result here?

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+      addAlert(t("notBeforeError", { error }), AlertVariant.danger);
+    }
+  };
+
+  const push = async (allClients: ClientRepresentation[]) => {
+    for (const client of allClients) {
+      if (client.adminUrl) {
+        const result = (await adminClient.clients.pushRevocation({
+          id: client.id!,
+        })) as unknown as GlobalRequestResult;
```
jonkoops, line 123: This looks like a bug in the admin client? Perhaps we can submit an issue for it?

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+      refresh();
+    } catch (error) {
+      addAlert(t("notBeforeError", { error }), AlertVariant.danger);
+    }
+  };
+
+  const push = async (allClients: ClientRepresentation[]) => {
+    for (const client of allClients) {
+      if (client.adminUrl) {
+        const result = (await adminClient.clients.pushRevocation({
```
jonkoops, line 121: Note that this will block on each request being made, I don't know if that is intentional, but this could be parallelized by using `Promise.all()`.

jonkoops (changes requested):

src/sessions/RevocationModal.tsx
jenny-s51: good idea

src/sessions/RevocationModal.tsx
jenny-s51: Opened https://github.com/keycloak/keycloak-nodejs-admin-client/issues/190

jonkoops (commented):
LGTM :heavy_check_mark:

jonkoops (approved):

ssilvert (changes requested):
If I'm on a realm other than master, "set to now" should not log me out.   Looks like you are calling the correct API, but your code sends a payload while the old console does not.

For push, it's calling the wrong API.  For "foo" realm, your code calls:
http://localhost:8080/auth/admin/realms/foo/clients/20aacb1c-09cf-46f7-aa0a-532a2c219fa6/push-revocation

Old console calls:
http://localhost:8080/auth/admin/realms/foo/push-revocation

ssilvert:
@jenny-s51 Also, I re-ran the tests and looks like they are still failing.  Need to figure out if it's the new code or an intermittent failure.

src/sessions/LogoutAllSessionsModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+      adminClient.keycloak.logout({ redirectUri: "" });
+    } catch (error) {
+      addAlert(t("logoutAllSessionsError", { error }), AlertVariant.danger);
+    }
+  };
+
+  return (
+    <Modal
+      variant={ModalVariant.small}
+      title={t("signOutAllActiveSessions") + "?"}
```
mfrances17, line 41: You shouldn't concatenate strings like this for L10N... it assumes English. Some languages use different punctuation, e.g. spanish uses upside down question mark at beginning and question mark at end. Best to just put the ? in the messages file string and let the translators sub the correct string.

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+          }
+        >
+          <TextInput
+            data-testid="not-before-input"
+            ref={register({ required: true, pattern: emailRegexPattern })}
+            autoFocus
+            isReadOnly
+            value={
+              realm?.notBefore === 0
+                ? (t("None") as string)
```
mfrances17, line 214: this should be lower case 'none' otherwise it's not doing a string lookup it's just forcing 'None'

mfrances17 (changes requested):
Functionality works great from testing the provided steps. A couple of easy string issues you should fix. Otherwise looks good.

jenny-s51:
@ssilvert @mfrances17 thank you for reviewing! Will make the corresponding updates

jenny-s51:
Opened a PR in the nodejs client that adds `pushRevocation` to `realms` so the "Push" calls the correct endpoint as @ssilvert mentioned: https://github.com/keycloak/keycloak-nodejs-admin-client/pull/194

@edewit would be great to get your eyes on it when you have a moment

jonkoops:
@jenny-s51 Looks like your change landed in 73a7cd42656ec7c7fdcec332d1dc84d428cb0976

jenny-s51:
@ssilvert @mfrances17 I've addressed your feedback! Please kindly re-review again when you have a moment

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+      await adminClient.realms.update(
+        { realm: realmName },
+        {
+          realm: realmName,
+          notBefore: Date.now() / 1000,
+        }
+      );
+      if (realmName === "master") {
+        adminClient.keycloak.logout({ redirectUri: "" });
+      }
```
ssilvert, line 99: We should not explicitly log out here.  If you are logged into a realm and you "Set to now" for your own realm, the adapter will log you out.

Conversely, if you "Set to now" for a realm you are not logged into, you will not be logged out.  Never forget that you can get to admin console by logging into a realm that is not master.

Bottom line is that lines 97-99 should be deleted.

ssilvert (commented):
@jenny-s51 This is still not working properly for realms other than master.  We need to talk this afternoon and decide if we should merge and follow up with an issue.

mfrances17 (approved):
LGTM... thanks for making the changes. I swore I had already approved, sorry for the lateness.

src/sessions/SessionsSection.tsx
```diff
@@ -32,15 +38,34 @@ export const SessionsSection = () => {
...
+  >([]);
   const [filterType, setFilterType] = useState(
     t("sessionsType.allSessions").toString()
   );
   const [key, setKey] = useState(0);
+  const [noSessions, setNoSessions] = useState(false);
+
+  const { whoAmI } = useWhoAmI();
+
+  console.log("🤞", whoAmI);
```
jonkoops, line 55: Hahaha, I feel this `console.log` statement. It can be removed however :smile:

cypress/support/pages/admin_console/manage/sessions/SessionsPage.ts
```diff
@@ -1,23 +1,19 @@
...
-    this.sessionTypeDrpDwn = ".pf-c-select__toggle";
-    this.sessionTypeList = ".pf-c-select__toggle + ul";
-    this.allSessionTypesOption = "all-sessions-option";
-    this.regularSSOOption = "regular-sso-option";
-    this.offlineOption = "offline-option";
-    this.directGrantOption = "direct-grant-option";
-    this.serviceAccountOption = "service-account-option";
-    this.selectedType = ".pf-c-select__toggle-text";
-  }
+  sessionTypeDrpDwn = ".pf-c-select__toggle";
```
jonkoops, line 2: Liking this cleanup :100:

src/components/view-header/ViewHeader.tsx
```diff
@@ -28,6 +28,7 @@ export type ViewHeaderProps = {
   badge?: string;
   badgeId?: string;
   badgeIsRead?: boolean;
+  dropdownIsDisabled?: boolean;
```
jonkoops, line 31: I believe the convention used by most props would be `isDropdownDisabled`, but works either way.

src/context/whoami/WhoAmI.tsx
```diff
@@ -20,6 +20,12 @@ export class WhoAmI {
     return this.me.displayName;
   }
 
+  public getRealm(): string {
```
jonkoops, line 23: There is no need to add a return type on this method as the type can be inferred from the returned value.
```suggestion
  public getRealm()
```

src/context/whoami/WhoAmI.tsx
```diff
@@ -20,6 +20,12 @@ export class WhoAmI {
...
+    if (this.me === undefined) return "";
+
+    return this.me.realm;
```
jonkoops, line 26: This code can be shortened using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```suggestion
    return this.me?.realm ?? "";
```

src/sessions/LogoutAllSessionsModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  ModalVariant,
+  TextContent,
+} from "@patternfly/react-core";
+import { useTranslation } from "react-i18next";
+import { useAdminClient } from "../context/auth/AdminClient";
+import { useRealm } from "../context/realm-context/RealmContext";
+import { useAlerts } from "../components/alert/Alerts";
+
+type RevocationModalProps = {
+  id?: string;
```
jonkoops, line 16: The `id` prop is never passed (or used), so I believe it can be removed.

src/sessions/LogoutAllSessionsModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const { t } = useTranslation("sessions");
+  const { addAlert } = useAlerts();
+
+  const { realm: realmName } = useRealm();
+  const adminClient = useAdminClient();
+
+  const logoutAllSessions = async () => {
+    try {
+      await adminClient.realms.logoutAll({ realm: realmName });
+      adminClient.keycloak.logout({ redirectUri: "" });
```
jonkoops, line 32: I see the `keycloak` property has a type of `any`, is this a known bug? If not can you report it on the Keycloak Admin client repo?

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { useAdminClient, useFetch } from "../context/auth/AdminClient";
+import { useRealm } from "../context/realm-context/RealmContext";
+import type ClientRepresentation from "keycloak-admin/lib/defs/clientRepresentation";
+import { useAlerts } from "../components/alert/Alerts";
+import type GlobalRequestResult from "keycloak-admin/lib/defs/globalRequestResult";
+
+type RevocationModalProps = {
+  handleModalToggle: () => void;
+  activeClients: ClientRepresentation[];
+  save: (realm?: UserRepresentation) => void;
```
jonkoops, line 29: I don't see `realm` being passed or used anywhere, so I think this could be removed.
```suggestion
  save: () => void;
```

src/sessions/RevocationModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+      refresh();
+    } catch (error) {
+      addAlert(t("notBeforeError", { error }), AlertVariant.danger);
+    }
+  };
+
+  const push = async () => {
+    const result = adminClient.realms.pushRevocation({
+      realm: realmName,
+    }) as unknown as GlobalRequestResult;
```
jonkoops, line 122: This looks like an issue with the types from the admin client, could you create an issue for that?

jonkoops (changes requested):

src/sessions/LogoutAllSessionsModal.tsx
jenny-s51: opened https://github.com/keycloak/keycloak-nodejs-admin-client/issues/205

src/sessions/RevocationModal.tsx
jenny-s51: nice catch

src/sessions/RevocationModal.tsx
jenny-s51: opened https://github.com/keycloak/keycloak-nodejs-admin-client/issues/206

src/components/view-header/ViewHeader.tsx
```diff
@@ -32,6 +32,8 @@ import { HelpItem } from "../help-enabler/HelpItem";
 export type ViewHeaderProps = {
   titleKey: string;
   badges?: ViewHeaderBadge[];
+  badgeId?: string;
```
jonkoops, line 35: The `badgeId` prop is never used and can be removed.

jonkoops (changes requested):

jonkoops (approved):

ssilvert (approved):
This works fine, but for some reason it always comes back with the message `No push sent. No admin URI configured or no registered cluster nodes available`

This message comes from clients/messages.ts.  I'm going to merge and log an issue.

PR #897, opened 2021-07-23
better waits for realm role tests

edewit:
Added some waits instead of timeouts waiting for backend loads to take place, this will hopefully fix these tests :)

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -72,9 +72,6 @@ describe("Realm settings", () => {
   };
 
   const deleteProvider = (providerName: string) => {
-    const keysUrl = `/auth/admin/realms/${realmName}/keys`;
-    cy.intercept(keysUrl).as("keysFetch");
-
     cy.getId("provider-name-link")
       .contains(providerName)
       .parent()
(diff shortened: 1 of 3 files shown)


jonkoops:
Looks like the test is still failing :cry:

edewit:
@jonkoops well.. the failing test is the `Realm settings` I'll have a look at that one as well

cypress/integration/realm_settings_test.spec.ts
```diff
@@ -72,22 +72,19 @@ describe("Realm settings", () => {
...
       .getId(realmSettingsPage.deleteAction)
       .click();
-    cy.wait(500).getId(realmSettingsPage.modalConfirm).click();
+    cy.getId(realmSettingsPage.modalConfirm).click();
 
     return this;
   };
 
-  const addBundle = () => {
+  /*const addBundle = () => {
```
jonkoops, line 87: Should we get rid of this code since it's commented out?

jonkoops (approved):

cypress/integration/realm_settings_test.spec.ts
edewit: I think it's temporary because of the localisation stuff that was breaking the build with keycloak 14, it will be fixed soon

PR #917, opened 2021-07-27
Enable 'no-empty-function' ESlint rule

jonkoops:
## Motivation
A leftover from trying to align our ESlint config with the recommended config from our plugins.


mfrances17 (approved):
LGTM, did a quick search and it looks like you caught them all as part of this PR.

PR #924, opened 2021-07-28
Add issue template to inform new contributors

jonkoops:

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,3 @@
+<!--
+This repository contains the pre-release Admin Console for Keycloak which is still in development and not enabled by default. If you are looking to report an issue with the released Keycloak Admin Console or other parts of Keycloak, please consult the Keycloak Community (https://www.keycloak.org/community) page to find the correct place to start your discussion.
+-->


edewit (changes requested):
can't we put this into html comments `<!--` and `-->` as we do for the other templates? That way one can leave it but it will not be shown. see: ./keycloak-admin-ui/.github/pull_request_template.md

.github/issue_template.md
```diff
@@ -0,0 +1,5 @@
+## ⚠️ Important ⚠️
+
+This repository contains the new Admin UI for Keycloak which is still experimental and not enabled by default. If you are looking to report an issue with the stable Keycloak Admin UI or other parts of Keycloak, please consult the [Keycloak Community](https://www.keycloak.org/community) page to find the correct place to start your discussion.
```
mfrances17, line 3: The docs refer to it as the 'admin console' or 'Administration Console', I would use either of those throughout instead of 'Admin UI'. 

Also, I recommend the following:
 'pre-release' instead of 'new'
 'still in development' vs 'experimental'
'released Keycloak Administration Console' vs 'stable Keycloak Admin UI'

mfrances17 (changes requested):
Some wording suggestions

.github/issue_template.md
andymunro: The initial screen for the UI calls it the "Administration Console," however everyone calls it the Admin Console and we are standardizing that in all the documentation.  So please call it the Admin Console.

jonkoops: @mfrances17 @andymunro I've incorporated the changes, can you take another look?

jonkoops:
> can't we put this into html comments `<!--` and `-->` as we do for the other templates? 

Done :+1:

andymunro (approved):
Looks good, Jon.

jenny-s51 (approved):
LGTM

edewit (approved):

mfrances17 (approved):
Changes look great, thanks for doing this Jon.

PR #925, opened 2021-07-28
Run Cypress tests on Keycloak theme

jonkoops:
## Motivation
Allows us to run the tests against the final product instead of our own dev server.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,53 +1,50 @@
-name: Cypress run
-
+name: Cypress
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
-
+    branches: [master]
 jobs:
   cypress-run:
-
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [16.x]
-
     steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v2
-      with:
-        node-version: ${{ matrix.node-version }}
-        check-latest: true
+      - uses: actions/checkout@v2
 
-    - run: npm ci
-    - run: npm run build
+      - name: Build Keycloak server
+        run: docker build --tag keycloak .
 
-    - name: Run Keycloak
-      run: ./start.mjs & sleep 40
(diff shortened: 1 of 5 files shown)


.github/workflows/cypress.yml
```diff
@@ -1,53 +1,50 @@
...
+        run: docker build --tag keycloak .
+
+      - name: Start Keycloak server
+        run: |
+          docker run --detach --publish 8080:8080 --env KEYCLOAK_USER=admin --env KEYCLOAK_PASSWORD=admin --name keycloak keycloak
+          sleep 10
+
+      - name: Configure Keycloak server
+        run: |
+          docker exec keycloak /opt/jboss/keycloak/bin/kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin
```
edewit, line 23: this is not needed as you start with `KEYCLOAK_USER`

.github/workflows/cypress.yml
```diff
@@ -1,53 +1,50 @@
...
-        path: assets/screenshots
\ No newline at end of file
+      - uses: actions/checkout@v2
+
+      - name: Build Keycloak server
+        run: docker build --tag keycloak .
+
+      - name: Start Keycloak server
+        run: |
+          docker run --detach --publish 8080:8080 --env KEYCLOAK_USER=admin --env KEYCLOAK_PASSWORD=admin --name keycloak keycloak
```
edewit, line 18: don't we need to add the flag? `-Dprofile.feature.newadmin=enabled`

.github/workflows/cypress.yml
jonkoops: Good catch, I've removed this.

.github/workflows/cypress.yml
jonkoops: Yeah, although I am not sure how.

.github/workflows/cypress.yml
jonkoops: Hmm, this causes an issue where it cannot reach the API. I'll leave it for now.

.github/workflows/cypress.yml
jonkoops: Added this as an argument.

edewit (approved):

PR #933, opened 2021-08-02
Convert more links to use new routes

jonkoops:

### DIFF
--- a/src/realm-settings/RealmSettingsSection.tsx
+++ b/src/realm-settings/RealmSettingsSection.tsx
@@ -55,12 +55,12 @@ export const EditProviderCrumb = () => {
     <>
       <Breadcrumb>
         <BreadcrumbItem
-          component={Link}
-          // @ts-ignore
-          to={toRealmSettings({ realm, tab: "keys" })}
-        >
-          {t("keys")}
-        </BreadcrumbItem>
+          render={(props) => (
+            <Link {...props} to={toRealmSettings({ realm, tab: "keys" })}>
+              {t("keys")}
+            </Link>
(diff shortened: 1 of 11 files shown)


src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -52,7 +53,11 @@ export const EditProviderCrumb = () => {
   return (
     <>
       <Breadcrumb>
-        <BreadcrumbItem to={`#/${realm}/realm-settings/keys`}>
+        <BreadcrumbItem
+          component={Link}
+          // @ts-ignore
+          to={toRealmSettings({ realm, tab: "keys" })}
```
jenny-s51, line 59: This should redirect back to `/realm-settings/keys`

jenny-s51 (changes requested):
Nice work @jonkoops - most of these paths work great! One small issue with the breadcrumb but otherwise LGTM. Thank you for putting up your routing PRs in medium-ish chunks like this -- it's helpful for reviewing :)

src/realm-settings/RealmSettingsSection.tsx
jonkoops: Good find! This seems to be a bug in Patternfly, I've created a PR to get it fixed: https://github.com/patternfly/patternfly-react/pull/6111

jonkoops: I've fixed this by using the `render` prop instead.

ssilvert:
@jenny-s51 We are merging this.  Please let us know if you have any issues.

PR #934, opened 2021-08-02
Realm settings(tokens): Add tokens tab

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-1103

## Verification Steps

1. Go to `Realm settings -> tokens`.
2. Populate form fields.
3. Save and verify data is saved accurately. 

## Checklist:


src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -341,6 +342,19 @@ export const RealmSettingsSection = () => {
...
             </Tab>
+            <Tab
+              id="tokens"
+              eventKey="tokens"
+              data-testid="rs-tokens-tab"
+              aria-label="tokens-tab"
+              title={<TabTitleText>{t("realm-settings:tokens")}</TabTitleText>}
+            >
+              <RealmSettingsTokensTab
+                key={key}
```
jonkoops, line 353: I don't believe a `key` prop is needed here.

src/realm-settings/SessionsTab.tsx
```diff
@@ -199,7 +199,7 @@ export const RealmSettingsSessionsTab = ({
           </FormAccess>
         </FormPanel>
         <FormPanel
-          title={t(".pf-c-data-list__item-draggable-iconclientSessionSettings")}
```
jonkoops, line 202: Good catch!

src/util.ts
```diff
@@ -94,3 +110,32 @@ export const getBaseUrl = (adminClient: KeycloakAdminClient) => {
 
 export const emailRegexPattern =
   /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+export const forHumans = (seconds: number) => {
```
jonkoops, line 114: This does not really work for other languages than English, it might be better to use the [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat) API here to localize.

src/realm-settings/TokensTab.tsx
```diff
@@ -0,0 +1,557 @@
...
+  };
+
+  useFetch(
+    () => adminClient.realms.findOne({ realm: realmName }),
+    (realm) => {
+      setRealm(realm);
+      setupForm(realm);
+    },
+    [realmName]
+  );
```
jonkoops, line 95: Why would we fetch the realm again if it is already being passed in by the parent component?

src/realm-settings/TokensTab.tsx
```diff
@@ -0,0 +1,557 @@
...
+
+  const save = async () => {
+    const attributes = flatten(form.getValues()["attributes"]);
```
jonkoops, line 98: Why do we need to perform this `flatten` operation? Under which circumstances would this be a nested structure?

jonkoops (changes requested):

src/util.ts
jenny-s51: Thank you for suggesting this API - I tested this function with it but `Intl.RelativeTimeFormat` doesn't like big numbers  (`error: values must be finite`) therefore I localized the strings.

src/realm-settings/TokensTab.tsx
jenny-s51: Any attribute of `attributes` that begins with the value `attributes.actionTokenGeneratedByUserLifespan` becomes a nested structure due to React Hook Form. RHF reads these attribute names as objects using dot notation, and therefore nests them.  There are four instances of this; here is an example of how the PUT call looks in the old console when these values are updated.

```
actionTokenGeneratedByUserLifespan.execute-actions: "360"
actionTokenGeneratedByUserLifespan.idp-verify-account-via-email: "300"
actionTokenGeneratedByUserLifespan.reset-credentials: "420"
actionTokenGeneratedByUserLifespan.verify-email: "120"
```
The new console needs to send the above format.

src/util.ts
jonkoops: That is already better, however we cannot rely on the order of things in language unfortunately. So in this case we'd have to interpolate the value into the translation.

jenny-s51: Good idea. updated

jonkoops (approved):

PR #936, opened 2021-08-03
Realm settings(localization): add locale dropdown to table

jenny-s51:
## Motivation

https://issues.redhat.com/browse/APPDUX-1066

## Brief Description

Adds locale filter to message bundle table as shown in https://marvelapp.com/prototype/56d9606/screen/76458142

## Verification Steps

1. Go to `RealmSettings.tsx` and uncomment the Localization tab.
2. Go to Realm Settings -> Localization.
3. As per the design, the locale filter in the table should be disabled until the form above is modified and saved.
4. Save the form and verify that the new locale select filter works as expected.

## Checklist:

### DIFF
--- a/src/PageNav.tsx
+++ b/src/PageNav.tsx
@@ -7,6 +7,7 @@ import {
   NavGroup,
   NavList,
   PageSidebar,
+  Divider,
 } from "@patternfly/react-core";
 
 import { RealmSelector } from "./components/realm-selector/RealmSelector";
@@ -90,13 +91,9 @@ export const PageNav: React.FunctionComponent = () => {
               <RealmSelector />
             </NavItem>
           </NavList>
-          {!isOnAddRealm && (
-            <NavGroup title="">
-              <LeftNav title="home" path="/" />
-            </NavGroup>
-          )}
+          <Divider />
           {showManage && !isOnAddRealm && (
-            <NavGroup title={t("manage")}>
+            <NavGroup aria-label={t("manage")} title={t("manage")}>
               <LeftNav title="clients" path="/clients" />
               <LeftNav title="clientScopes" path="/client-scopes" />
               <LeftNav title="realmRoles" path="/roles" />
@@ -108,7 +105,7 @@ export const PageNav: React.FunctionComponent = () => {
           )}
 
           {showConfigure && !isOnAddRealm && (
-            <NavGroup title={t("configure")}>
+            <NavGroup aria-label={t("configure")} title={t("configure")}>
               <LeftNav title="realmSettings" path="/realm-settings" />
               <LeftNav title="authentication" path="/authentication" />
               <LeftNav title="identityProviders" path="/identity-providers" />
--- a/src/components/table-toolbar/KeycloakDataTable.tsx
+++ b/src/components/table-toolbar/KeycloakDataTable.tsx
@@ -375,6 +375,7 @@ export function KeycloakDataTable<T>({
       {((data && data.length > 0) ||
         search !== "" ||
         isSearching ||
+        emptyState ||
         loading) && (
         <PaginatingTableToolbar
           count={data?.length || 0}
@@ -413,7 +414,8 @@ export function KeycloakDataTable<T>({
           {!loading &&
             (!data || data.length === 0) &&
             (search !== "" || !isSearching) &&
-            searchPlaceholderKey && (
+            searchPlaceholderKey &&
+            !emptyState && (
               <ListEmptyState
                 hasIcon={true}
                 icon={icon}
--- a/src/realm-settings/LocalizationTab.tsx
+++ b/src/realm-settings/LocalizationTab.tsx
@@ -5,13 +5,16 @@ import {
   ActionGroup,
   AlertVariant,
   Button,
+  Divider,
   FormGroup,
   PageSection,
   Select,
+  SelectGroup,
   SelectOption,
   SelectVariant,
   Switch,
   TextContent,
+  ToolbarItem,
 } from "@patternfly/react-core";
 
 import type RealmRepresentation from "keycloak-admin/lib/defs/realmRepresentation";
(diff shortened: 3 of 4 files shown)


src/PageNav.tsx
```diff
@@ -81,12 +81,12 @@ export const PageNav: React.FunctionComponent = () => {
             </NavItem>
           </NavList>
           {!isOnAddRealm && (
-            <NavGroup title="">
+            <NavGroup aria-labelledby="" title="">
```
jonkoops, line 84: Should we not be adding a value here?

src/PageNav.tsx
```diff
@@ -81,12 +81,12 @@ export const PageNav: React.FunctionComponent = () => {
...
               <LeftNav title="home" path="/" />
             </NavGroup>
           )}
           {showManage && !isOnAddRealm && (
-            <NavGroup title={t("manage")}>
+            <NavGroup aria-labelledby={t("manage")} title={t("manage")}>
```
jonkoops, line 89: Be aware that `aria-labelledby` references the `id` of an element that contains a descriptive text and should not be confused with `aria-label` which provides a text directly.

So probably what we need to do here is:
```suggestion
            <NavGroup aria-label={t("manage")} title={t("manage")}>
```

src/PageNav.tsx
```diff
@@ -98,7 +98,7 @@ export const PageNav: React.FunctionComponent = () => {
           )}
 
           {showConfigure && !isOnAddRealm && (
-            <NavGroup title={t("configure")}>
+            <NavGroup aria-labelledby={t("configure")} title={t("configure")}>
```
jonkoops, line 101:
```suggestion
            <NavGroup aria-label={t("configure")} title={t("configure")}>
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,7 +81,18 @@ export const LocalizationTab = ({
...
+    return [[]];
+  };
+
+  const tableLoader = async () => {
+    if (realm) {
+      const result = await adminClient.realms.getRealmLocalizationTexts({
+        realm: currentRealm,
+        selectedLocale: selectMenuLocale,
       });
       return Object.keys(result).map((key) => [key, result[key]]);
```
jonkoops, line 97: Not entirely sure, but we might be able to replace this with `Object.entries()`.
```suggestion
      return Object.entries(result);
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,7 +81,18 @@ export const LocalizationTab = ({
     if (realm) {
       const result = await adminClient.realms.getRealmLocalizationTexts({
         realm: realm.realm!,
-        selectedLocale: getValues("defaultLocale") || "en",
+        selectedLocale: selectMenuLocale || getValues("defaultLocale") || "en",
+      });
+      return Object.keys(result).map((key) => [key, result[key]]);
```
jonkoops, line 86: Not entirely sure, but we might be able to replace this with `Object.entries()`.
```suggestion
      return Object.entries(result);
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -87,6 +103,34 @@ export const LocalizationTab = ({
     setAddMessageBundleModalOpen(!addMessageBundleModalOpen);
   };
 
+  const options = [
+    <SelectGroup label={t("defaultLocale")} key="group1">
+      {watchSupportedLocales
+        .filter((item) => item === realm?.defaultLocale)
+        .map((locale: string, idx: number) => (
+          <SelectOption key={`supported-locale-${idx}`} value={locale}>
```
jonkoops, line 111: Is `locale` not already unique in this context? React [recommends](https://reactjs.org/docs/lists-and-keys.html#keys) using the `index` as a key only as a last resort.
```suggestion
          <SelectOption key={locale} value={locale}>
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -87,6 +103,34 @@ export const LocalizationTab = ({
     setAddMessageBundleModalOpen(!addMessageBundleModalOpen);
   };
 
+  const options = [
+    <SelectGroup label={t("defaultLocale")} key="group1">
+      {watchSupportedLocales
+        .filter((item) => item === realm?.defaultLocale)
+        .map((locale: string, idx: number) => (
```
jonkoops, line 110: These types are already known, so there is no need to specify them.
```suggestion
        .map((locale, idx) => (
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -87,6 +103,34 @@ export const LocalizationTab = ({
...
+          <SelectOption key={`supported-locale-${idx}`} value={locale}>
+            {t(`allSupportedLocales.${locale}`)}
+          </SelectOption>
+        ))}
+    </SelectGroup>,
+    <Divider key="divider" />,
+    <SelectGroup label={t("supportedLocales")} key="group2">
+      {watchSupportedLocales
+        .filter((item) => item !== realm?.defaultLocale)
+        .map((locale: string, idx: number) => (
```
jonkoops, line 120: These types are already known, so there is no need to specify them.
```suggestion
        .map((locale, idx) => (
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -87,6 +103,34 @@ export const LocalizationTab = ({
...
+          <SelectOption key={`supported-locale-${idx}`} value={locale}>
```
jonkoops, line 121: Is `locale` not already unique in this context? React [recommends](https://reactjs.org/docs/lists-and-keys.html#keys) using the `index` as a key only as a last resort.
```suggestion
          <SelectOption key={locale} value={locale}>
```

jonkoops (changes requested):

src/PageNav.tsx
jenny-s51: I'm not sure we need this `<NavGroup>` here at all - it doesn't render anything in the left nav. Do you think we can just put a divider component here instead?

src/realm-settings/LocalizationTab.tsx
jenny-s51: Thank you for the suggestion! Looks like the `loader` prop on `KeycloakDataTable` doesn't like this return type

src/realm-settings/LocalizationTab.tsx
jenny-s51: Thank you for the suggestion! Looks like the loader prop on KeycloakDataTable doesn't like this return type

src/PageNav.tsx
jenny-s51: Good suggestion! thank you for clarifying

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -137,6 +137,7 @@ export type DataListProps<T> = Omit<
   actionResolver?: IActionsResolver;
   searchTypeComponent?: ReactNode;
   toolbarItem?: ReactNode;
+  keepToolbar?: boolean;
```
edewit, line 140: no need to add this we already have a property to do this called `isSearching`, guess we could rename it instead.

edewit (changes requested):
had one strange error when testing this the dropdown on the table got disabled after a reload:
image

src/components/table-toolbar/KeycloakDataTable.tsx
jenny-s51: thank you - now using the prop `isSearching` on the data table within the localization tab

jenny-s51:
@edewit According to this mock, the actions menu should be disabled until Localization configs are saved: https://marvelapp.com/prototype/56d9606/screen/76458134

edewit (approved):

src/realm-settings/LocalizationTab.tsx
jonkoops: This is because the return type is mixed, the `catch` clause returns an empty array which is not assignable to `[string, string]`.

jonkoops (approved):

PR #947, opened 2021-08-06
add priority attribute when dragging data list elements

### LINKED ISSUE #603: Realm settings(keys): providers list updates
Sort by priority and add delete functionality to providers list as shown in: https://marvelapp.com/prototype/99i86j9/screen/75627938

jenny-s51:
## Motivation

Closes #603 

## Verification Steps

1. Go to Realm Settings -> Keys -> Providers list
2. Click and drag a provider to a new location
3. Inspect the network call and verify that the priority is changed based on the location in the providers list

## Checklist:

### DIFF
--- a/src/realm-settings/JavaKeystoreModal.tsx
+++ b/src/realm-settings/JavaKeystoreModal.tsx
@@ -21,7 +21,7 @@ import { useAlerts } from "../components/alert/Alerts";
 import type ComponentRepresentation from "keycloak-admin/lib/defs/componentRepresentation";
 import { HelpItem } from "../components/help-enabler/HelpItem";
 import { useServerInfo } from "../context/server-info/ServerInfoProvider";
-import { KeyProviderType } from "../util";
+import { KEY_PROVIDER_TYPE } from "../util";
 
 type JavaKeystoreModalProps = {
   providerType: string;
@@ -45,7 +45,8 @@ JavaKeystoreModalProps) => {
   const [isEllipticCurveDropdownOpen, setIsEllipticCurveDropdownOpen] =
     useState(false);
 
-  const allComponentTypes = serverInfo.componentTypes?.[KeyProviderType] ?? [];
+  const allComponentTypes =
+    serverInfo.componentTypes?.[KEY_PROVIDER_TYPE] ?? [];
 
   const save = async (component: ComponentRepresentation) => {
     try {
@@ -53,7 +54,7 @@ JavaKeystoreModalProps) => {
         ...component,
         parentId: component.parentId,
         providerId: providerType,
-        providerType: KeyProviderType,
+        providerType: KEY_PROVIDER_TYPE,
         config: { priority: ["0"] },
       });
       handleModalToggle();
--- a/src/realm-settings/JavaKeystoreModal.tsx
+++ b/src/realm-settings/JavaKeystoreModal.tsx
@@ -54,6 +54,7 @@ JavaKeystoreModalProps) => {
         parentId: component.parentId,
         providerId: providerType,
         providerType: KeyProviderType,
+        config: { priority: ["0"] },
       });
       handleModalToggle();
       addAlert(t("saveProviderSuccess"), AlertVariant.success);
--- a/src/realm-settings/JavaKeystoreModal.tsx
+++ b/src/realm-settings/JavaKeystoreModal.tsx
@@ -21,6 +21,7 @@ import { useAlerts } from "../components/alert/Alerts";
 import type ComponentRepresentation from "keycloak-admin/lib/defs/componentRepresentation";
 import { HelpItem } from "../components/help-enabler/HelpItem";
 import { useServerInfo } from "../context/server-info/ServerInfoProvider";
+import { KeyProviderType } from "../util";
 
 type JavaKeystoreModalProps = {
   providerType: string;
@@ -44,8 +45,7 @@ JavaKeystoreModalProps) => {
   const [isEllipticCurveDropdownOpen, setIsEllipticCurveDropdownOpen] =
     useState(false);
 
-  const allComponentTypes =
-    serverInfo.componentTypes?.["org.keycloak.keys.KeyProvider"] ?? [];
+  const allComponentTypes = serverInfo.componentTypes?.[KeyProviderType] ?? [];
 
   const save = async (component: ComponentRepresentation) => {
     try {
@@ -53,7 +53,7 @@ JavaKeystoreModalProps) => {
         ...component,
         parentId: component.parentId,
         providerId: providerType,
-        providerType: "org.keycloak.keys.KeyProvider",
+        providerType: KeyProviderType,
       });
       handleModalToggle();
       addAlert(t("saveProviderSuccess"), AlertVariant.success);
--- a/src/realm-settings/KeysProvidersTab.tsx
+++ b/src/realm-settings/KeysProvidersTab.tsx
@@ -82,8 +82,6 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
     serverInfo.componentTypes?.[KEY_PROVIDER_TYPE] ?? []
   ).map((item) => item.id);
 
-  const itemIds = components.map((component) => component.id!);
-
   const [itemOrder, setItemOrder] = useState<string[]>([]);
   const [providerDropdownOpen, setProviderDropdownOpen] = useState(false);
 
@@ -96,6 +94,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   const [liveText, setLiveText] = useState("");
 
   useEffect(() => {
+    const itemIds = components.map((component) => component.id!);
     setItemOrder(itemIds);
   }, [components, searchVal]);
 
@@ -138,20 +137,17 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
     setLiveText(t("common:onDragFinish"));
     const updateAll = components.map((component: ComponentData) => {
       const componentToSave = { ...component };
-      componentToSave.providerDescription = undefined;
+      delete componentToSave.providerDescription;
 
       return adminClient.components.update(
         { id: component.id! },
         {
           ...componentToSave,
-          parentId: component?.parentId,
-          providerId: component?.providerId,
-          providerType: KEY_PROVIDER_TYPE,
           config: {
             priority: [
               (
                 itemOrder.length -
-                itemOrder.indexOf(component.id || "") +
+                itemOrder.indexOf(component.id!) +
                 100
               ).toString(),
             ],
@@ -315,7 +311,6 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
             </ToolbarGroup>
           </>
         </Toolbar>
-
         <DataList
           aria-label={t("groups")}
           onDragFinish={onDragFinish}
@@ -326,11 +321,10 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
           isCompact
         >
           <DataListItem aria-labelledby={"aria"} id="data" key="data">
-            <DataListItemRow className="test" data-testid={"data-list-row"}>
+            <DataListItemRow className="test" data-testid="data-list-row">
               <DataListDragButton
                 className="header-drag-button"
                 aria-label="Reorder"
-                aria-labelledby="simple-item"
                 aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
                 aria-pressed="false"
                 isDisabled
@@ -338,13 +332,16 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
               <DataListItemCells
                 className="data-list-cells"
                 dataListCells={[
-                  <DataListCell className="name" key={"1"}>
+                  <DataListCell className="name" key="name">
                     <>{t("realm-settings:name")}</>
                   </DataListCell>,
-                  <DataListCell className="provider" key={"2"}>
+                  <DataListCell className="provider" key="provider">
                     <>{t("realm-settings:provider")}</>
                   </DataListCell>,
-                  <DataListCell className="provider-description" key={"3"}>
+                  <DataListCell
+                    className="provider-description"
+                    key="provider-description"
+                  >
                     <>{t("realm-settings:providerDescription")}</>
                   </DataListCell>,
                 ]}
@@ -354,20 +351,19 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
           {(filteredComponents.length === 0
             ? components
             : filteredComponents
-          ).map((component: ComponentData, idx: number) => (
+          ).map((component, idx) => (
             <DataListItem
               draggable
               aria-labelledby={"aria"}
               key={component.id}
               id={component.id}
             >
-              <DataListItemRow key={idx} data-testid={"data-list-row"}>
+              <DataListItemRow data-testid="data-list-row">
                 <DataListControl>
                   <Tooltip content={t("dragInstruction")} position="top">
                     <DataListDragButton
                       className="row-drag-button"
                       aria-label="Reorder"
-                      aria-labelledby="simple-item2"
                       aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
                       aria-pressed="false"
                     />
--- a/src/realm-settings/KeysProvidersTab.tsx
+++ b/src/realm-settings/KeysProvidersTab.tsx
@@ -325,7 +325,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
               <DataListDragButton
                 className="header-drag-button"
                 aria-label="Reorder"
-                aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
+                aria-describedby={t("common-help:dragHelp")}
                 aria-pressed="false"
                 isDisabled
               />
@@ -364,7 +364,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
                     <DataListDragButton
                       className="row-drag-button"
                       aria-label="Reorder"
-                      aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
+                      aria-describedby={t("common-help:dragHelp")}
                       aria-pressed="false"
                     />
                   </Tooltip>
--- a/src/realm-settings/KeysProvidersTab.tsx
+++ b/src/realm-settings/KeysProvidersTab.tsx
@@ -23,6 +23,7 @@ import {
   Toolbar,
   ToolbarGroup,
   ToolbarItem,
+  Tooltip,
 } from "@patternfly/react-core";
 import { SearchIcon } from "@patternfly/react-icons";
 
@@ -84,9 +85,6 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
 
   const [itemOrder, setItemOrder] = useState<string[]>([]);
   const [providerDropdownOpen, setProviderDropdownOpen] = useState(false);
-  // const [sortedByPriority, setSortedByPriority] = useState<
-  //   ComponentRepresentation[]
-  // >([]);
   const [currentComponent, setCurrentComponent] =
     useState<ComponentRepresentation>();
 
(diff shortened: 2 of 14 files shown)


src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -79,10 +79,12 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
     "org.keycloak.keys.KeyProvider"
   ].map((item) => item.id);
 
-  const itemIds = components.map((_, idx) => "data" + idx);
+  const itemIds = components.map((component) => component.id);
```
jonkoops, line 82: This is probably one of those cases where a model from Keycloak is missing type information about whether the field is actually optional. You might be able to use a non-null assertion here to get rid of the undefined values.
```suggestion
  const itemIds = components.map((component) => component.id!);
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -79,10 +79,12 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
 
-  const [itemOrder, setItemOrder] = useState<string[]>([]);
+  const [itemOrder, setItemOrder] = useState<(string | undefined)[]>([]);
```
jonkoops, line 84: And thus
```suggestion
  const [itemOrder, setItemOrder] = useState<string[]>([]);
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -93,7 +95,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   const [liveText, setLiveText] = useState("");
 
   useEffect(() => {
-    setItemOrder(["data", ...itemIds]);
+    setItemOrder([...itemIds]);
```
jonkoops, line 98: If the value is already an array there should be no need to spread it.
```suggestion
    setItemOrder(itemIds);
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -130,9 +133,23 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
     setLiveText(t("onDragCancel"));
   };
 
-  const onDragFinish = (itemOrder: string[]) => {
-    setItemOrder(["data", ...itemOrder.filter((i) => i !== "data")]);
+  const onDragFinish = async (itemOrder: string[]) => {
+    setItemOrder([...itemOrder]);
```
jonkoops, line 137: Same here, I don't think this needs to be spread.
```suggestion
    setItemOrder(itemOrder);
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -130,9 +133,23 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
     setLiveText(t("onDragCancel"));
+    try {
+      await adminClient.components.update(
+        { id },
+        {
+          ...currentComponent,
+          parentId: currentComponent?.parentId,
+          providerId: currentComponent?.providerId,
+          providerType: "org.keycloak.keys.KeyProvider",
```
jonkoops, line 146: I am seeing this string a lot in the project, perhaps it's time to put in a constant in a shared file?

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -285,7 +302,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
           onDragStart={onDragStart}
           onDragMove={onDragMove}
           onDragCancel={onDragCancel}
-          itemOrder={itemOrder}
+          itemOrder={itemOrder as string[]}
```
jonkoops, line 305: If the aforementioned changes are applied this can be removed.
```suggestion
          itemOrder={itemOrder}
```

jonkoops (changes requested):

mfrances17 (changes requested):
Code looks good with jon's suggestions. Drag and dropped worked well. 

Functionality does not seem to be working, though. When I drag and drop the keys in a different order in the Providers tab, it should change the order of keys by Provider in the Active keys list, but it doesn't. 

In the old console, I can go to Keys > Providers and assign different values to the priority, and when I go to the Active tab, the Algorithms are listed in order from high to low priority.

jenny-s51:
@mfrances17 nice catch - updated based @edewit's suggestion to change all the element's priorities so they are sorted correctly

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -131,8 +134,37 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   };
 
   const onDragFinish = (itemOrder: string[]) => {
-    setItemOrder(["data", ...itemOrder.filter((i) => i !== "data")]);
+    setItemOrder(itemOrder);
     setLiveText(t("common:onDragFinish"));
+    const updateAll = components.map((component: ComponentData) => {
+      const componentToSave = { ...component };
```
edewit, line 140: this isn't necessary componentToSave === component

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -131,8 +134,37 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
+                100
+              ).toString(),
+            ],
+          },
+        }
+      );
+    });
+
+    try {
+      Promise.all(updateAll);
```
edewit, line 164:
```suggestion
      await Promise.all(updateAll);
```

edewit: Don't we need a success message?

edewit (approved):
:+1: looks good

src/realm-settings/KeysProvidersTab.tsx
jonkoops: I think this is to prevent `providerDescription` from getting changed in memory.

src/util.ts
```diff
@@ -189,3 +188,5 @@ export const interpolateTimespan = (forHumans: string) => {
     });
   }
 };
+
+export const KeyProviderType = "org.keycloak.keys.KeyProvider";
```
jonkoops, line 192: I think it's preferred to have our constants as screaming snake case (no joke this is what is called apparently). Otherwise it might be confused with a class or enum.
```suggestion
export const KEY_PROVIDER_TYPE = "org.keycloak.keys.KeyProvider";
```

src/util.ts
```diff
@@ -89,7 +89,6 @@ export const convertFormValuesToObject = (
     const newKey = firstInstanceOnly
       ? key.replace(/-/, ".")
       : key.replace(/-/g, ".");
-    console.log(newKey);
```
jonkoops, line 92: Nice catch!

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -164,22 +164,30 @@ export const RealmSettingsSection = () => {
   const [currentUser, setCurrentUser] = useState<UserRepresentation>();
   const { whoAmI } = useWhoAmI();
 
-  const kpComponentTypes =
-    useServerInfo().componentTypes!["org.keycloak.keys.KeyProvider"];
+  const kpComponentTypes = useServerInfo().componentTypes![KeyProviderType];
+
+  const sortByPriority = (components: ComponentRepresentation[]) => {
```
jonkoops, line 169: Looks like this function has no direct dependencies inside of the component, so it can be placed outside of it.

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -164,22 +164,30 @@ export const RealmSettingsSection = () => {
...
+    const sortedComponents = components?.sort(
```
jonkoops, line 170: According to the function signature the `components` argument is always defined, so there should be no need to use optional chaining here.
```suggestion
    const sortedComponents = components.sort(
```

jonkoops: Be aware that `.sort()` uses an [in-place](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) algorithm, this means that the original array will be changes as well. You might want to create a new array before operating on it.

For example:
```suggestion
    const sortedComponents = [...components].sort(
```

Or
```suggestion
    const sortedComponents = Array.from(components).sort(
```

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -164,22 +164,30 @@ export const RealmSettingsSection = () => {
...
+      // sort in descending order
+      (a, b) => Number(b.config?.priority) - Number(a?.config?.priority || 0)
```
jonkoops, line 172: According to the type definition `priority` here can also be a `string[]`. In this case, or when passing in `undefined` into a `Number()` constructor this will result in a value of [`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) (not a number).

Since it's not possible to do arithmetic on `NaN`, this code might be causing issues.

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -164,22 +164,30 @@ export const RealmSettingsSection = () => {
   const [currentUser, setCurrentUser] = useState<UserRepresentation>();
   const { whoAmI } = useWhoAmI();
 
-  const kpComponentTypes =
-    useServerInfo().componentTypes!["org.keycloak.keys.KeyProvider"];
+  const kpComponentTypes = useServerInfo().componentTypes![KeyProviderType];
```
jonkoops, line 167: It might be better to use optional chaining with a default value here.
```suggestion
  const kpComponentTypes = useServerInfo().componentTypes?.[KeyProviderType] ?? [];
```

src/realm-settings/JavaKeystoreModal.tsx
```diff
@@ -44,16 +45,16 @@ JavaKeystoreModalProps) => {
...
 
   const save = async (component: ComponentRepresentation) => {
     try {
       await adminClient.components.create({
         ...component,
         parentId: component.parentId,
         providerId: providerType,
-        providerType: "org.keycloak.keys.KeyProvider",
+        providerType: KeyProviderType,
+        config: { priority: ["0"] },
```
jonkoops, line 57: This line wasn't here before, is that intentional?

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -75,11 +78,11 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   const [isCreateModalOpen, setIsCreateModalOpen] = useState(false);
 
   const serverInfo = useServerInfo();
-  const providerTypes = serverInfo.componentTypes![
-    "org.keycloak.keys.KeyProvider"
-  ].map((item) => item.id);
+  const providerTypes = serverInfo.componentTypes![KeyProviderType].map(
+    (item) => item.id
+  );
```
jonkoops, line 83: It might be better to use optional chaining with a default here in case the value is not set.
```suggestion
  const providerTypes = (
    serverInfo.componentTypes?.[KeyProviderType] ?? []
  ).map((item) => item.id);
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -93,7 +96,7 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   const [liveText, setLiveText] = useState("");
 
   useEffect(() => {
-    setItemOrder(["data", ...itemIds]);
+    setItemOrder(itemIds);
   }, [components, searchVal]);
```
jonkoops, line 100: Be aware that `components` does not seem to be a stable value (by reference equality) between renders, meaning this effect might be triggered every time the parent component renders.

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -75,11 +78,11 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
   const serverInfo = useServerInfo();
-  const providerTypes = serverInfo.componentTypes![
-    "org.keycloak.keys.KeyProvider"
-  ].map((item) => item.id);
+  const providerTypes = serverInfo.componentTypes![KeyProviderType].map(
+    (item) => item.id
+  );
 
-  const itemIds = components.map((_, idx) => "data" + idx);
+  const itemIds = components.map((component) => component.id!);
```
jonkoops, line 85: Looks like `itemIds` is only used to in an effect to set `itemOrder`. Perhaps it can be included in that effect only so this code is not executed unnecessarily?

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -131,8 +134,37 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
   };
 
   const onDragFinish = (itemOrder: string[]) => {
-    setItemOrder(["data", ...itemOrder.filter((i) => i !== "data")]);
+    setItemOrder(itemOrder);
     setLiveText(t("common:onDragFinish"));
+    const updateAll = components.map((component: ComponentData) => {
+      const componentToSave = { ...component };
+      componentToSave.providerDescription = undefined;
```
jonkoops, line 141: An alternative approach for this could be:
```suggestion
      const componentToSave = { ...component, providerDescription: undefined };
```
Or
```suggestion
      const componentToSave = { ...component };
      delete componentToSave.providerDescription;
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -131,8 +134,37 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
+
+      return adminClient.components.update(
+        { id: component.id! },
+        {
+          ...componentToSave,
+          parentId: component?.parentId,
+          providerId: component?.providerId,
```
jonkoops, line 148: Are these fields not already provided by spreading `componentToSave`?

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -131,8 +134,37 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
+          providerType: KeyProviderType,
+          config: {
+            priority: [
+              (
+                itemOrder.length -
+                itemOrder.indexOf(component.id || "") +
```
jonkoops, line 154: I think this is one of those cases where `id` is always present but the types are not correct. You can use a non-null assertion here to fix that.
```suggestion
                itemOrder.indexOf(component.id!) +
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -317,22 +349,24 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
           {(filteredComponents.length === 0
             ? components
             : filteredComponents
-          ).map((component: ComponentData, idx) => (
+          ).map((component: ComponentData, idx: number) => (
```
jonkoops, line 352: The types can be inferred here from the value of the array we're mapping on, so the type annotations can be omitted.
```suggestion
          ).map((component, idx) => (
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -317,22 +349,24 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
             <DataListItem
               draggable
               aria-labelledby={"aria"}
-              key={`data${idx}`}
-              id={`data${idx}`}
+              key={component.id}
+              id={component.id}
             >
               <DataListItemRow key={idx} data-testid={"data-list-row"}>
```
jonkoops, line 359: The `key` prop here is not needed as these elements are static.
```suggestion
              <DataListItemRow data-testid={"data-list-row"}>
```

src/realm-settings/KeysProvidersTab.tsx
```diff
@@ -317,22 +349,24 @@ export const KeysTabInner = ({ components, refresh }: KeysTabInnerProps) => {
...
-                    aria-labelledby="simple-item2"
-                    aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
-                    aria-pressed="false"
-                  />
+                  <Tooltip content={t("dragInstruction")} position="top">
+                    <DataListDragButton
+                      className="row-drag-button"
+                      aria-label="Reorder"
+                      aria-labelledby="simple-item2"
+                      aria-describedby="Press space or enter to begin dragging, and use the arrow keys to navigate up or down. Press enter to confirm the drag, or any other key to cancel the drag operation."
```
jonkoops, line 366: Some of these ARIA attributes will need to be translated. It is also not needed to add an `aria-labelledby` attribute if an `aria-label` is already provided.

src/realm-settings/RSAGeneratedModal.tsx
```diff
@@ -43,16 +44,16 @@ export const RSAGeneratedModal = ({
...
 
   const save = async (component: ComponentRepresentation) => {
     try {
       await adminClient.components.create({
         ...component,
         parentId: component.parentId,
         providerId: providerType,
-        providerType: "org.keycloak.keys.KeyProvider",
+        providerType: KeyProviderType,
+        config: { priority: ["0"] },
```
jonkoops, line 56: This line was not there before, is that intentional?

src/realm-settings/RSAModal.tsx
```diff
@@ -70,7 +70,8 @@ export const RSAModal = ({
           ...component,
           parentId: component.parentId,
           providerId: providerType,
-          providerType: "org.keycloak.keys.KeyProvider",
+          providerType: KeyProviderType,
+          config: { priority: ["0"] },
```
jonkoops, line 74: This line was not there before, is that intentional?

jonkoops (changes requested):

src/realm-settings/JavaKeystoreModal.tsx
jenny-s51: Yes, the old console sets the default priority for every new provider to 0

src/realm-settings/RealmSettingsSection.tsx
jenny-s51: Interesting - in the old console it's impossible to save a provider unless the priority field is set to be strictly an integer, therefore I didn't think `Number()` would cause any problems.

In the new console I updated the code to always set the default priority to `"0"` as it does in the old console. Could there still be instances when it is `undefined` or `string[]`?

src/util.ts
jenny-s51: Great suggestion - also TIL 🐍

src/realm-settings/KeysProvidersTab.tsx
jenny-s51: @edewit updated to include a success message

src/realm-settings/KeysProvidersTab.tsx
jenny-s51: good idea

src/realm-settings/RSAModal.tsx
jenny-s51: Yes, the old console sets the default priority for every new provider to 0

src/realm-settings/RSAGeneratedModal.tsx
jenny-s51: Yes, the old console sets the default priority for every new provider to 0

jenny-s51 (commented):
@jonkoops thank you for your thorough review! Your feedback and suggestions are always helpful and much appreciated

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -149,6 +149,15 @@ const RealmSettingsHeader = ({
   );
 };
 
+const sortByPriority = (components: ComponentRepresentation[]) => {
+  const sortedComponents = [...components].sort(
+    // sort in descending order
+    (a, b) => Number(b.config?.priority) - Number(a?.config?.priority || 0)
+  );
```
jonkoops, line 156: I think we can catch all the edge cases here by casting the values to a `number` first and then checking if those values are `NaN` (defaulting to 0 if they are). Since both `Number([])` and `Number(undefined)` will produce `NaN` only 'good' values will pass through here.

I think the type is actually wrong and this value specifically would always be of type `string`, but it's probably good to cover possible edge cases if we are not entirely sure.
```suggestion
  const sortedComponents = [...components].sort((a, b) => {
    const priorityA = Number(a.config?.priority);
    const priorityB = Number(b.config?.priority);

    return (
      (!isNaN(priorityB) ? priorityB : 0) - (!isNaN(priorityA) ? priorityA : 0)
    );
  });
```

jonkoops (changes requested):
PR is looking good @jenny-s51, I think with this last change we can get it merged.

src/realm-settings/RealmSettingsSection.tsx
jenny-s51: defaulting to 0 here is a good idea - thank you @jonkoops

jonkoops (approved):
LGTM!

mfrances17 (approved):
LGTM, nice work. The ordering now works as expected. One thing I will mention, is that when you switch tabs to the Key list, the new ordering is not reflected unless you manually refresh. That may or may not be a problem we want to address at some point.

PR #954, opened 2021-08-09
added test for adding service account role

edewit:
fixing: #494

### DIFF
--- a/cypress/support/pages/admin_console/manage/RoleMappingTab.ts
+++ b/cypress/support/pages/admin_console/manage/RoleMappingTab.ts
@@ -20,7 +20,7 @@ export default class RoleMappingTab {
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


cypress/support/pages/admin_console/manage/RoleMappingTab.ts
```diff
@@ -19,8 +20,8 @@ export default class RoleMappingTab {
     return this;
   }
 
-  clickAssignRole() {
-    cy.getId(this.assignRole).click();
+  clickAssignRole(notEmpty: boolean = true) {
```
jonkoops, line 23: No need to specify a type here as the TypeScript compiler can infer the type from the assignment.
```suggestion
  clickAssignRole(notEmpty = true) {
```

jonkoops (changes requested):

jonkoops (approved):

Issue #405, opened 2021-02-28
Fix Access Token Lifespan


ssilvert:
Access Token Lifespan does not allow -1 as a value.  This should behave the same as the old console as noted in the PR:
https://github.com/keycloak/keycloak-admin-ui/pull/373#issuecomment-786128188

edewit:
Didn't we say that the `-1` was strange and even you didn't know if it's the same as empty. So the more user friendly way would be to allow for blank as in not set?

ssilvert:
@edewit Should this one be closed?

jonkoops:
@yih-wang @xianli123 What is the UX opinion on this? I can create a fix that either allows `-1` or we can come up with a nicer solution, perhaps a toggle or something?

yih-wang:
@ssilvert Can you remind me that what's the meaning when the value is "-1"? I totally forgot that...

jonkoops:
> @ssilvert Can you remind me that what's the meaning when the value is "-1"? I totally forgot that...

If a value of `-1` is provided here it means that the token will never expire. Seems quite user unfriendly, so it would be nice to have some UI around it. For example a checkbox that says 'Token never expires' which hides the input or disables it.

yih-wang:
@jonkoops Do you know the default value of this field? Now the default is empty, I'm also not sure what it means...

jonkoops:
That's a good question, I don't know exactly what the default is here when the field is empty.

yih-wang:
@jonkoops For the pattern, we can do something like this
![image](https://user-images.githubusercontent.com/23331885/128687246-9b674aff-920e-4578-9715-3e1476953656.png)

edewit:
@yih-wang the default it empty, guess nobody knows what that means... but my guess it's the same as `-1` I'll add the dropdown you suggested empty and `-1`

yih-wang:
@edewit There're a lot of fields in the console use empty as the default status. They're so confused. @ssilvert Is it possible to change that to some specific values? Or at least clarify what the empty means in the tooltip.

edewit:
@yih-wang I can make this into a general component and apply it everywhere it makes sense? Or are there other places where the text "Expires in" doesn't make sense?

yih-wang:
[...]

PR #955, opened 2021-08-09
added component to set lifespan of token

edewit:
fixes: #405

### DIFF
--- a/src/clients/advanced/TokenLifespan.tsx
+++ b/src/clients/advanced/TokenLifespan.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { Control, Controller } from "react-hook-form";
+import { Control, Controller, FieldValues } from "react-hook-form";
 import { useTranslation } from "react-i18next";
 import {
   FormGroup,
@@ -20,7 +20,7 @@ type TokenLifespanProps = {
   id: string;
   name: string;
   defaultValue: string;
-  control: Control<Record<string, any>>;
+  control: Control<FieldValues>;
   units?: Unit[];
 };
 
@@ -59,7 +59,6 @@ export const TokenLifespan = ({
           <Split hasGutter>
             <SplitItem>
               <Select
-                toggleId={`toggle-${id}`}
                 variant={SelectVariant.single}
                 onToggle={(isExpanded) => setOpen(isExpanded)}
                 isOpen={open}
@@ -68,19 +67,17 @@ export const TokenLifespan = ({
                   setOpen(false);
                 }}
                 selections={[
-                  value === "" || value === -1 ? t(never) : t(expires),
+                  typeof value === "number" && value !== -1
+                    ? t(expires)
+                    : t(never),
                 ]}
               >
-                <SelectOption key={never} value={-1}>
-                  {t(never)}
-                </SelectOption>
-                <SelectOption key={expires} value={60}>
-                  {t(expires)}
-                </SelectOption>
+                <SelectOption value={-1}>{t(never)}</SelectOption>
+                <SelectOption value={60}>{t(expires)}</SelectOption>
               </Select>
             </SplitItem>
             <SplitItem>
-              {value !== "" && value !== -1 && (
+              {typeof value === "number" && value !== -1 && (
                 <TimeSelector units={units} value={value} onChange={onChange} />
               )}
             </SplitItem>
--- a/src/clients/advanced/TokenLifespan.tsx
+++ b/src/clients/advanced/TokenLifespan.tsx
@@ -1,0 +1,92 @@
+import React, { useState } from "react";
+import { Control, Controller } from "react-hook-form";
+import { useTranslation } from "react-i18next";
+import {
+  FormGroup,
+  Select,
+  SelectOption,
+  SelectVariant,
+  Split,
+  SplitItem,
+} from "@patternfly/react-core";
+
+import {
+  TimeSelector,
+  Unit,
+} from "../../components/time-selector/TimeSelector";
+import { HelpItem } from "../../components/help-enabler/HelpItem";
+
+type TokenLifespanProps = {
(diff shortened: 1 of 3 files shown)


src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <SelectOption key={never} value={-1}>
+                  {t(never)}
+                </SelectOption>
+                <SelectOption key={expires} value={60}>
+                  {t(expires)}
+                </SelectOption>
+              </Select>
+            </SplitItem>
+            <SplitItem>
+              {value !== "" && value !== -1 && (
```
jonkoops, line 83: Looks like `TimeSelector` accepts a `value` of type `number` only. We can narrow down the type in the condition by using `typeof`, which will allow the TypeScript compiler to infer the type in this branch.
```suggestion
              {typeof value === "number" && value !== -1 && (
```

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  setOpen(false);
+                }}
+                selections={[
+                  value === "" || value === -1 ? t(never) : t(expires),
+                ]}
+              >
+                <SelectOption key={never} value={-1}>
+                  {t(never)}
+                </SelectOption>
+                <SelectOption key={expires} value={60}>
```
jonkoops, line 77: I don't think a `key` prop is required here, as the children are static.

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+                isOpen={open}
+                onSelect={(_, value) => {
+                  onChange(value);
+                  setOpen(false);
+                }}
+                selections={[
+                  value === "" || value === -1 ? t(never) : t(expires),
+                ]}
+              >
+                <SelectOption key={never} value={-1}>
```
jonkoops, line 74: I don't think a `key` prop is required here, as the children are static.

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+                variant={SelectVariant.single}
+                onToggle={(isExpanded) => setOpen(isExpanded)}
+                isOpen={open}
+                onSelect={(_, value) => {
+                  onChange(value);
+                  setOpen(false);
+                }}
+                selections={[
+                  value === "" || value === -1 ? t(never) : t(expires),
+                ]}
```
jonkoops, line 72: We can narrow down the type in the condition by using `typeof`, which will allow the TypeScript compiler to infer the type.

```suggestion
                selections={[
                  typeof value === "number" && value !== -1
                    ? t(expires)
                    : t(never),
                ]}
```

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+    >
+      <Controller
+        name={name}
+        defaultValue={defaultValue}
+        control={control}
+        render={({ onChange, value }) => (
+          <Split hasGutter>
+            <SplitItem>
+              <Select
+                toggleId={`toggle-${id}`}
```
jonkoops, line 62: Do we need to specify the `toggleId` prop? I think the component generates it's own ids internally if we do not specify it.

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+  Unit,
+} from "../../components/time-selector/TimeSelector";
+import { HelpItem } from "../../components/help-enabler/HelpItem";
+
+type TokenLifespanProps = {
+  id: string;
+  name: string;
+  defaultValue: string;
+  control: Control<Record<string, any>>;
+  units?: Unit[];
```
jonkoops, line 24: The `units` prop is always specified, so I think it can be made required.
```suggestion
  units: Unit[];
```

src/clients/advanced/TokenLifespan.tsx
```diff
@@ -0,0 +1,92 @@
...
+  TimeSelector,
+  Unit,
+} from "../../components/time-selector/TimeSelector";
+import { HelpItem } from "../../components/help-enabler/HelpItem";
+
+type TokenLifespanProps = {
+  id: string;
+  name: string;
+  defaultValue: string;
+  control: Control<Record<string, any>>;
```
jonkoops, line 23:
```suggestion
  control: Control<FieldValues>;
```

Should give you the same type, closer matching the interface from React Form Hook.

jonkoops (changes requested):

src/clients/advanced/TokenLifespan.tsx
edewit: They are needed

jonkoops: Why?

src/clients/advanced/TokenLifespan.tsx
edewit: but it's optional for the `TimeSelector`

jonkoops: Sure, but all components that use this one provide them. IMHO it's best to be as strict as possible out of the box, you can always loosen the type when needed. This prevents logic from being introduced that handles possible empty values, when if fact this logic is redundant.

edewit: if you leave it empty you get all the units `["seconds", "minutes", "hours", "days"]`

jonkoops: Sure, but if someone ever makes logic inside of this component and touches these props they have to take into account it could be empty.

jonkoops (approved):

jonkoops:
LGTM :+1:

PR #967, opened 2021-08-10
changed render of link so search can find the text

edewit:
fixes: #965

### DIFF
--- a/src/realm-roles/RolesList.tsx
+++ b/src/realm-roles/RolesList.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { ReactNode, useState } from "react";
 import { Link, useHistory, useRouteMatch } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import { AlertVariant, Button, ButtonVariant } from "@patternfly/react-core";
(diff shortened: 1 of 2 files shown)


src/realm-roles/RolesList.tsx
```diff
@@ -34,11 +34,17 @@ type RolesListProps = {
...
 };
 
-const RoleLink = ({ role }: { role: RoleRepresentation }) => {
+const RoleLink = ({
+  children,
+  role,
+}: {
+  role: RoleRepresentation;
+  children: ReactNode;
+}) => {
```
jonkoops, line 43: You can use the `FunctionComponent` type here from React, which already includes the correct type for `children`.

```suggestion
type RoleLinkProps = {
  role: RoleRepresentation;
};

const RoleLink: FunctionComponent<RoleLinkProps> = ({ children, role }) => {
```

src/realm-roles/RolesList.tsx
```diff
@@ -69,7 +75,9 @@ export const RolesList = ({
 
   const RoleDetailLink = (role: RoleRepresentation) => (
     <>
-      <RoleLink role={role} />
+      <RoleLink role={role}>
+        <>{role.name}</>
```
jonkoops, line 79: Why do we need a fragment here? I believe you can render the children directly.
```suggestion
        {role.name}
```

jonkoops (changes requested):

jonkoops (approved):

PR #987, opened 2021-08-11
added user in role for client tab

edewit:
fixes: #984


src/realm-roles/UsersInRoleTab.tsx
```diff
@@ -9,24 +9,34 @@ import { KeycloakDataTable } from "../components/table-toolbar/KeycloakDataTable
...
+        max,
+      });
+    } else {
+      usersWithRole = await adminClient.roles.findUsersWithRole({
+        name: role.name!,
+        first,
+        max,
+      });
+    }
     return usersWithRole || [];
```
jonkoops, line 40: I think this code can be shortened by directly returning the result.
```suggestion

    if (role.clientRole) {
      return adminClient.clients.findUsersWithRole({
        roleName: role.name!,
        id: clientId,
        first,
        max,
      });
    }

    return adminClient.roles.findUsersWithRole({
      name: role.name!,
      first,
      max,
    });
```

jonkoops (changes requested):

jonkoops (approved):

Issue #971, opened 2021-08-10
[UX Review] Remove the eye icon of the Bind credentials after creating an Ldap provider


yih-wang:
Location: User federation -> Ldap provider -> Settings 

The eye icon makes no sense at this time as the users are not permitted to see the plain text.
![image](https://user-images.githubusercontent.com/23331885/128820898-57281c3f-7628-47e0-8d46-a3fb7342fa04.png)
![image](https://user-images.githubusercontent.com/23331885/128820917-44932e4d-b756-495f-bada-5146d7dfd3fe.png)

PR #991, opened 2021-08-11
change to TextInput when editing 

edewit:
Because keycloak doesn't store the password as clear text and returns the password as '***' it doesn't make sense to reveal the text

fixes: #971

### DIFF
--- a/src/components/password-input/PasswordInput.tsx
+++ b/src/components/password-input/PasswordInput.tsx
@@ -8,7 +8,15 @@ import {
 } from "@patternfly/react-core";
 import { EyeIcon, EyeSlashIcon } from "@patternfly/react-icons";
 
-const PasswordInputBase = ({ innerRef, ...rest }: TextInputProps) => {
+type PasswordInputProps = TextInputProps & {
+  hasReveal?: boolean;
+};
+
+const PasswordInputBase = ({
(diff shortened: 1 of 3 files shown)


jonkoops (approved):

jonkoops:
Since this is a UX review ticker I think that someone from the UXD team will have to take a look at it before merging. @xianli123 @yih-wang could you take a look at it.

mfrances17 (changes requested):
Seems like it would make more sense to add this change to the PasswordInput component itself, instead of burying a new component within LDAP settings... what do you think?

yih-wang:
LGTM

edewit:
@mfrances17 all I do is change the password input field out for a normal text field when it's in edit mode. Are you suggesting to add an edit mode property to the `PasswordInput` field itself?

mfrances17:
@edewit yeah i figured since we've already create a custom password component, it could just be one more prop we add to it. another quicker but less desirable option would be to just lift the new PasswordComponent out of ldap settings and define it as a separate keycloak component in \Components. but then we'd have two specialized keycloak password components. just a suggestion.

edewit:
@mfrances17 done

jonkoops (approved):

PR #992, opened 2021-08-11
Prefer optional chaining over validating references

jonkoops:
Use optional chaining in all places over validating references. Introduce an ESLint rule to enforce this.

## Motivation
A common review comment is to use optional chaining over traditional reference validation (e.g. `foo && foo.bar`). To save time on these trivial review comments it's easier to enforce this as an ESLint rule so it can be caught early in development.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,8 @@ module.exports = {
     },
   },
   rules: {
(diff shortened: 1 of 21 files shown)


mfrances17 (approved):
LGTM

Issue #692, opened 2021-06-10
[UX Review] Groups - Global search


yih-wang:
These issues haven't been addressed yet. @edewit 

1. A breadcrumb is missing so I can't find a way to return to the group list. Reference the design here - https://marvelapp.com/prototype/228971ig/screen/79872064
![image](https://user-images.githubusercontent.com/23331885/121533726-25211500-ca33-11eb-86ff-97339270e4a9.png)

2. This page needs to be updated to align with the design - https://marvelapp.com/prototype/228971ig/screen/79872065
![image](https://user-images.githubusercontent.com/23331885/121534013-69141a00-ca33-11eb-9c03-1bb1de7bbeaa.png)

jonkoops:
@yih-wang I am taking a look at these issues. For #1 I cannot reproduce the issue, perhaps it has been fixed in the meantime? Could you give me a reproducible scenario?

For #2, what parts of the design are missing exactly? I see that the path is no longer displayed at all for example.

yih-wang:
@jonkoops 

> For nr. 1 I cannot reproduce the issue, perhaps it has been fixed in the meantime? Could you give me a reproducible scenario?

Yes, I think it's been fixed, but the breadcrumb is still not correct. It should be Groups > Search group (align with the option name). Also I'm wondering whether the header should also be aligned with the option name - that is what we usually do. So the Search for groups can be changed to Search group. What do you think?
![image](https://user-images.githubusercontent.com/23331885/128658648-ccf1cc22-49a5-4b86-87ed-4dd0541c6433.png)

> For nr. 2, what parts of the design are missing exactly? I see that the path is no longer displayed at all for example.

I've checked the latest version of the search result screen. The padding is not correct. So the comparison between the search screen of the Groups and Users.
![image](https://user-images.githubusercontent.com/23331885/128658930-f9345f0c-b8e9-4bf1-ab30-df7d3245940a.png)
![image](https://user-images.githubusercontent.com/23331885/128659034-9c1f7c37-971f-4147-826a-8e76af91709d.png)

A new issue I found is after I search for something, a new level is added to the breadcrumb, which is not necessary. In the screenshot above, the breadcrumb should still be "Groups > Search group" instead of "Search > Search groups > Group details"

PR #998, opened 2021-08-12
fixed breadcrumb for search and seach box distance

edewit:
fixes: #692

### DIFF
--- a/src/components/bread-crumb/GroupBreadCrumbs.tsx
+++ b/src/components/bread-crumb/GroupBreadCrumbs.tsx
@@ -22,6 +22,7 @@ export const GroupBreadCrumbs = () => {
     });
   }, [history]);
 
+  const isLastGroup = (index: number) => subGroups.length - 1 === index;
   return (
     <>
       {subGroups.length !== 0 && (
@@ -31,7 +32,7 @@ export const GroupBreadCrumbs = () => {
           </BreadcrumbItem>
           {subGroups.map((group, i) => (
             <BreadcrumbItem key={i} isActive={subGroups.length - 1 === i}>
-              {subGroups.length - 1 !== i && (
+              {!isLastGroup(i) && (
                 <Link
                   to={location.pathname.substr(
                     0,
@@ -42,7 +43,10 @@ export const GroupBreadCrumbs = () => {
                   {group.name}
                 </Link>
               )}
-              {subGroups.length - 1 === i && <>{t("groups:groupDetails")}</>}
+              {isLastGroup(i) && group.id !== "search" && (
+                <>{t("groups:groupDetails")}</>
+              )}
+              {isLastGroup(i) && group.id === "search" && <>{group.name}</>}
             </BreadcrumbItem>
           ))}
         </Breadcrumb>
--- a/src/components/bread-crumb/GroupBreadCrumbs.tsx
+++ b/src/components/bread-crumb/GroupBreadCrumbs.tsx
@@ -22,7 +22,6 @@ export const GroupBreadCrumbs = () => {
     });
   }, [history]);
 
-  const isLastGroup = (index: number) => subGroups.length - 1 === index;
   return (
     <>
       {subGroups.length !== 0 && (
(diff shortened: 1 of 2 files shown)


src/components/bread-crumb/GroupBreadCrumbs.tsx
```diff
@@ -31,7 +32,7 @@ export const GroupBreadCrumbs = () => {
           </BreadcrumbItem>
           {subGroups.map((group, i) => (
             <BreadcrumbItem key={i} isActive={subGroups.length - 1 === i}>
```
jonkoops, line 34: Perhaps we can use the group id for the key instead?
```suggestion
            <BreadcrumbItem key={group.id} isActive={subGroups.length - 1 === i}>
```

jonkoops, line 34: I think we can use the `isLastGroup()` function here as well.
```suggestion
            <BreadcrumbItem key={i} isActive={isLastGroup(i)}>
```

src/components/bread-crumb/GroupBreadCrumbs.tsx
```diff
@@ -42,7 +43,10 @@ export const GroupBreadCrumbs = () => {
                   {group.name}
                 </Link>
               )}
-              {subGroups.length - 1 === i && <>{t("groups:groupDetails")}</>}
+              {isLastGroup(i) && group.id !== "search" && (
+                <>{t("groups:groupDetails")}</>
+              )}
+              {isLastGroup(i) && group.id === "search" && <>{group.name}</>}
```
jonkoops, line 49: There is no need to use a fragment here, as a string is also a valid JSX element. We can also use a ternary operation to reduce the amount of calls to `isLastGroup()`.

```suggestion
              {isLastGroup(i) &&
                (group.id === "search" ? group.name : t("groups:groupDetails"))}
```

src/components/bread-crumb/GroupBreadCrumbs.tsx
```diff
@@ -22,6 +22,7 @@ export const GroupBreadCrumbs = () => {
     });
   }, [history]);
 
+  const isLastGroup = (index: number) => subGroups.length - 1 === index;
```
jonkoops, line 25: Instead of defining a function which we have to call every time we need a result, we can also define a variable in the function body of the `map()` over the subgroups to get the value once.

```tsx
{subGroups.map((group, i) => {
  const isLastGroup = i === subGroups.length - 1;

  return (
    <BreadcrumbItem key={group.id} isActive={isLastGroup}>
      ...
    </BreadcrumbItem>
  );
})}
```

src/groups/SearchGroups.tsx
```diff
@@ -55,16 +58,19 @@ export const SearchGroups = () => {
...
-          {group.name}
-        </Link>
-      </>
-    );
-  };
+  const GroupNameCell = (group: SearchGroup) => (
+    <>
+      <Link
+        key={group.id}
+        to={`/${realm}/groups/search/${group.link}`}
```
jonkoops, line 65: Any chance we could use one of the link functions from the route definition here?

src/groups/SearchGroups.tsx
```diff
@@ -55,16 +58,19 @@ export const SearchGroups = () => {
...
+        onClick={() =>
+          setSubGroups([{ name: t("searchGroups"), id: "search" }, group])
```
jonkoops, line 67: Interesting, we're setting an internal state that is derived from the URL state as well? Can't we only derive this state from the path entirely?

src/groups/SearchGroups.tsx
```diff
@@ -108,44 +114,43 @@ export const SearchGroups = () => {
...
-                {term}
-              </Chip>
-            ))}
-          </ChipGroup>
-        </Form>
         <KeycloakDataTable
           key={key}
+          isSearching
+          toolbarItem={
+            <>
```
jonkoops, line 123: Why do we need this fragment? Looks like there is only a single child element.

jonkoops (changes requested):

src/groups/SearchGroups.tsx
edewit: otherwise it doesn't work, I don't know why

src/groups/SearchGroups.tsx
edewit: we could if we had a way to put the group name and it's ID on the url.

src/groups/SearchGroups.tsx
edewit: don't see how to do that quickly

jonkoops: Let's skip it for now then.

src/groups/SearchGroups.tsx
jonkoops: Yeah, this seems like an improvement we can do at a later point.

src/components/bread-crumb/GroupBreadCrumbs.tsx
```diff
@@ -29,22 +29,28 @@ export const GroupBreadCrumbs = () => {
...
-                  {group.name}
-                </Link>
-              )}
-              {subGroups.length - 1 === i && <>{t("groups:groupDetails")}</>}
-            </BreadcrumbItem>
-          ))}
+          {subGroups.map((group, i) => {
+            const isLastGroup = i === subGroups.length - 1;
+            return (
+              <BreadcrumbItem key={i} isActive={isLastGroup}>
```
jonkoops, line 35: Can we use the group id here for the key prop?
```suggestion
              <BreadcrumbItem key={group.id} isActive={isLastGroup}>
```

jonkoops (changes requested):

yih-wang (changes requested):
1. Enter (on keyboard) doesn't work when I want to trigger the search
2. The search bar and the chips are next to each other. I think there should be an 8px (or around) space between them. I haven't found a code example in Patternfly, only a screenshot in the guidelines like this 
image
3. The search bar should be left-aligned with the header (like the main page of the Groups)

yih-wang (approved):
LGTM

jonkoops (approved):

Issue #941, opened 2021-08-05
Implement ability to rename a realm


jonkoops:
In the old admin console it's possible to rename the realm. However in our version this field has been explicitly disabled.

PR #1001, opened 2021-08-12
made realm name editable

edewit:
fixes: #941

### DIFF
--- a/src/context/realm-context/RealmContext.tsx
+++ b/src/context/realm-context/RealmContext.tsx
@@ -45,14 +45,10 @@ export const RealmContextProvider: FunctionComponent = ({ children }) => {
   useEffect(() => adminClient.setConfig({ realmName: realm }), [realm]);
 
   const set = (realm: string) => {
-    if (
-      realms.length === 0 ||
-      realms.findIndex((r) => r.realm == realm) !== -1
-    ) {
-      recentUsed.setRecentUsed(realm);
-      setRealm(realm);
-    }
+    recentUsed.setRecentUsed(realm);
+    setRealm(realm);
   };
+
   return (
     <RealmContext.Provider
       value={{
@@ -60,6 +56,9 @@ export const RealmContextProvider: FunctionComponent = ({ children }) => {
         setRealm: set,
         realms,
         refresh: async () => {
+          //this is needed otherwise the realm find function will not return
+          //new or renamed realms because of the cached realms in the token (perhaps?)
+          await adminClient.keycloak?.updateToken(Number.MAX_VALUE);
           const list = await adminClient.realms.find();
           updateRealmsList(list);
         },
--- a/src/realm-settings/RealmSettingsSection.tsx
+++ b/src/realm-settings/RealmSettingsSection.tsx
@@ -6,6 +6,7 @@ import {
   DropdownItem,
   DropdownSeparator,
   PageSection,
+  Spinner,
   Tab,
   Tabs,
   TabTitleText,
@@ -26,7 +27,7 @@ import { useRealm } from "../context/realm-context/RealmContext";
 import { useServerInfo } from "../context/server-info/ServerInfoProvider";
 import { LocalizationTab } from "./LocalizationTab";
 import { useWhoAmI } from "../context/whoami/WhoAmI";
-import { KEY_PROVIDER_TYPE, toUpperCase } from "../util";
+import { convertToFormValues, KEY_PROVIDER_TYPE, toUpperCase } from "../util";
 import { RealmSettingsEmailTab } from "./EmailTab";
 import { EventsTab } from "./event-config/EventsTab";
 import { RealmSettingsGeneralTab } from "./GeneralTab";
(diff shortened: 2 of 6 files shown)


src/context/realm-context/RealmContext.tsx
```diff
@@ -45,21 +45,20 @@ export const RealmContextProvider: FunctionComponent = ({ children }) => {
...
   return (
     <RealmContext.Provider
       value={{
         realm,
         setRealm: set,
         realms,
         refresh: async () => {
+          //this is needed otherwise the realm find function will not return
+          //new or renamed realms because of the cached realms in the token (perhaps?)
+          await adminClient.keycloak?.updateToken(Number.MAX_VALUE);
```
jonkoops, line 61: That's interesting :thinking:

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -272,35 +297,33 @@ export const RealmSettingsSection = () => {
...
-                  >
-                    <KeysProvidersTab
-                      realmComponents={realmComponents}
-                      keyProviderComponentTypes={kpComponentTypes}
-                      refresh={refresh}
-                    />
-                  </Tab>
-                </Tabs>
-              )}
+                  <KeysListTab realmComponents={realmComponents!} />
```
jonkoops, line 311: If the aforementioned statement is changed there is no need to use a non-null assertion here.
```suggestion
                  <KeysListTab realmComponents={realmComponents} />
```

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -272,35 +297,33 @@ export const RealmSettingsSection = () => {
...
+                </Tab>
+                <Tab
+                  id="providers"
+                  data-testid="rs-providers-tab"
+                  aria-label="rs-providers-tab"
+                  eventKey={1}
+                  title={<TabTitleText>{t("providers")}</TabTitleText>}
+                >
+                  <KeysProvidersTab
+                    realmComponents={realmComponents!}
```
jonkoops, line 321: If the aforementioned statement is changed there is no need to use a non-null assertion here.
```suggestion
                    realmComponents={realmComponents}
```

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -317,24 +340,20 @@ export const RealmSettingsSection = () => {
...
-                  reset={() => resetForm(realm)}
-                  realm={realm}
-                />
-              )}
+              <LocalizationTab
+                key={key}
+                refresh={refresh}
+                save={save}
+                reset={() => resetForm(realm)}
+                realm={realm!}
```
jonkoops, line 348: If the aforementioned statement is changed there is no need to use a non-null assertion here.
```suggestion
                realm={realm}
```

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -354,7 +373,10 @@ export const RealmSettingsSection = () => {
               aria-label="tokens-tab"
               title={<TabTitleText>{t("realm-settings:tokens")}</TabTitleText>}
             >
-              <RealmSettingsTokensTab reset={() => resetForm(realm)} />
+              <RealmSettingsTokensTab
+                realm={realm!}
```
jonkoops, line 377: If the aforementioned statement is changed there is no need to use a non-null assertion here.
```suggestion
                realm={realm}
```

src/realm-settings/TokensTab.tsx
```diff
@@ -70,36 +69,15 @@ export const RealmSettingsTokensTab = ({
     javaKeystoreAlgOptions!
   );
 
-  const form = useForm<RealmRepresentation>();
-  const { control } = useFormContext();
+  const form = useFormContext<RealmRepresentation>();
+  const control = form.control;
```
jonkoops, line 73: You can use destructuring to shorten this statement. In fact we might actually want to create an ESLint rule for this (that's for another time).

```suggestion
  const { control } = form;
```

jonkoops (changes requested):

src/context/realm-context/RealmContext.tsx
edewit: The token contains realm and access information if we don't refresh it after a realm has been created you get errors

jonkoops (approved):

ssilvert:
13 Realm Settings tests are failing.  I'm re-running the tests but that probably won't fix the problem.

jonkoops:
@stianst This is a known issue, looks like some of the review suggestions have broken the tests.

edewit:
@ssilvert I thought I fixed it yesterday, but guess it's not enough I'll have another look

src/realm-settings/TokensTab.tsx
```diff
@@ -70,36 +69,15 @@ export const RealmSettingsTokensTab = ({
...
   const offlineSessionMaxEnabled = useWatch({
     control,
     name: "offlineSessionMaxLifespanEnabled",
     defaultValue: realm?.offlineSessionMaxLifespanEnabled,
   });
 
-  const setupForm = (realm: RealmRepresentation) => {
-    const { ...formValues } = realm;
-    form.reset(formValues);
-    Object.entries(realm).map((entry) => {
```
jenny-s51, line 85: This was needed here in order to populate the last four fields of the form

jenny-s51 (commented):
Looks and works great @edewit! However removing the `setupForm` from the TokensTab causes the last four fields of the tokens settings to not be populated with the data from the server

mfrances17 (approved):
LGTM - code looks fine, and testing worked as expected

src/realm-settings/TokensTab.tsx
edewit: I moved it to the `RealmSettingsSection.tsx`

jonkoops:
@edewit Looks like the tests are still failing? I've ran it a couple of times, but always seems to fail.

edewit:
yes it seems that it doesn't save the `IdP account email verification` token is not saved, but it's also broken on master

jenny-s51 (approved):
Looks great!

jonkoops (approved):

PR #1015, opened 2021-08-17
Initial version of config settings for executions

edewit:
still needs to render different controls based on the types

## Verification Steps
Go to authentication section click on one of the flows some executions can have settings indicated by a cog wheel on the right side
clicking it will open a modal dialog to add the settings or clear / update existing onces.

TODO is rendering different inputs based on the type of the field
The fields are dynamic based off a backend call

### DIFF
--- a/src/authentication/components/ExecutionConfigModal.tsx
+++ b/src/authentication/components/ExecutionConfigModal.tsx
@@ -27,11 +27,13 @@ type ExecutionConfigModalForm = {
   [index: string]: string;
 };
 
+type ExecutionConfigModalProps = {
+  execution: ExpandableExecution;
+};
+
 export const ExecutionConfigModal = ({
   execution,
-}: {
-  execution: ExpandableExecution;
-}) => {
+}: ExecutionConfigModalProps) => {
   const { t } = useTranslation("authentication");
   const adminClient = useAdminClient();
   const { addAlert, addError } = useAlerts();
--- a/src/authentication/components/ExecutionConfigModal.tsx
+++ b/src/authentication/components/ExecutionConfigModal.tsx
@@ -43,7 +43,7 @@ export const ExecutionConfigModal = ({
     configDescription: AuthenticatorConfigInfoRepresentation,
     config?: AuthenticatorConfigRepresentation
   ) => {
-    configDescription.properties.map(
+    configDescription.properties!.map(
       (property: ConfigPropertyRepresentation) => {
         setValue(
           property.name!,
@@ -79,8 +79,8 @@ export const ExecutionConfigModal = ({
   );
 
   useEffect(() => {
-    if (configDescription) setupForm(configDescription, config), [open];
-  });
+    if (configDescription) setupForm(configDescription, config);
+  }, [show]);
 
   const save = async (changedConfig: { [index: string]: string }) => {
     try {
--- a/src/authentication/components/ExecutionConfigModal.tsx
+++ b/src/authentication/components/ExecutionConfigModal.tsx
@@ -23,6 +23,10 @@ import { useAdminClient, useFetch } from "../../context/auth/AdminClient";
 import { useAlerts } from "../../components/alert/Alerts";
 import { HelpItem } from "../../components/help-enabler/HelpItem";
 
+type ExecutionConfigModalForm = {
+  [index: string]: string;
+};
+
 export const ExecutionConfigModal = ({
   execution,
 }: {
@@ -37,7 +41,8 @@ export const ExecutionConfigModal = ({
   const [configDescription, setConfigDescription] =
     useState<AuthenticatorConfigInfoRepresentation>();
 
-  const { register, errors, setValue, handleSubmit } = useForm();
+  const { register, errors, setValue, handleSubmit } =
+    useForm<ExecutionConfigModalForm>();
 
   const setupForm = (
     configDescription: AuthenticatorConfigInfoRepresentation,
@@ -82,7 +87,7 @@ export const ExecutionConfigModal = ({
     if (configDescription) setupForm(configDescription, config);
   }, [show]);
 
-  const save = async (changedConfig: { [index: string]: string }) => {
+  const save = async (changedConfig: ExecutionConfigModalForm) => {
     try {
       if (config) {
         const newConfig = {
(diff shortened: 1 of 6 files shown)


src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+      const configDescription =
+        await adminClient.authenticationManagement.getConfigDescription({
+          providerId: execution.providerId!,
+        });
+      let config: AuthenticatorConfigRepresentation | undefined;
+      if (execution.authenticationConfig) {
+        config = await adminClient.authenticationManagement.getConfig({
+          id: execution.authenticationConfig,
+        });
+      }
```
jonkoops, line 71: You can probably use a ternary here.
```suggestion
      const config = execution.authenticationConfig
        ? await adminClient.authenticationManagement.getConfig({
            id: execution.authenticationConfig,
          })
        : undefined;
```

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+    ({ configDescription, config }) => {
+      setConfigDescription(configDescription);
+      setConfig(config);
+    },
+    []
+  );
+
+  useEffect(() => {
+    if (configDescription) setupForm(configDescription, config), [open];
+  });
```
jonkoops, line 83: This effect has no dependencies, which means it will run after each render. Is that the intended behavior?

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const [configDescription, setConfigDescription] =
+    useState<AuthenticatorConfigInfoRepresentation>();
+
+  const { register, errors, setValue, handleSubmit } = useForm();
+
+  const setupForm = (
+    configDescription: any,
+    config?: AuthenticatorConfigRepresentation
+  ) => {
+    configDescription.properties.map(
```
jonkoops, line 46: This `map()` operation does not actually get assigned to anything, perhaps you meant to do a `forEach()`?
```suggestion
    configDescription.properties.forEach(
```

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+      (property: ConfigPropertyRepresentation) => {
+        setValue(
+          property.name!,
+          config?.config?.[property.name!] || property.defaultValue || ""
```
jonkoops, line 50: Note that this will use values that are empty strings from the config. If you want to explicitly check if it is a nullish value then the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) might be a better option.
```suggestion
          config?.config?.[property.name!] ?? property.defaultValue ?? ""
```

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const { t } = useTranslation("authentication");
+  const adminClient = useAdminClient();
+  const { addAlert, addError } = useAlerts();
+
+  const [show, setShow] = useState(false);
+  const [config, setConfig] = useState<AuthenticatorConfigRepresentation>();
+  const [configDescription, setConfigDescription] =
+    useState<AuthenticatorConfigInfoRepresentation>();
+
+  const { register, errors, setValue, handleSubmit } = useForm();
```
jonkoops, line 40: You can provide a generic type here to make the fields inside this form strictly typed.
```suggestion
  const { register, errors, setValue, handleSubmit } = useForm<MyFormFields>();
```

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+    },
+    ({ configDescription, config }) => {
+      setConfigDescription(configDescription);
+      setConfig(config);
+    },
+    []
+  );
+
+  useEffect(() => {
+    if (configDescription) setupForm(configDescription, config), [open];
```
jonkoops, line 82: There seems to be an array literal here that doesn't do anything, I think it can be removed.
```suggestion
    if (configDescription) setupForm(configDescription, config);
```

jonkoops (changes requested):
Still need to look at more of the code but here is a preliminary review.

src/authentication/components/ExecutionConfigModal.tsx
edewit: I don't think this is better

src/authentication/components/ExecutionConfigModal.tsx
edewit: oops... that seems like a weird little refactor error

src/authentication/components/ExecutionConfigModal.tsx
edewit: do we really care about the difference between ?? and ||

src/authentication/components/ExecutionConfigModal.tsx
edewit: it's the useEffect dependency I fixed it

src/authentication/components/ExecutionConfigModal.tsx
jonkoops: I guess it's a matter of opinion, I prefer the ternary because it allows us to use a constant and because the type can be inferred automatically.

edewit: Yes that is true, but I feel it's a lot more difficult to read

jonkoops: I think that's valid, let's consider this resolved :ok_hand:

src/authentication/components/ExecutionConfigModal.tsx
jonkoops: Depends if you want to allow empty strings here (if they are values that could occur).

src/authentication/components/FlowHeader.tsx
```diff
@@ -20,11 +20,10 @@ export const FlowHeader = () => {
...
           dataListCells={[
             <DataListCell key="step" id="headerName">
-              <>{t("steps")}</>
-            </DataListCell>,
-            <DataListCell key="requirement">
-              <>{t("requirement")}</>
+              {t("steps")}
             </DataListCell>,
+            <DataListCell key="requirement">{t("requirement")}</DataListCell>,
+            <DataListCell key="config"></DataListCell>,
```
jonkoops, line 26: This cell `DataListCell` doesn't need to have text like the others?

jonkoops (approved):

src/authentication/components/FlowHeader.tsx
edewit: No it doesn't have a label according to the design it only contains icons buttons

PR #1023, opened 2021-08-19
Add a constant for the default locale

jonkoops:
Adds a constant for the default locale for this application. Also adds a `getLocale()` method to `WhoAmI` to get the locale preferred by the user.

## Motivation
Having a single constant for the default locale will make it easier to refactor the application. This is also required for a later PR that will take advantage of the `getLocale()` method for `WhoAmI`.

### DIFF
--- a/src/realm-settings/LocalizationTab.tsx
+++ b/src/realm-settings/LocalizationTab.tsx
@@ -27,6 +27,7 @@ import { ListEmptyState } from "../components/list-empty-state/ListEmptyState";
 import { AddMessageBundleModal } from "./AddMessageBundleModal";
 import { useAlerts } from "../components/alert/Alerts";
 import { useRealm } from "../context/realm-context/RealmContext";
+import { DEFAULT_LOCALE } from "../i18n";
 
 type LocalizationTabProps = {
   save: (realm: RealmRepresentation) => void;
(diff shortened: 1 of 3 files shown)


src/realm-settings/LocalizationTab.tsx
```diff
@@ -54,7 +55,7 @@ export const LocalizationTab = ({
   const [supportedLocalesOpen, setSupportedLocalesOpen] = useState(false);
   const [defaultLocaleOpen, setDefaultLocaleOpen] = useState(false);
   const [filterDropdownOpen, setFilterDropdownOpen] = useState(false);
-  const [selectMenuLocale, setSelectMenuLocale] = useState("en");
+  const [selectMenuLocale, setSelectMenuLocale] = useState(DEFAULT_LOCALE);
```
jonkoops, line 58: @jenny-s51 I added the default locale here to match your existing code. However do we always want to default to English here, or should it depend on the user's selected preferred locale?

mfrances17 (approved):
LGTM - code looks fine, but disclaimer i did not test changing my locale.

src/realm-settings/LocalizationTab.tsx
mfrances17: @jonkoops @jenny-s51 If we stick to our mantra of doing what the legacy keycloak did, it defaults to English not the user's locally configured locale. Not saying that's wrong or right, just that is how it works now. See here:
https://www.keycloak.org/docs/4.8/server_admin/#internationalization

jonkoops: Seems to suggest that English is the default configured locale, so it can be changed to a different language. However we need to find a way to read that out at some point. Ideally I would like this to be provided by the Keycloak server at bootstrap, just like the other environment variables.

It's out of scope for this PR, but something we'll have to look into before going live. @ssilvert what's your take on this?

PR #1029, opened 2021-08-20
instead of renaming properties use flatten and unflatten

edewit:
for keycloak config and attribute properties


edewit:
doesn't work for ldap

PR #1037, opened 2021-08-20
update users empty state

jenny-s51:
## Checklist:


src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -414,8 +414,7 @@ export function KeycloakDataTable<T>({
           {!loading &&
             (!data || data.length === 0) &&
             (search !== "" || !isSearching) &&
-            searchPlaceholderKey &&
-            !emptyState && (
```
edewit, line 418: Isn't this still needed for other screens?

src/user/UsersSection.tsx
```diff
@@ -59,10 +59,8 @@ export const UsersSection = () => {
       ]);
     },
     (response) => {
-      //should *only* list users when no user federation is configured and uses count > 100
-      setListUsers(
-        !((response[0] && response[0].length > 0) || response[1] > 100)
-      );
+      //should *only* list users when no user federation is configured
+      setListUsers(!(response[0] && response[0].length > 0));
```
edewit, line 63: don't need the users `count` query here anymore right?

src/user/UsersSection.tsx
```diff
@@ -83,15 +81,23 @@ export const UsersSection = () => {
...
+    const hasFederatedProviders = await adminClient.components.find(testParams);
+
+    const searchParam = search || "";
     if (searchParam) {
       params.search = searchParam;
       setSearch(searchParam);
     }
 
-    if (!listUsers && !searchParam) {
+    if (hasFederatedProviders.length > 0 && !searchParam) {
```
edewit, line 97: hasFederatedProviders === listUsers don't need another query here it's already been done

edewit (changes requested):

src/user/UsersSection.tsx
jenny-s51: I tried this initially but using listUsers doesn't work in the loader because the state is already set in the hook

src/components/table-toolbar/KeycloakDataTable.tsx
jenny-s51: yes, removed it to fix a bug - reverting for now and opened a separate issue #1040

src/user/UsersSection.tsx
edewit: but it used to work, it's kinda expensive to do the query two times, you could also do:

```typescript
        <KeycloakDataTable
          key={key}
          loader={listUsers ? () => Promise.resolve([]) : loader}
```

src/components/table-toolbar/KeycloakDataTable.tsx
jonkoops: This issue has been fixed by Erik, just needs a rebase to get it in.

jonkoops:
@jenny-s51 Can you rebase the PR on top of the fixes made by Erik?

jenny-s51:
@edewit fixed - thank you for your feedback and for your update to KDT. @jonkoops rebased

src/user/UsersSection.tsx
```diff
@@ -53,16 +52,12 @@ export const UsersSection = () => {
...
-      ]);
+      return adminClient.components.find(testParams);
     },
     (response) => {
-      //should *only* list users when no user federation is configured and uses count > 100
-      setListUsers(
-        !((response[0] && response[0].length > 0) || response[1] > 100)
-      );
+      //should *only* list users when no user federation is configured
+      setListUsers(!(response && response.length > 0));
```
jonkoops, line 59: Looks like the `response` is always defined according to the type, so this condition can possibly be simplified.
```suggestion
      setListUsers(!(response.length > 0));
```
Or
```suggestion
      setListUsers(response.length === 0);
```

edewit (approved):
:partying_face: great

src/user/UsersSection.tsx
```diff
@@ -179,89 +175,84 @@ export const UsersSection = () => {
...
+          searchPlaceholderKey="users:searchForUser"
+          canSelectAll
+          onSelect={(rows) => setSelectedRows([...rows])}
+          emptyState={
+            !listUsers ? (
+              <TextContent className="kc-search-users-text">
+                <Text>{t("searchForUserDescription")}</Text>
+              </TextContent>
+            ) : (
+              <>
```
jonkoops, line 192: This fragment can be removed as it only has a single child.

jonkoops (changes requested):

jonkoops (approved):

Issue #1041, opened 2021-08-23
Client Role detail screen crashes


ssilvert:
1. Go to Clients --> Client Details --> Roles
2. Create a new role
3. Click on the new role to go to client details

You will get a 404 because the system tried to do a GET with bad URL.  See `//` between clients and roles:
GET http://localhost:8180/auth/admin/realms/master/clients//roles/fasdf/users?first=0&max=11

PR #1048, opened 2021-08-24
changed to append to url, for client role details

edewit:
fixes: #1041

### DIFF
--- a/src/realm-roles/routes/RealmRole.ts
+++ b/src/realm-roles/routes/RealmRole.ts
@@ -1,5 +1,5 @@
 import type { LocationDescriptorObject } from "history";
-import { generatePath } from "react-router-dom";
+import { useRouteMatch } from "react-router-dom";
 import type { RouteDef } from "../../route-config";
 import { RealmRoleTabs } from "../RealmRoleTabs";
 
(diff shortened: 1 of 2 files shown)


src/realm-roles/routes/RealmRole.ts
```diff
@@ -24,6 +24,9 @@ export const RealmRoleRoute: RouteDef = {
 
 export const toRealmRole = (
   params: RealmRoleParams
-): LocationDescriptorObject => ({
-  pathname: generatePath(RealmRoleRoute.path, params),
-});
+): LocationDescriptorObject => {
+  const { url } = useRouteMatch();
+  return {
+    pathname: `${url}/${params.id}/details`,
```
jonkoops, line 30: This is only true if you are navigating from a route which is already on the correct path. If this function is called in any other location the result will be an invalid URL.

jonkoops (changes requested):

src/realm-roles/routes/RealmRole.ts
edewit: To be honest I don't know what the correct way is to solve it but this function should append `${id}/details` as it is also used in client details just going to realm roles is wrong

jonkoops: @edewit Let's merge this for now, can you create an issue to resolve this?

jonkoops (approved):

Issue #1045, opened 2021-08-23
Duplicate flow saves wrong name


ssilvert:
1. Go to Authentication
2. Click on `Duplicate` from the kabab menu of an existing flow such as `Browser`
3. Fill in a new name for your duplicated flow such as `My flow`.
4. The flow will be named `Copy of browser` instead of `My flow`

PR #1049, opened 2021-08-24
fixed flow name

edewit:
fixes: #1045

### DIFF
--- a/src/authentication/DuplicateFlowModal.tsx
+++ b/src/authentication/DuplicateFlowModal.tsx
@@ -46,12 +46,12 @@ export const DuplicateFlowModal = ({
     try {
       await adminClient.authenticationManagement.copyFlow({
         flow: name,
-        newName: form.name,
+        newName: form.alias,
       });
       if (form.description !== description) {
         const newFlow = (
           await adminClient.authenticationManagement.getFlows()
-        ).find((flow) => flow.alias === form.name)!;
+        ).find((flow) => flow.alias === form.alias)!;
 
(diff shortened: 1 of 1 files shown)


src/authentication/DuplicateFlowModal.tsx
```diff
@@ -46,12 +46,12 @@ export const DuplicateFlowModal = ({
...
       await adminClient.authenticationManagement.copyFlow({
         flow: name,
-        newName: form.name,
+        newName: form.alias,
       });
       if (form.description !== description) {
         const newFlow = (
           await adminClient.authenticationManagement.getFlows()
-        ).find((flow) => flow.alias === form.name)!;
+        ).find((flow) => flow.alias === form.alias)!;
```
jonkoops, line 54: We should not be using the non-null assertion operator here as the result from a `find()` operation could possibly be `undefined`.
```suggestion
        ).find((flow) => flow.alias === form.alias);
```

jonkoops (changes requested):

src/authentication/DuplicateFlowModal.tsx
edewit: how can it not be found as it's just been created 1 line above

jonkoops: Of course, but there are no guarantees.

jonkoops (approved):
Looks good enough :+1:

ssilvert (approved):
Tested and it's working now.  Approved.

PR #1056, opened 2021-08-24
Events search fixes

agagancarczyk:
## Motivation
- Fixed issue #1055
- Fixed issue #1057
- Removed redundant Divider
- Added refresh button
- Added admin search form

## Brief Description
Improved user events search.

## Verification Steps
1. Go to /events user tab and verify increased form width so that the longest event type is fully visible.
2. Select userId or event type and do a search, click the refresh button, and verify that all events display.
3. Select non-existing userId and do a search, verify that grey line display underneath the chips.
4. Select admin events tab and verify that search form exists.

## Checklist:

## Additional Notes
None


src/events/EventsSection.tsx
```diff
@@ -362,6 +366,11 @@ export const EventsSection = () => {
                   </Form>
                 </Dropdown>
               </FlexItem>
+              <FlexItem>
+                <Button onClick={refresh} data-testid="refresh-btn">
```
jonkoops, line 370: @xianli123 Why is it this overview has a refresh button but no other view in the application has this? What exactly is it's purpose?

In this case it seems more that this is a 'Reset' button rather than a 'Refresh' button.

jonkoops (approved):
LGTM, altough I have doubts about how useful the 'Refresh' button is.

src/events/EventsSection.tsx
agagancarczyk: I agree with @jonkoops . I have checked the old admin console and there is "Reset" button which when selected clears the search form and shows all (unfiltered) results.

Screenshot 2021-08-24 at 16 33 31

cc @xianli123

xianli123:
@jonkoops @agagancarczyk The "**Refresh**" button and the "**Reset**" button have different meanings. The "**Refresh**" button is used to update events in case the user misses something new. To some extent, this is a new feature. But the "**Reset**" button means clearing all the filters and then add new filters. 

In the new design, if users want to "reset" the filter, they may need to click the "X“ in the chips or directly edit the search form. Do you think if it makes sense?

jonkoops:
> In the new design, if users want to "reset" the filter, they may need to click the "X“ in the chips or directly edit the search form. Do you think if it makes sense?

Yeah, that makes sense, however the way the 'Refresh' button is implemented in this PR it will actually reset the filters that the user has activated. So that doesn't seem like it matches the designed intention.

agagancarczyk:
@xianli123 @jonkoops I can change the behaviour of the "Refresh" button, but I think I need more clarification. What should happen when the user selects "Refresh" after performing a search and having chips displayed? Should the chips disappear, search form clear and all (unfiltered) events show? Thanks

xianli123:
After clicking the Refresh button, the search results should be displayed according to the chips. Because the chips belong to the Search bar, and the Refresh button comes after the search bar. To some extent, the search results should be displayed according to the search bar namely the chips.

agagancarczyk:
> After clicking the Refresh button, the search results should be displayed according to the chips. Because the chips belong to the Search bar, and the Refresh button comes after the search bar. To some extent, the search results should be displayed according to the search bar namely the chips.

Sounds good to me. Thanks @xianli123

PR #1058, opened 2021-08-24
Identity Providers(SAML): Adds data table for SAML mappers

jenny-s51:
## Motivation

https://issues.redhat.com/browse/CLUX-223

## Brief Description

Adds data table for mappers of SAML Identity Providers

## Verification Steps

1. In the old console, go to Identity Providers -> Add Provider -> SAML v2.0
2. Save the new provider and scroll up to the Mappers tab.
3. Click the mappers tab and create a few mappers with different names and mapper types.
4. Now in the new console, go to Identity Providers and click on the SAML provider you created.
5. Click the Mappers tab and verify the data matches the table in the old console 

## Checklist:

### DIFF
--- a/src/identity-providers/add/DetailSettings.tsx
+++ b/src/identity-providers/add/DetailSettings.tsx
@@ -33,6 +33,8 @@ import { OIDCGeneralSettings } from "./OIDCGeneralSettings";
 import { SamlGeneralSettings } from "./SamlGeneralSettings";
 import { OIDCAuthentication } from "./OIDCAuthentication";
 import { ReqAuthnConstraints } from "./ReqAuthnConstraintsSettings";
+import { KeycloakDataTable } from "../../components/table-toolbar/KeycloakDataTable";
+import { ListEmptyState } from "../../components/list-empty-state/ListEmptyState";
 
 type HeaderProps = {
   onChange: (value: boolean) => void;
@@ -96,11 +98,18 @@ export const DetailSettings = () => {
   const { realm } = useRealm();
 
   useFetch(
-    () => adminClient.identityProviders.findOne({ alias: alias }),
-    (provider) => {
-      if (provider) {
-        setProvider(provider);
-        Object.entries(provider).map((entry) => setValue(entry[0], entry[1]));
+    () =>
+      Promise.all([
+        adminClient.identityProviders.findOne({ alias: alias }),
+        adminClient.identityProviders.findMappers({ alias: alias }),
+        adminClient.identityProviders.findMapperTypes({ alias: alias }),
+      ]),
+    ([fetchedProvider]) => {
+      if (fetchedProvider) {
+        setProvider({ ...provider });
+        Object.entries(fetchedProvider).map(([key, value]) =>
+          setValue(key, value)
+        );
       }
     },
     []
@@ -118,6 +127,14 @@ export const DetailSettings = () => {
     } catch (error) {
       addError("identity-providers:updateError", error);
     }
+  };
+
+  type IdPWithMapperAttributes = {
+    name: string;
+    category: string;
+    helpText: string;
+    id: string;
+    identityProviderMapper: string;
   };
 
   const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
@@ -140,6 +157,31 @@ export const DetailSettings = () => {
 
   const isOIDC = providerId.includes("oidc");
   const isSAML = providerId.includes("saml");
+
+  const loader = async () => {
+    const loaderMappers = await adminClient.identityProviders.findMappers({
+      alias,
+    });
+
+    const loaderMapperTypes =
+      await adminClient.identityProviders.findMapperTypes({ alias });
+
+    const components = loaderMappers?.map((loaderMapper) => {
+      const provider = Object.values(loaderMapperTypes).filter(
+        (loaderMapperType) => {
+          return loaderMapper.identityProviderMapper! === loaderMapperType.id!;
+        }
+      ) as unknown as IdPWithMapperAttributes[];
+
+      return {
+        ...provider[0],
+        name: loaderMapper.name,
+        type: provider[0]?.name,
+      } as IdPWithMapperAttributes;
+    });
+
+    return components;
+  };
 
   if (isOIDC) {
     sections.splice(1, 0, t("oidcSettings"));
@@ -222,7 +264,7 @@ export const DetailSettings = () => {
                       data-testid={"revert"}
                       variant="link"
                       onClick={() => {
-                        reset(provider);
+                        reset();
                       }}
                     >
                       {t("common:revert")}
@@ -230,6 +272,39 @@ export const DetailSettings = () => {
                   </ActionGroup>
                 </FormAccess>
               </ScrollForm>
+            </Tab>
+            <Tab
+              id="mappers"
+              eventKey="mappers"
+              title={<TabTitleText>{t("common:mappers")}</TabTitleText>}
+            >
+              <KeycloakDataTable
+                emptyState={
+                  <ListEmptyState
+                    message={t("identity-providers:noMappers")}
+                    instructions={t("identity-providers:noMappersInstructions")}
+                    primaryActionText={t("identity-providers:addMapper")}
+                  />
+                }
+                loader={loader}
+                isPaginated
+                ariaLabelKey="identity-providers:mappersList"
+                searchPlaceholderKey="identity-providers:searchForMapper"
+                columns={[
+                  {
+                    name: "name",
+                    displayKey: "common:name",
+                  },
+                  {
+                    name: "category",
+                    displayKey: "common:category",
+                  },
+                  {
+                    name: "type",
+                    displayKey: "common:type",
+                  },
+                ]}
+              />
             </Tab>
           </KeycloakTabs>
         </FormProvider>
--- a/src/identity-providers/add/DetailSettings.tsx
+++ b/src/identity-providers/add/DetailSettings.tsx
@@ -160,15 +160,15 @@ export const DetailSettings = () => {
     ]);
 
     const components = loaderMappers.map((loaderMapper) => {
-      const provider = Object.values(loaderMapperTypes).filter(
+      const mapperType = Object.values(loaderMapperTypes).find(
         (loaderMapperType) =>
           loaderMapper.identityProviderMapper! === loaderMapperType.id!
       );
 
       const result: IdPWithMapperAttributes = {
-        ...provider[0],
+        ...mapperType,
         name: loaderMapper.name!,
-        type: provider[0].name!,
+        type: mapperType?.name!,
       };
 
       return result;
--- a/src/identity-providers/add/DetailSettings.tsx
+++ b/src/identity-providers/add/DetailSettings.tsx
@@ -97,7 +97,7 @@ export const DetailSettings = () => {
     useParams<{ providerId: string; alias: string }>();
 
   const form = useForm<IdentityProviderRepresentation>();
-  const { handleSubmit, setValue, getValues, reset } = form;
+  const { handleSubmit, getValues, reset } = form;
 
   const adminClient = useAdminClient();
   const { addAlert, addError } = useAlerts();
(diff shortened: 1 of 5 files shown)


src/identity-providers/add/DetailSettings.tsx
```diff
@@ -95,11 +98,19 @@ export const DetailSettings = () => {
...
 
   useFetch(
-    () => adminClient.identityProviders.findOne({ alias: id }),
+    () => {
+      return Promise.all([
+        adminClient.identityProviders.findOne({ alias: id }),
+        adminClient.identityProviders.findMappers({ alias: id }),
+        adminClient.identityProviders.findMapperTypes({ alias: id }),
+      ]);
+    },
```
jonkoops, line 107: Since this is an arrow function you can return the value implicitly to shorten this function.
```suggestion
    () =>
      Promise.all([
        adminClient.identityProviders.findOne({ alias: id }),
        adminClient.identityProviders.findMappers({ alias: id }),
        adminClient.identityProviders.findMapperTypes({ alias: id }),
      ]),
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -95,11 +98,19 @@ export const DetailSettings = () => {
...
     (provider) => {
```
jonkoops, line 108: Instead of using an index to access the provider you can de-structure the array and get the fist item.
```suggestion
    ([provider]) => {
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -95,11 +98,19 @@ export const DetailSettings = () => {
...
       if (provider) {
-        setProvider(provider);
-        Object.entries(provider).map((entry) => setValue(entry[0], entry[1]));
+        setProvider({ ...provider[0] });
+        Object.entries(provider[0]).map((entry) =>
+          setValue(entry[0], entry[1])
+        );
```
jonkoops, line 113: If you destructure the entry this code can be simplified.
```suggestion
        Object.entries(provider).map(([key, value) =>
          setValue(key], value)
        );
```

I think you can even use `reset()` instead.
```suggestion
        reset(provider);
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -140,6 +159,35 @@ export const DetailSettings = () => {
...
 
+  const loader = async () => {
+    const loaderMappers = await adminClient.identityProviders.findMappers({
+      alias: id,
+    });
+
+    const loaderMapperTypes =
+      await adminClient.identityProviders.findMapperTypes({ alias: id });
+
+    const components = (loaderMappers as IdPWithMapperAttributes[])?.map(
```
jonkoops, line 170: The `loaderMappers` variable already has a type associated with it, why does it need to be cast to another type?

jonkoops: There should be no need to use an optional chain here as according to the types this is always an array.
```suggestion
    const components = (loaderMappers as IdPWithMapperAttributes[]).map(
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -140,6 +159,35 @@ export const DetailSettings = () => {
...
+      (loaderMapper: IdPWithMapperAttributes) => {
+        const provider = Object.values(loaderMapperTypes)?.filter(
```
jonkoops, line 172: `Object.values()` always returns an array, so there is no need to use an optional chain here.
```suggestion
        const provider = Object.values(loaderMapperTypes).filter(
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -140,6 +159,35 @@ export const DetailSettings = () => {
...
+          (loaderMapperType) => {
+            return (
+              loaderMapper.identityProviderMapper! === loaderMapperType?.id!
```
jonkoops, line 175: The `loaderMapperType` variable is always defined so there is no need to optional chain this.
```suggestion
              loaderMapper.identityProviderMapper! === loaderMapperType.id!
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -140,6 +159,35 @@ export const DetailSettings = () => {
...
+            );
+          }
+        ) as unknown as IdPWithMapperAttributes[];
+
+        return {
+          ...provider[0],
+          name: loaderMapper?.name,
```
jonkoops, line 182: The `loaderMapper` variable is always defined so there is no need to use optional chaining  here.
```suggestion
          name: loaderMapper.name,
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -228,6 +276,51 @@ export const DetailSettings = () => {
...
+              eventKey="mappers"
+              title={<TabTitleText>{t("common:mappers")}</TabTitleText>}
+            >
+              <KeycloakDataTable
+                emptyState={
+                  <ListEmptyState
+                    message={t("identity-providers:noMappers")}
+                    instructions={t("identity-providers:noMappersInstructions")}
+                    primaryActionText={t("identity-providers:addMapper")}
+                    //   onPrimaryAction={() => setAddDialogOpen(true)}
```
jonkoops, line 290: Commented out code.

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -228,6 +276,51 @@ export const DetailSettings = () => {
...
+                  />
+                }
+                loader={loader as any}
```
jonkoops, line 293: There should be no need to cast this.
```suggestion
                loader={loader}
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -228,6 +276,51 @@ export const DetailSettings = () => {
...
+                isPaginated
+                ariaLabelKey="clients:clientList"
+                searchPlaceholderKey="clients:searchForClient"
+                toolbarItem={
+                  <>
```
jonkoops, line 298: This fragment can be removed.

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -228,6 +276,51 @@ export const DetailSettings = () => {
...
+                    <ToolbarItem>
+                      {/* @ts-ignore */}
+                      {/* TODO: Implement Add Mapper */}
+                      {/* <Button to={toAddIdentityProviderMapperTab({ realm })}>
```
jonkoops, line 302: Why is this not implemented?

jonkoops:
@jenny-s51 Looks like the tests are failing, you might want to rebase on `master`.

ssilvert (approved):
Looks like a good start.

@jenny-s51 Why is there no Marvel for this?

jenny-s51:
@ssilvert thank you! yes good question - I asked at one of our recent team meetings and @mfrances17 @yih-wang mentioned that the new implementation should just match the old console

yih-wang:
@ssilvert The mappers of IdP follow the designs of the old console.

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -141,6 +158,31 @@ export const DetailSettings = () => {
...
+  const loader = async () => {
+    const loaderMappers = await adminClient.identityProviders.findMappers({
+      alias,
+    });
+
+    const loaderMapperTypes =
+      await adminClient.identityProviders.findMapperTypes({ alias });
+
+    const components = loaderMappers?.map((loaderMapper) => {
+      const provider = Object.values(loaderMapperTypes).filter(
```
jonkoops, line 170: I believe `Object.values` can be removed here and the result will still be the same.
```suggestion
      const provider = loaderMapperTypes.filter(
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -141,6 +158,31 @@ export const DetailSettings = () => {
...
+        (loaderMapperType) => {
+          return loaderMapper.identityProviderMapper! === loaderMapperType.id!;
+        }
+      ) as unknown as IdPWithMapperAttributes[];
```
jonkoops, line 174: I believe you could use a `Array#find()` here to get the result you are looking for (as only the first index is ever used).

```suggestion
      const mapperType = loaderMapperTypes.find(
        (loaderMapperType) =>
          loaderMapper.identityProviderMapper! === loaderMapperType.id!
      );
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -141,6 +158,31 @@ export const DetailSettings = () => {
...
+
+      return {
+        ...provider[0],
+        name: loaderMapper.name,
+        type: provider[0]?.name,
+      } as IdPWithMapperAttributes;
```
jonkoops, line 180: Casting the value to a type using `as` might allow fields to be missed in the type check at compile time. It's better to use the type annotation at the front:

```ts
const result: IdPWithMapperAttributes = {
  ...provider[0],
  name: loaderMapper.name,
  type: provider[0]?.name,
}
```

jonkoops (changes requested):

src/identity-providers/add/DetailSettings.tsx
jenny-s51: A `.filter is not a function` error gets thrown here unless `Object.values` is used

jenny-s51:
@jonkoops Thank you for your helpful and insightful feedback - I've addressed all your comments.

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -96,11 +105,15 @@ export const DetailSettings = () => {
...
+    () =>
+      Promise.resolve([
+        adminClient.identityProviders.findOne({ alias: alias }),
+      ]),
+    async ([fetchedProvider]) => {
+      if (fetchedProvider) {
+        Object.entries(fetchedProvider).map(([key, value]) =>
+          setValue(key, value)
+        );
       }
```
jonkoops, line 117: This can be reduced/improved a bit by making the following changes:

1. There is no need to wrap the call in `Promise.resolve()` as it already returns a Promise
2. There is no need to make the callback `async` as there is no `await` used inside of the function.
3. There is no need to destructure the result as it's only a single value.
4. There is no need to check the value of the result before assignment (according to the return type it's value is always present). 
5. `forEach()` should be used instead of `.map()` as there is no resulting value from the map operation.

With these changes you'll end up with the following:
```suggestion
    () => adminClient.identityProviders.findOne({ alias: alias }),
    (provider) => {
      Object.entries(provider).forEach(([key, value]) => setValue(key, value));
    },
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -141,6 +153,30 @@ export const DetailSettings = () => {
...
 
+  const loader = async () => {
+    const [loaderMappers, loaderMapperTypes] = await Promise.all([
+      adminClient.identityProviders.findMappers({ alias }),
+      adminClient.identityProviders.findMapperTypes({ alias }),
+    ]);
+
+    const components = loaderMappers.map((loaderMapper) => {
+      const mapperType = Object.values(loaderMapperTypes).find(
+        (loaderMapperType) =>
```
jonkoops, line 164: @jenny-s51 You are correct that this `Object.values()` call in required, however it looks like the Admin Client is giving us a faulty type here. We are expecting a type of `Record<string, IdentityProviderMapperRepresentation>`, but the admin client is giving us a `IdentityProviderMapperRepresentation[]` type instead.

jonkoops: I've created a PR (https://github.com/keycloak/keycloak-nodejs-admin-client/pull/248) on the Admin Client to correct this issue.

jonkoops (changes requested):

src/identity-providers/add/DetailSettings.tsx
jenny-s51: That makes sense - thank you

jonkoops (approved):
LGTM :heavy_check_mark:

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -96,11 +105,15 @@ export const DetailSettings = () => {
...
-        Object.entries(provider).map((entry) => setValue(entry[0], entry[1]));
+    () =>
+      Promise.resolve([
+        adminClient.identityProviders.findOne({ alias: alias }),
+      ]),
+    async ([fetchedProvider]) => {
+      if (fetchedProvider) {
+        Object.entries(fetchedProvider).map(([key, value]) =>
+          setValue(key, value)
+        );
```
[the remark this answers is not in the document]
jenny-s51: good call, thank you

PR #1062, opened 2021-08-25
Remove useless React fragments

jonkoops:
Change code to remove useless React fragments. Also adds an ESLint rule to enforce this.

## Motivation
This comes up as a review comment every once and a while, enforcing it as a rule will prevent some trivial review comments.


jonkoops:
Might be best to check out the diff in your IDE as Github is not indentation/syntax-aware in the diffs. Changes are exclusively removing fragments and fixing indentation.

Issue #1027, opened 2021-08-20
[UX Review] Update the toolbar of Client scopes list


xianli123:
Hi @jonkoops @mfrances17 , the client scope page and the client details page have two layouts of the toolbar for the client scope list. They have almost the same functionality, therefore we want to unify them. 

After our discussion, we decided to change the Client scopes page in the nav bar just like the picture shows below. 
<img width="796" alt="Screen Shot 2021-08-20 at 4 12 18 PM" src="https://user-images.githubusercontent.com/53931519/130202459-d5e47785-62aa-43bb-81f1-065285812a98.png">

You can see this toolbar is the same as the Setup page inside the Client scope tab on the client details page. And I created some wireframes, please check them. https://marvelapp.com/prototype/5cf7cfd/section/1437257

PR #1064, opened 2021-08-25
Make filtering of client scopes same for clients and client scope section

edewit:
fixes: #1027


src/client-scopes/ClientScopesSection.tsx
```diff
@@ -52,13 +62,28 @@ export const ClientScopesSection = () => {
     ClientScopeDefaultOptionalType[]
   >([]);
 
-  const loader = async () => {
+  const [searchType, setSearchType] = useState<SearchType>("name");
+  const [searchTypeType, setSearchTypeType] = useState<AllClientScopes>(
+    AllClientScopes.none
```
jonkoops, line 67: Enums values are supposed to start with a capital letter like [the TypeScript documentation](https://www.typescriptlang.org/docs/handbook/enums.html). But this is a separate issue from this PR.

src/client-scopes/ClientScopesSection.tsx
```diff
@@ -72,9 +97,11 @@ export const ClientScopesSection = () => {
               )
             ? ClientScope.optional
             : AllClientScopes.none,
-        };
+        } as Row;
```
jonkoops, line 100: Instead of casting this value to a different type it might be better to create a variable and use a proper type to catch type-errors.

```ts
const row: Row = {
 ...
}

return row;
```

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,188 @@
...
+  clientScopeTypesSelectOptions,
+} from "../../components/client-scope/ClientScopeTypes";
+import type { Row } from "../../clients/scopes/ClientScopes";
+
+export type SearchType = "name" | "type" | "protocol";
+export enum ProtocolType {
+  all = "all",
+  saml = "saml",
+  openid = "openid-connect",
+}
```
jonkoops, line 24: Enums values are supposed to start with a capital letter like in [the TypeScript documentation](https://www.typescriptlang.org/docs/handbook/enums.html).
```suggestion
export enum ProtocolType {
  All = "all",
  Saml = "saml",
  Openid = "openid-connect",
}
```

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,188 @@
...
+} from "@patternfly/react-core";
+import { FilterIcon } from "@patternfly/react-icons";
+
+import {
+  AllClientScopes,
+  clientScopeTypesSelectOptions,
+} from "../../components/client-scope/ClientScopeTypes";
+import type { Row } from "../../clients/scopes/ClientScopes";
+
+export type SearchType = "name" | "type" | "protocol";
```
jonkoops, line 19: Why not make this an enum as well?

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,188 @@
...
+export enum ProtocolType {
+  all = "all",
+  saml = "saml",
+  openid = "openid-connect",
+}
+type ProtocolTypeKeys = keyof typeof ProtocolType;
+
+export const nameFilter =
+  (search: string = "") =>
```
jonkoops, line 28: No need to specify the type if a default is provided from which the type can be inferred.
```suggestion
  (search = "") =>
```

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,188 @@
...
+              className="keycloak__client-scopes__searchtype"
+              onToggle={(open) => setOpen(open)}
+              isOpen={open}
+              selections={[
+                type === AllClientScopes.none
+                  ? t("common:allTypes")
+                  : t(`common:clientScope.${type}`),
+              ]}
+              onSelect={(_, value) => {
+                console.log(value);
```
jonkoops, line 144: Stray console log

jonkoops (changes requested):

ssilvert:
@edewit The test that's failing here is not one of the usual suspects.  Maybe worth a look?

edewit:
uhhh `realm_settings_test.spec.ts` deleteProviders seems to be one of the usal suspects though

mfrances17 (dismissed):
The issue requested that the layout of the toolbars look the same, but i'm seeing a 'Change type' command from the vertical ellipsis in the Client Scopes toolbar, and a 'Change type to' dropdown from the Clients > Client details > Client scopes tab.

client-scopes


client

jonkoops (approved):

edewit:
@mfrances17 now both pages have the same dropdown to change the type

jonkoops:
@edewit build is failing now.

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,165 @@
...
+      className="keycloak__client-scopes__searchtype"
+      toggle={
+        <DropdownToggle
+          id="toggle-id"
+          onToggle={(open) => setSearchToggle(open)}
+        >
+          <FilterIcon /> {t(`clientScopeSearch.${searchType}`)}
+        </DropdownToggle>
+      }
+      aria-label="Select Input"
```
jonkoops, line 82: This label needs to be translated.

src/client-scopes/details/SearchFilter.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+import {
+  AllClientScopes,
+  AllClientScopeType,
+  clientScopeTypesSelectOptions,
+} from "../../components/client-scope/ClientScopeTypes";
+import type { Row } from "../../clients/scopes/ClientScopes";
+
+export type SearchType = "name" | "type" | "protocol";
+export const Protocols = ["all", "saml", "openid-connect"];
```
jonkoops, line 21: Since this is variable it should start with a lower case.

```suggestion
export const protocols = ["all", "saml", "openid-connect"];
```

Or if you want to use a constant case:
```suggestion
export const PROTOCOLS = ["all", "saml", "openid-connect"];
```

jonkoops: Note that you can also tell the TypeScript compiler that the type is a readonly tuple. So instead of having a type of `string[]` it will become `readonly ["all", "saml", "openid-connect"]`.

This also will reflects down to `ProtocolType` which will be of type `"all" | "saml" | "openid-connect"` instead of type `string`.

```suggestion
export const protocols = ["all", "saml", "openid-connect"] as const;
```

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -427,7 +425,11 @@ export function KeycloakDataTable<T>({
           {loading && <Loading />}
         </PaginatingTableToolbar>
       )}
-      {!loading && (!data || data?.length === 0) && search === "" && emptyState}
+      {!loading &&
+        (!data || data?.length === 0) &&
+        search === "" &&
+        !isSearching &&
+        emptyState}
```
jonkoops, line 432: This statement is getting very complex. Perhaps it could be extracted into a function.

jonkoops (changes requested):

jonkoops (approved):

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: I'll add that to another PR

PR #1083, opened 2021-08-27
refactor realm settings

edewit:
also fixes: #1038


mfrances17 (approved):
LGTM, code looks good and the tokens are now saving correctly in several different combinations in testing.

PR #1094, opened 2021-08-30
add action menu to current user

edewit:


src/user/messages.ts
```diff
@@ -43,8 +43,16 @@ export default {
     notVerified: "Not verified",
     requiredUserActions: "Required user actions",
     addUser: "Add user",
+    impersonate: "Impersonate",
+    impersonateConfirm: "Impersonate user?",
+    impersonateConfirmDialog:
+      "Are you sure you want to login as this user? If this user is in the same realm with you, your current login session will be logged out before you log in as this user",
```
mfrances17, line 49: Should be: "Are you sure you want to log in..." not "Are you sure you want to login..."
Also it's a full sentence so should end in a period.

mfrances17 (changes requested):
Code looks fine, a couple string suggestions. Delete and impersonate tested ok, although other than logging you out and going to the account console, not sure exactly what impersonate is supposed to do (no description or testing steps provided).

ssilvert (changes requested):
This all looks good except "Impersonate" isn't actually working.  It takes me to the account console but I am logged out.

What should happen is that I go to the account console and it appears that I am logged in as the user I am impersonating.

edewit:
@ssilvert Right, I should have mentioned that on the PR it only works when keycloak is serving the page.

ssilvert:
> @ssilvert Right, I should have mentioned that on the PR it only works when keycloak is serving the page.

@edewit OK.  So if the plan is to fix that in a future PR I can approve this one.  Is that what you want to do?

edewit:
@ssilvert no need to fix it later. It already works just not in dev mode, only when the admin client is installed.

ssilvert:
> @ssilvert no need to fix it later. It already works just not in dev mode, only when the admin client is installed.

Oh, I see.  I'll give it a try running as a theme then.

ssilvert:
@edewit Running as a theme works as you said.  I'm ready to approve except that the test failures are not ones I recognize.  Running the tests again.

ssilvert (approved):

mfrances17 (approved):
LGTM, thanks for updating

PR #1098, opened 2021-08-30
Don't complain about Windows line endings

ssilvert:
## Motivation
Prettier setting to allow Windows users to maintain existing line endings.  Otherwise, prettier complains in VS code.  The line endings are handled by git on commit so that we should never see Windows line endings in our repo.

REVIEWER: Please test to make sure you can still do normal commits with this change.  The commit hook depends on the eslint config, which is what this PR is changing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@
 module.exports = {
   root: true,
   parserOptions: {
-    project: './tsconfig.json',
+    project: "./tsconfig.json",
   },
   env: {
     node: true,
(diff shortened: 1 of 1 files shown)


edewit (approved):
:tada: works on linux as well

jonkoops:
@ssilvert aren't the the line endings being fixed by ESLint in the pre-commit hook? If that is the case this might allow Windows line endings to slip in.

ssilvert:
> @ssilvert aren't the the line endings being fixed by ESLint in the pre-commit hook? If that is the case this might allow Windows line endings to slip in.

I don't know.  If that happens we will need to take another look.  I do think it will be obvious if we start to see line ending changes on my commits though.

edewit:
Git will take care of line endings for sure

PR #1102, opened 2021-08-31
Comment tests

ssilvert:
## Motivation
Cypress test suite fails too often.

## Brief Description
Commented the two tests that often fail.  Hopefully, this will make the test suit more stable.

## Verification Steps
Run the Cypress tests

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -71,7 +71,7 @@ describe("Realm settings", () => {
     return this;
   };
 
-  const deleteProvider = (providerName: string) => {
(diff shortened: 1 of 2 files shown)


mfrances17 (approved):
LGTM, tests seemed to have passed. Saw that you also filed issues on both of these so this is a go for me.

jenny-s51 (approved):

jonkoops:
FYI instead of commenting out tests we can also use `describe.skip()` or `it.skip()` to skip tests. See the Cypress [documentation](https://docs.cypress.io/guides/references/bundled-tools#Mocha) for more information.

PR #1105, opened 2021-08-31
Adds routing for and fixes add new/save existing identity providers

mfrances17:
## Motivation
Allows users to save identity providers as the alias name, as is intended. Also allows you to edit an existing identity provider and save it.

## Brief Description
Prior to this PR, you were able to successfully create new OIDC or SAML identity providers but they would be saved with the providerId name (e.g. 'saml' or 'oidc', etc). This fix will save the id provider by the specified alias name. This also adds new routing for and the ability to edit an existing id provider and save results, which wasn't working previously.

## Verification Steps
1. Go to Identity Providers and add a new SAML provider.
2. Type in a specific alias name instead of the default 'saml'
3. Turn Use entity descriptor to off.
4. Add a Single Sign-On service URL, such as 'https://localhost.8180/auth/realms/sso'.
5. Click Add and verify that it saved correctly, and you now see the settings tab.
6. Go back to the Identity providers list and verify that it exists as named.
7. Click the SAML provider and verify that the settings open for it, and that you can change the Alias or Display order, Save, and those changes are persisted.

## Checklist:

## Additional Notes
None

### DIFF
--- a/src/identity-providers/IdentityProvidersSection.tsx
+++ b/src/identity-providers/IdentityProvidersSection.tsx
@@ -72,7 +72,7 @@ export const IdentityProvidersSection = () => {
   const DetailLink = (identityProvider: IdentityProviderRepresentation) => (
     <Link
       key={identityProvider.providerId}
-      to={`/${realm}/identity-providers/${identityProvider.providerId}/settings`}
+      to={`/${realm}/identity-providers/${identityProvider.providerId}/${identityProvider.alias}/settings`}
     >
       {identityProvider.alias}
       {!identityProvider.enabled && (
--- a/src/identity-providers/IdentityProvidersSection.tsx
+++ b/src/identity-providers/IdentityProvidersSection.tsx
@@ -34,6 +34,7 @@ import { upperCaseFormatter } from "../util";
 import { useConfirmDialog } from "../components/confirm-dialog/ConfirmDialog";
 import { ProviderIconMapper } from "./ProviderIconMapper";
 import { ManageOderDialog } from "./ManageOrderDialog";
+import { toIdentityProviderTab } from "./routes/IdentityProviderTab";
 
 export const IdentityProvidersSection = () => {
   const { t } = useTranslation("identity-providers");
@@ -72,9 +73,15 @@ export const IdentityProvidersSection = () => {
   const DetailLink = (identityProvider: IdentityProviderRepresentation) => (
     <Link
       key={identityProvider.providerId}
-      to={`/${realm}/identity-providers/${identityProvider.providerId}/${identityProvider.alias}/settings`}
+      to={toIdentityProviderTab({
+        realm,
+        providerId: identityProvider.providerId!,
+        alias: identityProvider.alias!,
+      })}
     >
-      {identityProvider.alias}
+      {identityProvider.displayName
+        ? identityProvider.displayName
+        : identityProvider.alias}
       {!identityProvider.enabled && (
         <Badge
           key={`${identityProvider.providerId}-disabled`}
--- a/src/identity-providers/add/DetailSettings.tsx
+++ b/src/identity-providers/add/DetailSettings.tsx
@@ -20,7 +20,6 @@ import { FormAccess } from "../../components/form-access/FormAccess";
 import { ScrollForm } from "../../components/scroll-form/ScrollForm";
 import { ViewHeader } from "../../components/view-header/ViewHeader";
 import { useFetch, useAdminClient } from "../../context/auth/AdminClient";
-import { toUpperCase } from "../../util";
 import { GeneralSettings } from "./GeneralSettings";
 import { AdvancedSettings } from "./AdvancedSettings";
 import { useConfirmDialog } from "../../components/confirm-dialog/ConfirmDialog";
@@ -61,7 +60,7 @@ const Header = ({ onChange, value, save, toggleDeleteDialog }: HeaderProps) => {
     <>
       <DisableConfirm />
       <ViewHeader
-        titleKey={toUpperCase(alias)}
+        titleKey={alias}
         divider={false}
         dropdownItems={[
           <DropdownItem key="delete" onClick={() => toggleDeleteDialog()}>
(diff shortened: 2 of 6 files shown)


src/identity-providers/routes/IdentityProviderTab.ts
```diff
@@ -7,12 +7,13 @@ export type IdentityProviderTab = "settings";
...
   realm: string;
-  id: string;
+  providerId: string;
+  alias: string;
   tab?: IdentityProviderTab;
 };
 
 export const IdentityProviderTabRoute: RouteDef = {
-  path: "/:realm/identity-providers/:id/:tab?",
+  path: "/:realm/identity-providers/:providerId/:alias/:tab?",
```
jenny-s51, line 16: `providerId` can be removed here and we can just use the `alias`. Otherwise the URL looks like `http://localhost:8080/#/master/identity-providers/keycloak-oidc/keycloak-oidc/settings`

src/identity-providers/IdentityProvidersSection.tsx
```diff
@@ -72,7 +72,7 @@ export const IdentityProvidersSection = () => {
   const DetailLink = (identityProvider: IdentityProviderRepresentation) => (
     <Link
       key={identityProvider.providerId}
-      to={`/${realm}/identity-providers/${identityProvider.providerId}/settings`}
+      to={`/${realm}/identity-providers/${identityProvider.providerId}/${identityProvider.alias}/settings`}
     >
       {identityProvider.alias}
```
jenny-s51, line 77: If `displayName` is set, this should render the `displayName` here like in the old console.
```suggestion
      {identityProvider.displayName
        ? identityProvider.displayName
        : identityProvider.alias}
```

jenny-s51 (changes requested):
Looking good @mfrances17! I noticed this too and implemented the same changes here https://github.com/keycloak/keycloak-admin-ui/compare/master...jenny-s51:samlAddMappers as part of a future PR for adding mappers to saml providers.

Just a couple things to fix, otherwise LGTM. Would be great if we could try to merge this ASAP because there is a good chance it might fix the tests for https://github.com/keycloak/keycloak-admin-ui/pull/1058.

src/identity-providers/routes/IdentityProviderTab.ts
mfrances17: @jenny-s51 actually, you need both... they are only repeated because you named the alias the same as the provider type. but in most cases it would be more like `http://localhost:8080/#/master/identity-providers/saml/my-alias/settings`

src/identity-providers/IdentityProvidersSection.tsx
mfrances17: @jenny-s51 this is true, but i was going to fix that in a future PR. i wanted to get this in ASAP because i was not able to edit existing id providers, so it makes it difficult to test that all the advanced fields are working. so i'd like to handle that outside of this PR if that's ok.

src/identity-providers/routes/IdentityProviderTab.ts
edewit: You don't need both, because you load the identity provider based on the `alias` then you also have the `providerId`, no need to put both in the url

mfrances17: As discussed offline, we need both providerId and alias... providerId is used so we know which settings form to load based on the provider type (SAML, OIDC, etc). So no change required here.

src/identity-providers/IdentityProvidersSection.tsx
mfrances17: Added this, thank you!

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -109,8 +111,8 @@ export const DetailSettings = () => {
     const p = provider || getValues();
     try {
       await adminClient.identityProviders.update(
-        { alias: id },
-        { ...p, alias: id, providerId: id }
+        { alias: alias },
```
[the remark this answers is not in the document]
mfrances17: Yes, much better, thanks. I fixed all of these to use the shorter version.

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -109,8 +111,8 @@ export const DetailSettings = () => {
...
+        { ...p, alias: alias, providerId: providerId }
```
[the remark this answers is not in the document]
mfrances17: Fixed

src/identity-providers/IdentityProvidersSection.tsx
```diff
@@ -72,7 +72,7 @@ export const IdentityProvidersSection = () => {
   const DetailLink = (identityProvider: IdentityProviderRepresentation) => (
     <Link
       key={identityProvider.providerId}
-      to={`/${realm}/identity-providers/${identityProvider.providerId}/settings`}
+      to={`/${realm}/identity-providers/${identityProvider.providerId}/${identityProvider.alias}/settings`}
```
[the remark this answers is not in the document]
mfrances17: Yep, I changed it to use the Jon method, thanks.

edewit (approved):

jenny-s51 (approved):
LGTM @mfrances17! Just FYI, currently if you change the `alias` of the SAML provider, the change is not persisted as described in your verification steps... though in the old console, the `alias` field is set to read-only anyway (wondering if there was a reason behind that). 

Should `alias` be set to read-only here or do you think we should open an issue? @mfrances17 @edewit in the new design it was set to be editable - however it looks like the URL would have to change based on the new saved alias

mfrances17:
> Should `alias` be set to read-only here or do you think we should open an issue? @mfrances17 @edewit in the new design it was set to be editable - however it looks like the URL would have to change based on the new saved alias

@jenny-s51 i noticed that during my testing, and then i tested it in the old console and realized it was readonly. i'll file an issue to cover this. it's a bit more complex - if we are to have it work like the new console, the alias field should only be writeable if you are creating a new provider, and then read only when you want to edit an existing provider. the display name should be the alias name, but only if display name is not null. if display name exists, it has precedence to be used as the display name instead of the alias. and display name should be writeable on new or when editing an existing provider. i'll add all this to the issue.

mfrances17:
Filed issue https://github.com/keycloak/keycloak-admin-ui/issues/1116 for the alias/display issues.

PR #1117, opened 2021-09-02
Time selector was generating duplicate id's

ssilvert:
## Motivation
The Time selector was generating duplicate id's when you go to Realm Settings.  

## Brief Description
I think it should be OK to let the id be auto-generated if needed.  I don't expect that anything relied on this ID since it was using a timestamp.


edewit (approved):
right I've seen it before and thought I fixed it already... totally agree we don't need an id here

PR #1127, opened 2021-09-06
Warn about unnecessary conditions

jonkoops:
Adds an ESLint rule to warn about unnecessary conditions. A warning was chosen as there is too much code to fix for the error level.

## Motivation
This comes up as a review comment every once and a while, enforcing it as a rule will prevent some trivial review comments.


jonkoops:
I've created #1128 to follow up on this problem so the warnings can eventually be removed.

agagancarczyk (approved):

PR #1139, opened 2021-09-08
Process UXD feedback for client revocation

### LINKED ISSUE #958: [UX Review] Add tooltip to the Push button of Revocation
Location: Client -> Advanced -> Revocation

1. The current location of the "Set to now" and "Clear" buttons are confusing. A better solution would be -

image

2. In the old console, a tooltip will display when hovering on the tooltip. This is missing in the current design.

image

The solution is - When the admin URL (in the settings tab) isn’t set, Push is disabled, hovering will show a tooltip (use [this component](https://www.patternfly.org/v4/components/button#aria-disabled-link-as-button-with-tooltip))

image

3. Update the tooltip of this field to "Revoke any tokens issued before this time for this client. To push the policy, you should set an effective admin URL in the Settings tab first"

jonkoops:
Closes #958

### DIFF
--- a/cypress/integration/clients_test.spec.ts
+++ b/cypress/integration/clients_test.spec.ts
@@ -120,18 +120,6 @@ describe("Clients test", function () {
       new AdminClient().deleteClient(client);
     });
 
-    it("Revocation", () => {
-      advancedTab.checkNone();
-
-      advancedTab.clickSetToNow().checkSetToNow();
-      advancedTab.clickClear().checkNone();
-
-      advancedTab.clickPush();
-      masthead.checkNotificationMessage(
-        "No push sent. No admin URI configured or no registered cluster nodes available"
-      );
-    });
-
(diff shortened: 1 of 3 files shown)


yih-wang (changes requested):
@jonkoops Thanks for the updates, that looks great!

Only one minor issue that hasn't been addressed.

- For the third point I mentioned in the original issue, I meant to change the tooltip of the "Not before" label to "Revoke any tokens issued before this time for this client. To push the policy, you should set an effective admin URL in the Settings tab first". That hasn't been addressed in this commit.

jonkoops:
@yih-wang Ahh, I see. I misunderstood the intention there. I've updated the tooltip to reflect this, can you take another look?

yih-wang (approved):
LGTM, thanks!

agagancarczyk (approved):

PR #1118, opened 2021-09-02
Identity Providers(Mappers): Implement "Add Mapper" Screen/Functionality for OIDC and SAML IdPs

jenny-s51:
## Motivation

https://issues.redhat.com/browse/CLUX-230
https://issues.redhat.com/browse/CLUX-231

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] User-visible strings are using the react-i18next framework (useTranslation)
- [x] Help has been implemented
- [x] axe report has been run and resulting a11y issues have been resolved
- [x] Unit tests have been created/updated

PR #1140, opened 2021-09-08
Identity Providers(Mappers): Edit mappers

jenny-s51:
## Motivation

Dependent on #1118 <--- needs merge before this PR is ready for review

## Brief Description

Adds edit functionality to IdP mappers.

## Verification Steps

1. Go to Identity Providers
2. Create/Select an IdP
3. Go to mappers tab
4.  Create/Select an IdP mapper
5. Edit the mapper and verify the data is saved correctly. 

## Checklist:


ssilvert:
@jenny-s51 The failing test looks legit.  I think it will need to be fixed before we merge.

src/identity-providers/add/AddMapper.tsx
```diff
@@ -50,55 +52,137 @@ export const AddMapper = () => {
...
+            // }
+          },
+
+          ({ allMapperTypes, allRoles }) => {
+            setMapperTypes(allMapperTypes);
+            setRoles(allRoles);
+          },
+          []
+        );
+  }
```
jonkoops, line 159: It's [not allowed](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) to render hooks conditionally, so this has to be a single `useFetch` with any conditional logic inside the function body. The calls to the Admin Client in this function are also not dependent on each other, which means they can be ran in parallel using [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all).

I tried to refactor this code a little and this is what I ended up with (untested):
```ts
useFetch(
  () =>
    Promise.all([
      id ? adminClient.identityProviders.findOneMapper({ alias, id }) : null,
      adminClient.identityProviders.findMapperTypes({ alias }),
      !id ? adminClient.roles.find() : null,
    ]),
  ([mapper, mapperTypes, roles]) => {
    if (mapper) {
      setCurrentMapper(mapper);
      setupForm(mapper);
    }

    setMapperTypes(mapperTypes);

    if (roles) {
      setRoles(roles);
    }
  },
  []
);
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -106,6 +109,19 @@ export const DetailSettings = () => {
   const history = useHistory();
   const { realm } = useRealm();
 
+  const MapperLink = ({ name, mapperId }: IdPWithMapperAttributes) => (
+    <Link
+      to={toIdentityProviderEditMapper({
+        realm,
+        alias: alias!,
```
jonkoops, line 116: The `alias` prop is of type `string` and thus cannot be `undefined`, this means the non-null assertion can be removed.
```suggestion
        alias,
```

src/identity-providers/add/DetailSettings.tsx
```diff
@@ -106,6 +109,19 @@ export const DetailSettings = () => {
...
+        providerId: providerId,
```
jonkoops, line 117: If the key and value share the same name the part after the colon can be removed.
```suggestion
        providerId,
```

src/identity-providers/routes/EditMapper.ts
```diff
@@ -0,0 +1,26 @@
...
+import { AddMapper } from "../add/AddMapper";
+
+export type IdentityProviderTab = "mappers";
+
+export type IdentityProviderEditMapperParams = {
+  realm: string;
+  id?: string;
+  tab?: IdentityProviderTab;
+  alias?: string;
+  providerId?: string;
```
jonkoops, line 13: All these fields are required in the route path, so they should be required in this type as well. The `tab` field is never used in the path, so it can be removed. It is also a good idea to keep the order of these fields the same as they appear in the route path.

```suggestion
  providerId: string;
  alias: string;
  id: string;
```

src/realm-roles/AssociatedRolesModal.tsx
```diff
@@ -42,7 +44,7 @@ export const AssociatedRolesModal = (props: AssociatedRolesModalProps) => {
   const [key, setKey] = useState(0);
   const refresh = () => setKey(new Date().getTime());
 
-  const { id } = useParams<{ id: string }>();
+  const { id } = useParams<RealmRoleParams>();
```
jonkoops, line 47: Nice, thanks for cleaning that up!

src/realm-roles/AssociatedRolesModal.tsx
```diff
@@ -143,19 +145,22 @@ export const AssociatedRolesModal = (props: AssociatedRolesModalProps) => {
...
+        (fetchedRole) => {
+          if (fetchedRole) {
+            setName(fetchedRole.name!);
+          } else {
+            setName(t("createRole"));
+          }
+        },
+        []
+      );
+  }
```
jonkoops, line 163: It's [not allowed](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) to render hooks conditionally, so this has to be a single `useFetch` with any conditional logic inside the function body.
```suggestion
  useFetch(
    () =>
      props.isMapperId
        ? adminClient.roles.findOneById({ id })
        : Promise.resolve(null),
    (role) => setName(role ? role.name! : t("createRole")),
    []
  );
```

jonkoops (changes requested):

src/identity-providers/add/AddMapper.tsx
jenny-s51: great idea! thank you

src/identity-providers/routes/IdentityProvider.ts
```diff
@@ -1,20 +1,32 @@
...
+
+export type IdentityProviderCreateParams = {
   realm: string;
-  id: string;
+  providerId?: string;
+};
+
+export type IdentityProviderParams = IdentityProviderCreateParams & {
+  alias: string;
+  tab: IdentityProviderTabs;
```
jonkoops, line 16: Each route should have no dependencies on other routes, and neither should their types. The reason for this is to prevent accidentally introducing new params when editing another route.
```suggestion
export type IdentityProviderParams = {
  realm: string;
  providerId: string;
  alias: string;
  tab: IdentityProviderTabs;
```

src/user/UserIdentityProviderLinks.tsx
```diff
@@ -92,8 +92,9 @@ export const UserIdentityProviderLinks = () => {
   const idpLinkRenderer = (idp: FederatedIdentityRepresentation) => {
     return (
       <Link
-        to={toIdentityProviderTab({
+        to={toIdentityProvider({
           realm,
+          // providerId: idp.identityProvider!,
```
jonkoops, line 97: Some commented out code here, but this field should be passed to create a valid link,

jonkoops (changes requested):

src/identity-providers/routes/IdentityProvider.ts
```diff
@@ -1,20 +1,32 @@
...
+import { AddIdentityProvider } from "../add/AddIdentityProvider";
+import { DetailSettings } from "../add/DetailSettings";
 
-export type IdentityProviderParams = {
+type IdentityProviderTabs = "settings" | "mappers";
+
+export type IdentityProviderCreateParams = {
   realm: string;
-  id: string;
+  providerId?: string;
```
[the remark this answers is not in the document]
jenny-s51: An error is thrown at https://github.com/keycloak/keycloak-admin-ui/pull/1140#discussion_r711979977 unless the `providerId` is made optional here.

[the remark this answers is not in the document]
jonkoops: In that case the `providerId` should be checked or more likely use a non-null assertion as it is coming from a field that is always present in the responses from the API. So for example:

```ts
providerId: providerId!,
```

src/identity-providers/routes/IdentityProviderCreate.ts
```diff
@@ -0,0 +1,22 @@
+import type { LocationDescriptorObject } from "history";
+import { generatePath } from "react-router-dom";
+import type { RouteDef } from "../../route-config";
+import { AddIdentityProvider } from "../add/AddIdentityProvider";
+
+export type IdentityProviderCreateParams = {
+  realm: string;
+  providerId?: string;
```
jonkoops, line 8: The `providerId` field here should be made required like in the path of the route.
```suggestion
  providerId: string;
```

src/user/UserIdentityProviderLinks.tsx
```diff
@@ -94,7 +94,7 @@ export const UserIdentityProviderLinks = () => {
       <Link
         to={toIdentityProvider({
           realm,
-          // providerId: idp.identityProvider!,
+          providerId: idp.identityProvider,
```
jonkoops, line 97: If the `providerId` field is made required in the route you can use the non-null assertion operator to force this value to be interpreted as defined.
```suggestion
          providerId: idp.identityProvider!,
```

src/identity-providers/routes/IdentityProvider.ts
```diff
@@ -1,43 +1,26 @@
...
-};
-
-export type IdentityProviderParams = IdentityProviderCreateParams & {
   alias: string;
   tab: IdentityProviderTabs;
 };
 
 export const IdentityProviderRoute: RouteDef = {
-  path: "/:realm/identity-providers/:providerId/:alias/:tab",
+  path: "/:realm/identity-providers/:providerId?/:alias/:tab",
```
jonkoops, line 16: Once `providerId` is made required it should be made required here as well. 
```suggestion
  path: "/:realm/identity-providers/:providerId/:alias/:tab",
```

src/identity-providers/routes/IdentityProvider.ts
```diff
@@ -1,43 +1,26 @@
 import type { LocationDescriptorObject } from "history";
 import { generatePath } from "react-router-dom";
 import type { RouteDef } from "../../route-config";
-import { AddIdentityProvider } from "../add/AddIdentityProvider";
 import { DetailSettings } from "../add/DetailSettings";
 
 type IdentityProviderTabs = "settings" | "mappers";
```
jonkoops, line 6: Like in other places in the name of a type is generally singular.
```suggestion
type IdentityProviderTab = "settings" | "mappers";
```

src/identity-providers/add/AddMapper.tsx
```diff
@@ -532,16 +531,17 @@ export const AddMapper = () => {
...
-            onClick={() =>
-              history.push(
-                toIdentityProvider({
+            component={(props) => (
+              <Link
+                {...props}
+                to={toIdentityProvider({
                   realm,
-                  providerId: providerId,
+                  providerId,
```
jonkoops, line 539: Once `providerId` is required you can use the non-null assertion operator here.
```suggestion
                  providerId: providerId!,
```

src/identity-providers/add/AddIdentityProvider.tsx
```diff
@@ -40,14 +38,14 @@ export const AddIdentityProvider = () => {
...
-        providerId: providerId,
+        providerId,
         alias: providerId,
       });
       addAlert(t("createSuccess"), AlertVariant.success);
       history.push(
         toIdentityProvider({
           realm,
-          providerId: providerId,
+          providerId,
```
jonkoops, line 48: Once `providerId` is required you can use the non-null assertion operator here.
```suggestion
          providerId: providerId!,
```

src/identity-providers/IdentityProvidersSection.tsx
```diff
@@ -113,11 +110,18 @@ export const IdentityProvidersSection = () => {
...
-            data-testid={provider.id}
-            onClick={() => navigateToCreate(provider.id)}
-          >
-            {provider.name}
-          </DropdownItem>
+            component={
+              <Link
+                to={toIdentityProviderCreate({
+                  realm,
+                  providerId: provider.id,
```
jonkoops, line 117: Once `providerId` is required you can use the non-null assertion operator here.
```suggestion
                  providerId: providerId!,
```

jonkoops (changes requested):

ssilvert:
@jenny-s51 This only works for Hardcoded User Session Attribute.  Are you planning to implement the other mapper types in a subsequent PR?

ssilvert (changes requested):
Help text is incorrect for Name and Mapper type.  Note that in the old console, the help text for Mapper type changes based on which one you select.

jenny-s51:
@ssilvert would you mind providing some more info? I am not able to replicate this -- currently I can edit mappers for all mapper types.

jenny-s51:
> Help text is incorrect for Name and Mapper type. Note that in the old console, the help text for Mapper type changes based on which one you select.

@ssilvert Nice catch - I'm on it

PR #1148, opened 2021-09-13
Stop running unit tests on commit

jonkoops:
## Motivation
Running the unit tests on every commit is a tad annoying, since the CI already takes care of this let's drop it.


edewit (approved):
yeah let's leave that to the CI to execute

PR #1149, opened 2021-09-13
Realm setting profiles

agagancarczyk:
## Motivation
1. Add radio button switch between "From view" and "JSON editor" as per https://marvelapp.com/prototype/h11728e/screen/81529504 and https://marvelapp.com/prototype/h11728e/screen/81529214

Implementation: 
Screenshot 2021-09-13 at 14 23 40

Screenshot 2021-09-13 at 14 23 54

2. Add "Delete profile?" dialog as per https://marvelapp.com/prototype/h11728e/screen/81529212

Implementation: 
Screenshot 2021-09-13 at 14 31 51

## Verification Steps
Navigate to /realm-settings page, select the "Client policies" tab, and verify:
 1. Table display a list of client policies.
 2. Selecting <more> icon allows selecting "Delete" button, which display "Delete profile?" dialogue.
 3. Switching between "From view" and "JSON editor" is possible.
 4. Switching to "JSON editor" display client policies JSON in edit mode, "Save" and "Reload" buttons are placed underneath the editor.

## Checklist:


src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,185 @@
...
+  const [profiles, setProfiles] = useState<ClientProfilesRepresentation>();
+  const [show, setShow] = useState(false);
+
+  const loader = async () => {
+    const clientProfiles = await adminClient.clientPolicies.listProfiles({
+      realm: realmName,
+      includeGlobalProfiles: true,
+    });
+    setProfiles(clientProfiles);
+    return clientProfiles.globalProfiles;
```
jenny-s51, line 38: If I try to make a new custom profile in the old console, it doesn't show up in the data table because this is only returning the `globalProfiles`; we should be returning both `profiles` and `globalProfiles`.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,185 @@
...
+              {name}
+            </Link>{" "}
+            <Label key={`label-${name}`} color="blue">
+              {t("global")}
+            </Label>
+          </>
+        )}
+        {profiles?.profiles?.length! > 0 && (
+          <Link to={""} key={`link-${name}`}>
+            {name}
```
jenny-s51, line 79: If I try to create a new custom profile in the old console, this renders the name in the data list twice:

Screen Shot 2021-09-13 at 4 04 06 PM

jenny-s51 (changes requested):
Nice work @agagancarczyk! Profiles are looking good. 

Left some comments and suggestions below 🙂

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,185 @@
...
+                toggleDeleteDialog();
+              },
+            },
+          ]}
+          columns={[
+            {
+              name: "name",
+              displayKey: t("clientProfileName"),
+              cellFormatters: [cellFormatter],
+              transforms: [cellWidth(10)],
```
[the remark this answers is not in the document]
agagancarczyk: Fixed - thank you @jenny-s51 !

src/realm-settings/messages.ts
```diff
@@ -202,6 +202,24 @@ export default {
     clientPoliciesTab: "Client policies tab",
     clientProfilesSubTab: "Client profiles subtab",
     clientPoliciesSubTab: "Client policies subtab",
+    profilesConfigType: "Configure via:",
+    profilesConfigTypes: {
+      fromView: "From view",
```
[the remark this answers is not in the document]
agagancarczyk: Fixed - thank you @jenny-s51 !

src/realm-settings/RealmSettingsSection.css
```diff
@@ -180,4 +180,8 @@ article.pf-c-card.pf-m-flat.kc-login-settings-template
 .kc-help-text {
   margin: var(--pf-global--spacer--sm);
   font-size: var(--pf-global--FontSize--sm);
+}
+
+.kc-profiles-config-section {
+  align-items: center;
 }
```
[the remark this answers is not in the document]
agagancarczyk: Fixed - thank you @jenny-s51 !

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,185 @@
...
+              name="fromView"
+              onChange={() => formatChange(false)}
+              label={t("profilesConfigTypes.fromView")}
+              id="fromView-radioBtn"
+            />
+          </FlexItem>
+          <FlexItem>
+            <Radio
+              isChecked={show}
+              name="jsonEditor"
```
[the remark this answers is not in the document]
agagancarczyk: Fixed - thank you @jenny-s51 !

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,185 @@
...
+        <Flex className="kc-profiles-config-section">
+          <FlexItem>
+            <Title headingLevel="h1" size="md">
+              {t("profilesConfigType")}
+            </Title>
+          </FlexItem>
+          <FlexItem>
+            <Radio
+              isChecked={!show}
+              name="fromView"
```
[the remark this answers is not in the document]
agagancarczyk: Fixed - thank you @jenny-s51 !

src/realm-settings/ProfilesTab.tsx
agagancarczyk: Fixed - thanks @jenny-s51

src/realm-settings/ProfilesTab.tsx
agagancarczyk: Fixed - thanks @jenny-s51

package-lock.json
```diff
@@ -16416,11 +16416,6 @@
         "safer-buffer": "^2.0.2",
         "tweetnacl": "~0.14.0"
       },
-      "bin": {
```
jonkoops, line 16419: Hmmm, strange this was removed but seems to be fine :man_shrugging:

jenny-s51 (changes requested):
Looking awesome @agagancarczyk ! Thank you for making those updates.

One more thing:

Screen Shot 2021-09-14 at 9 22 58 AM

According to the mock it looks like custom profiles should not have the "Global" label rendered next to them: https://marvelapp.com/prototype/h11728e/screen/81529211

Will you be adding tests to this PR? Once those are done looks good to merge 🙂

agagancarczyk:
> Looking awesome @agagancarczyk ! Thank you for making those updates.
> 
> One more thing:
> 
> Screen Shot 2021-09-14 at 9 22 58 AM
> 
> According to the mock it looks like custom profiles should not have the "Global" label rendered next to them: https://marvelapp.com/prototype/h11728e/screen/81529211
> 
> Will you be adding tests to this PR? Once those are done looks good to merge 🙂

Thank you @jenny-s51 I'll fix this issue and add some tests :)

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+    const allClientProfiles = clientProfiles.globalProfiles?.concat(
+      clientProfiles.profiles!
+    );
+
+    return allClientProfiles ?? [];
+  };
+
+  function formatChange(show: boolean) {
+    setShow(show);
+  }
```
jonkoops, line 47: Since this function is basically just a wrapped call to `setShow()` with the same arguments perhaps we can use `setShow()` directly here.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+
+  function createNewProfile() {
+    // create a new profile
+  }
+
+  const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
+    titleKey: t("deleteClientProfileConfirmTitle"),
+    messageKey: t("deleteClientProfileConfirm"),
+    continueButtonLabel: "delete",
```
jonkoops, line 56: This string will have to be translated.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+    continueButtonVariant: ButtonVariant.danger,
+    onConfirm: async () => {
+      try {
+        // delete client profile here
+        addAlert(t("deleteClientSuccess"), AlertVariant.success);
+      } catch (error) {
+        addError(t("deleteClientError"), error);
+      }
+    },
+  });
```
jonkoops, line 66: Since the delete functionality is not implemented on the API level I think it's better to leave out this code, otherwise we might ship a product with a button that does not work.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+
+  function cellFormatter(name?: IFormatterValueType) {
+    return (
+      <>
+        <Link to={""} key={`link-${name}`}>
```
jonkoops, line 71: This link is missing a route, but I am assuming this will come later.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+          {name}
+        </Link>{" "}
+        {profiles?.globalProfiles?.length! > 0 && (
```
jonkoops, line 74: Using the non-null assertion operator here can cause issues at run-time. Make sure to check the value first.
```suggestion
        {profiles?.globalProfiles && profiles.globalProfiles.length > 0 && (
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+          searchPlaceholderKey="realm-settings:clientProfileSearch"
+          isPaginated
+          loader={loader}
+          toolbarItem={
+            <ToolbarItem>
+              <Button
+                id="createProfile"
+                onClick={() => createNewProfile()}
+                data-testid="createProfile"
+              >
```
jonkoops, line 128: This button should probably use a link later on instead of using `onClick` to navigate.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+            />
+          }
+        />
+      ) : (
+        <>
+          <div className="pf-u-mt-md pf-u-ml-lg">
+            <CodeEditor
+              isLineNumbersVisible
+              isLanguageLabelVisible
+              code={JSON.stringify(profiles, null, 2)}
```
jonkoops, line 166: Calling `JSON.stringify()` on every render can be quite expensive. It is better to memoize this value so it is not recalculated every time.

```ts
const code = useMemo(() => JSON.stringify(profiles, null, 2), [profiles]);
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+              language={Language.json}
+              height={"30rem"}
```
jonkoops, line 168: Strings can be passed directly,
```suggestion
              height="30rem"
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+            />
+          </div>
+          <div className="pf-u-mt-md">
+            <Button
+              variant={ButtonVariant.primary}
+              className="pf-u-mr-md pf-u-ml-lg"
+            >
+              Save
```
jonkoops, line 176: This string will have to be translated.

src/realm-settings/ProfilesTab.tsx
```diff
@@ -1,12 +1,184 @@
...
+            </Button>
+            <Button variant={ButtonVariant.secondary}>Reload</Button>
```
jonkoops, line 178: This string will have to be translated.

src/realm-settings/RealmSettingsSection.css
```diff
@@ -180,4 +180,16 @@ article.pf-c-card.pf-m-flat.kc-login-settings-template
...
   margin: var(--pf-global--spacer--sm);
   font-size: var(--pf-global--FontSize--sm);
+}
+
+.kc-profiles-config-section {
+  align-items: center;
+}
+
+.kc-form-radio-btn > input {
+  transform: scale(1.6);
```
jonkoops, line 190: Do we want to use a `scale` for this? It's not possible to set the height and width?

jonkoops (changes requested):

src/realm-settings/ProfilesTab.tsx
agagancarczyk: Actually, when I investigated this again yesterday, I have noticed that in the old console they are using PUT request for deleting non-global profiles. I can try to implement it in the same way. wdyt?

src/realm-settings/ProfilesTab.tsx
agagancarczyk: Yes, this will come later :)

src/realm-settings/ProfilesTab.tsx
agagancarczyk: yes, you are correct. I'll be working on this next.

agagancarczyk:
> > Looking awesome @agagancarczyk ! Thank you for making those updates.
> > One more thing:
> > Screen Shot 2021-09-14 at 9 22 58 AM
> > According to the mock it looks like custom profiles should not have the "Global" label rendered next to them: https://marvelapp.com/prototype/h11728e/screen/81529211
> > Will you be adding tests to this PR? Once those are done looks good to merge 🙂
> 
> Thank you @jenny-s51 I'll fix this issue and add some tests :)

Fixed

src/realm-settings/RealmSettingsSection.css
agagancarczyk: I think it's ok, let's leave it.

src/realm-settings/ProfilesTab.tsx
jonkoops: Oh wow, that sounds like a hack on the API side hahah. Yeah, if you can get that to work, why not :man_shrugging:

src/realm-settings/RealmSettingsSection.css
jonkoops: Aight, I am no CSS magician like y'all so let's roll with it :smile:

src/realm-settings/ProfilesTab.tsx
```diff
@@ -15,36 +15,51 @@ import { useTranslation } from "react-i18next";
...
 import { Link } from "react-router-dom";
-import type ClientProfilesRepresentation from "@keycloak/keycloak-admin-client/lib/defs/clientProfilesRepresentation";
 import "./RealmSettingsSection.css";
+import type ClientPolicyExecutorRepresentation from "@keycloak/keycloak-admin-client/lib/defs/clientPolicyExecutorRepresentation";
+
+type ClientProfile = {
+  description?: string;
+  executors?: ClientPolicyExecutorRepresentation[];
+  name?: string;
+  global?: boolean;
```
jonkoops, line 27: Looking at the code in the loader I think you can make this field required as it is always added to the object before `setProfiles()` is called.
```suggestion
  global: boolean;
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -15,36 +15,51 @@ import { useTranslation } from "react-i18next";
...
+        global: true,
+      })
     );
 
+    const profiles = allProfiles.profiles?.map((profiles) => ({
+      ...profiles,
+      global: false,
+    }));
+
+    const allClientProfiles = globalProfiles?.concat(profiles!);
```
jonkoops, line 56: Since `profiles` theoretically could be `undefined` it might be better to fall back to default here rather than using the non-null assertion operator.
```suggestion
    const allClientProfiles = globalProfiles?.concat(profiles ?? []);
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -65,20 +80,18 @@ export const ProfilesTab = () => {
...
-        {profiles?.globalProfiles?.length! > 0 && (
-          <Label key={`label-${name}`} color="blue">
-            {t("global")}
-          </Label>
-        )}
-      </>
-    );
-  }
+  const cellFormatter = (row: ClientProfile) => (
+    <Link to={""} key={`label-${row.name}`}>
```
jonkoops, line 84: When providing a `key` prop it only has to be unique within the context it is used. I think you can pass `row.name` directly to achieve the same result.
```suggestion
    <Link to={""} key={row.name}>
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -65,20 +80,18 @@ export const ProfilesTab = () => {
...
+      {row.name} {""}
+      {row.global ? (
+        <Label key={`label-${row.global}`} color="blue">
+          {t("global")}
+        </Label>
+      ) : (
+        ""
+      )}
```
jonkoops, line 92: Since the `Label` component is shown conditionally and not in a `.map()` I believe there is no need to provide a `key` prop here. Also since the else condition is rendering an empty string perhaps it can be omitted?

```suggestion
      {row.global && <Label color="blue">{t("global")}</Label>}
```

jonkoops (changes requested):

src/realm-settings/ProfilesTab.tsx
agagancarczyk: You're right! thanks @jonkoops

src/realm-settings/ProfilesTab.tsx
agagancarczyk: like this idea - nice catch! ty

agagancarczyk:
Great work @jonkoops !

agagancarczyk:
> > Looking awesome @agagancarczyk ! Thank you for making those updates.
> > One more thing:
> > Screen Shot 2021-09-14 at 9 22 58 AM
> > According to the mock it looks like custom profiles should not have the "Global" label rendered next to them: https://marvelapp.com/prototype/h11728e/screen/81529211
> > Will you be adding tests to this PR? Once those are done looks good to merge 🙂
> 
> Thank you @jenny-s51 I'll fix this issue and add some tests :)

@jenny-s51 I'll need to merge this without tests because all tests are still failing. I'll add tests for all client profiles in my next PR - hope that's ok.

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "keycloak-admin-ui",
```
jonkoops, line 8: This change is generated by older versions of NPM. Make sure you are running the latest version of NPM (`npm install -g npm@latest`). After running `npm install` this should be removed automatically.

src/realm-settings/routes/NewClientProfile.ts
```diff
@@ -0,0 +1,9 @@
+import type { RouteDef } from "../../route-config";
+import { NewClientProfileForm } from "../NewClientProfileForm";
+
+export const NewClientProfileRoute: RouteDef = {
```
jonkoops, line 4: Can you also create a `toNewClientProfile()` function to accompany this route?

src/realm-settings/ProfilesTab.tsx
```diff
@@ -136,13 +127,18 @@ export const ProfilesTab = () => {
             <ToolbarItem>
               <Button
                 id="createProfile"
-                onClick={() => createNewProfile()}
+                onClick={() =>
+                  history.push(
+                    `/${realmName}/realm-settings/clientPolicies/new-client-profile`
+                  )
+                }
```
jonkoops, line 134: Make sure to use a link here and use a function for the route.

```tsx
<Button component={Link} to={toNewClientProfile({ realm })}></Button>
```

src/realm-settings/ProfilesTab.tsx
```diff
@@ -81,15 +78,9 @@ export const ProfilesTab = () => {
   });
 
   const cellFormatter = (row: ClientProfile) => (
-    <Link to={""} key={`label-${row.name}`}>
+    <Link to={""} key={row.name}>
```
jonkoops, line 81: We don't have a link for this page yet?

src/realm-settings/NewClientProfileForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useRealm } from "../context/realm-context/RealmContext";
+
+export const NewClientProfileForm = () => {
+  const { t } = useTranslation("realm-settings");
+  const history = useHistory();
+  const { register, errors } = useForm();
+  const { realm } = useRealm();
+
+  function save() {
+    //TODO
```
jonkoops, line 25: This still needs to be implemented? Do you want to do that in this PR or the next?

src/realm-settings/NewClientProfileForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+              data-testid="realm-settings-client-profile-save-button"
+            >
+              {t("common:save")}
+            </Button>
+            <Button
+              onClick={() =>
+                history.push(`/${realm}/realm-settings/clientPolicies`)
+              }
+              variant="link"
+            >
```
jonkoops, line 77: Make sure to use a Link button here that actually renders an `a` element that is styled like a button.

jonkoops (changes requested):

src/realm-settings/routes/NewClientProfile.ts
agagancarczyk: Sure @jonkoops!

src/realm-settings/ProfilesTab.tsx
agagancarczyk: I have added this, but typescript is complaining about having "to=", would you know why this is working fine for other buttons but not mine?

src/realm-settings/ProfilesTab.tsx
agagancarczyk: No, but I'll need to implement this next. It seems to be included in the designs.

src/realm-settings/NewClientProfileForm.tsx
agagancarczyk: I'll work on this next. I'd prefer to include it in the next PR, if that's ok.

src/realm-settings/ProfilesTab.tsx
jonkoops: Yeah, this is a known issue in PatternFly. It can be resolved by adding a `// @ts-ignore`. I've pushed a commit to make the build pass again.

src/realm-settings/ProfilesTab.tsx
jonkoops: Then consider this issue resolved.

src/realm-settings/NewClientProfileForm.tsx
jonkoops: :ok: :+1:

src/realm-settings/NewClientProfileForm.tsx
```diff
@@ -71,9 +71,10 @@ export const NewClientProfileForm = () => {
             </Button>
             <Button
               id="cancelCreateProfile"
-              component={Link}
               // @ts-ignore
-              to={toClientPolicies({ realm })}
+              onClick={() =>
```
jonkoops, line 75: This can still be a link component:

```
component={Link}
to={`/${realm}/realm-settings/clientPolicies`}
```

jonkoops (changes requested):

jonkoops (approved):
LGTM :heavy_check_mark:

PR #1150, opened 2021-09-14
fixes some issue on the events tab

edewit:
* user details can be empty
* fix the number of rows displayed with collapse

### DIFF
--- a/src/components/table-toolbar/KeycloakDataTable.tsx
+++ b/src/components/table-toolbar/KeycloakDataTable.tsx
@@ -36,7 +36,7 @@ type BaseRow<T> = {
 
 type Row<T> = BaseRow<T> & {
   selected: boolean;
-  isOpen: boolean;
+  isOpen?: boolean;
   disableSelection: boolean;
   disableActions: boolean;
 };
@@ -203,6 +203,15 @@ export function KeycloakDataTable<T>({
   const [key, setKey] = useState(0);
   const refresh = () => setKey(new Date().getTime());
 
+  const renderCell = (columns: (Field<T> | DetailField<T>)[], value: T) => {
+    return columns.map((col) => {
+      if (col.cellRenderer) {
+        return { title: col.cellRenderer(value) };
+      }
+      return _.get(value, col.name);
+    });
+  };
+
   const convertToColumns = (data: T[]): (Row<T> | SubRow<T>)[] => {
     return data
       .map((value, index) => {
@@ -215,24 +224,14 @@ export function KeycloakDataTable<T>({
             selected: !!selected.find(
               (v) => _.get(v, "id") === _.get(value, "id")
             ),
-            isOpen: false,
-            cells: columns.map((col) => {
-              if (col.cellRenderer) {
-                return { title: col.cellRenderer(value) };
-              }
-              return _.get(value, col.name);
-            }),
+            isOpen: detailColumns?.[0]?.enabled?.(value) ? false : undefined,
+            cells: renderCell(columns, value),
           },
         ];
         if (detailColumns?.[0]?.enabled?.(value)) {
           row.push({
             parent: index * 2,
-            cells: detailColumns!.map((col) => {
-              if (col.cellRenderer) {
-                return { title: col.cellRenderer(value) };
-              }
-              return _.get(value, col.name);
-            }),
+            cells: renderCell(detailColumns, value),
           } as SubRow<T>);
         }
         return row;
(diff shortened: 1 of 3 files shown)


src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -215,24 +224,14 @@ export function KeycloakDataTable<T>({
...
               (v) => _.get(v, "id") === _.get(value, "id")
             ),
-            isOpen: false,
-            cells: columns.map((col) => {
-              if (col.cellRenderer) {
-                return { title: col.cellRenderer(value) };
-              }
-              return _.get(value, col.name);
-            }),
+            isOpen: detailColumns?.[0]?.enabled?.(value) ? false : undefined,
```
jonkoops, line 227: If I get this correctly `isOpen` is `false` if `enabled()` is `true`? Otherwise it's also a falsy value, is that correct behavior? I was expecting something like:

```suggestion
            isOpen: detailColumns?.[0]?.enabled?.(value) ?? false,
```

jonkoops: Since this logic is used twice in this function perhaps it can be extracted into a separate variable?

```ts
const isColumnEnabled = detailColumns?.[0]?.enabled?.(value) ?? false;
```

jonkoops (changes requested):

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: No, it should only be `false` when "detailColumns" are not undefined and enabled for that value, but your right I can make it a function

edewit:
fixes: #1156

ssilvert:
> * fix the number of rows displayed with collapse

What do you mean by this?

ssilvert:
BTW, I verified that this does fix user events where there is no user, like failed login.  That works nicely.

I'm just not sure what I should test for "fix the number of rows displayed with collapse"

edewit:
@ssilvert it does fix #1156 the problem was the collapsible rows where counted as rows so the number of rows show was always half of the actual just like you describe in issue 1156

edewit:
@ssilvert right forgot about the pagination numbers, fixed those now as well

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -215,24 +226,14 @@ export function KeycloakDataTable<T>({
...
               (v) => _.get(v, "id") === _.get(value, "id")
             ),
-            isOpen: false,
-            cells: columns.map((col) => {
-              if (col.cellRenderer) {
-                return { title: col.cellRenderer(value) };
-              }
-              return _.get(value, col.name);
-            }),
+            isOpen: isDetailColumnsEnabled(value) ? false : undefined,
```
jonkoops, line 229: Is the result of this expression not always `false` or `undefined`? Perhaps you can default to `false` here?
```suggestion
            isOpen: false,
```

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -203,7 +203,18 @@ export function KeycloakDataTable<T>({
...
   const refresh = () => setKey(new Date().getTime());
 
+  const renderCell = (columns: (Field<T> | DetailField<T>)[], value: T) => {
+    return columns.map((col) => {
+      if (col.cellRenderer) {
+        return { title: col.cellRenderer(value) };
+      }
+      return _.get(value, col.name);
+    });
+  };
```
jonkoops, line 213: Since arrow functions return the value from the last statement this code could be shortened a bit:

```suggestion
  const renderCell = (columns: (Field<T> | DetailField<T>)[], value: T) =>
    columns.map((col) => {
      if (col.cellRenderer) {
        return { title: col.cellRenderer(value) };
      }
      return _.get(value, col.name);
    });
```

You could make it even more compact if you use a ternary: 

```suggestion
  const renderCell = (columns: (Field<T> | DetailField<T>)[], value: T) =>
    columns.map((col) =>
      col.cellRenderer
        ? { title: col.cellRenderer(value) }
        : _.get(value, col.name)
    );
```

jonkoops (changes requested):

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: false is not the same as undefined when we always set it to false it will always show the collapse icon we only want that when detail columns are enabled

ssilvert:
> @ssilvert right forgot about the pagination numbers, fixed those now as well

Yes, verified that this fixes #1156

src/components/table-toolbar/KeycloakDataTable.tsx
jonkoops: I see, thanks for clarifying.

ssilvert:
Group test should be fixed now if you rebase.

jonkoops (approved):

PR #1161, opened 2021-09-15
changed to use code editor

edewit:
use the `<CodeEditor` on the `<JsonFileUpload` component

### DIFF
--- a/src/components/json-file-upload/JsonFileUpload.tsx
+++ b/src/components/json-file-upload/JsonFileUpload.tsx
@@ -152,7 +152,7 @@ export const JsonFileUpload = ({
           >
             <CodeEditor
               isLineNumbersVisible
-              code={fileUpload.value as string}
+              code={fileUpload.value}
               language={Language.json}
               height="128px"
               onChange={(value, event) =>
--- a/src/components/json-file-upload/JsonFileUpload.tsx
+++ b/src/components/json-file-upload/JsonFileUpload.tsx
@@ -8,9 +8,10 @@ import {
   FileUploadProps,
 } from "@patternfly/react-core";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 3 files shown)


src/components/json-file-upload/JsonFileUpload.tsx
```diff
@@ -60,7 +61,7 @@ export const JsonFileUpload = ({
     } else {
       setFileUpload({
         ...fileUpload,
-        value,
+        value: value as string,
```
jonkoops, line 64: Are you sure this value will always be a string?

src/components/json-file-upload/JsonFileUpload.tsx
```diff
@@ -147,7 +148,18 @@ export const JsonFileUpload = ({
             dropzoneProps={{
               accept: ".json",
             }}
-          />
+            hideDefaultPreview
+          >
+            <CodeEditor
+              isLineNumbersVisible
+              code={fileUpload.value as string}
```
jonkoops, line 155: This is already always a string according to the type, so this type cast can be removed.
```suggestion
              code={fileUpload.value}
```

src/components/json-file-upload/JsonFileUpload.tsx
```diff
@@ -147,7 +148,18 @@ export const JsonFileUpload = ({
...
+              language={Language.json}
+              height="128px"
+              onChange={(value, event) =>
```
jonkoops, line 158: Would be nice if we get an actual type for this event, perhaps we can open a bug upstream?

jonkoops (changes requested):

src/components/json-file-upload/JsonFileUpload.tsx
ssilvert: +1

jonkoops (approved):

PR #1164, opened 2021-09-15
update to PF 4.135

jenny-s51:
## Motivation
Update Patternfly to the latest releases so we stay current with the latest changes.

## Brief Description
As part of the Patternfly release process, we perform sanity tests of the new proposed versions on the keycloak admin UI and file issues as needed. This occurs about every 3 weeks. As part of that process, we should update keycloak to the latest versions on a regular cadence.

## Verification Steps
Use the keycloak admin UI and verify that no visual or behavioral issues have been introduced, and that there are no new console warnings/errors.

## Checklist:

## Additional Notes
In addition to manual sanity testing, the full cypress test suite was run and no new failures occurred as a result of the PF upgrade.


jonkoops (approved):
Changes look good, clicking around nothing looks broken.

jonkoops:
@jenny-s51 Looks like we have some tests that are failing pretty consistently, can you take a look?

jenny-s51:
@jonkoops yes, I did some investigation yesterday and it looks like an issue was introduced from the PF upgrade (the "Create Realm" button in the realm selector is no longer clickable). I made this known in the #pf-release slack channel and the devs are aware. Waiting on feedback from the PF team regarding the best way to proceed here

mfrances17 (commented):
LGTM, these latest versions fix the issue we were seeing with the realm selector.

Just FYI... I usually don't push to keycloak master until PF officially releases, because the version numbers can and often do potentially change because someone finds something in late validation testing. Once they send an all-clear in the pf-release channel and unfreeze the repo, then I push.

mfrances17 (approved):

PR #1168, opened 2021-09-16
Configure Cypress to use the correct types

jonkoops:
## Motivation
Fixes the type errors that were getting displayed when using an IDE to look at the Cypress test files. This is accomplished by including the Cypress types only in the directory of the tests, and limiting the types of Jest to the `src` directory to avoid the two from conflicting.

This PR also replaces the custom command we were using to get elements by their 'test id' with commands from `@testing-library/cypress`. The reason for this is that I could not get the type working for our custom command. It also makes sense to use Testing Library as we are already using it for the tests in Jest, which should already be familiar.

There is a caveat to be aware of, since we are no longer including all the files in the root ts-config the test files for Jest and Cypress are no longer checked at compile time. This is actually fine as we should not compile any other files when running `npm run build` other than the files in the `src` directory that end up in the final build.

The type-checks for Jest and Cypress will be re-enabled in a subsequent PR that will add these checks in their own respective tools (rather than through Snowpack).


edewit (approved):
using a library always better then rolling your own ;)

agagancarczyk (approved):

jonkoops:
Looks like some of the tests are failing, I'll take a look at that before merging this.

cypress/support/pages/admin_console/manage/realm_settings/RealmSettingsPage.ts
```diff
@@ -185,73 +185,73 @@ export default class RealmSettingsPage {
...
 
     return this;
   }
 
   deleteProvider(providerName: string) {
-    cy.getId("provider-name-link")
+    cy.findByTestId("provider-name-link")
       .contains(providerName)
       .parent()
       .siblings(".pf-c-data-list__item-action")
```
mfrances17, line 257: missed getIds on line 259 & 261

cypress/support/pages/admin_console/manage/realm_roles/AssociatedRolesPage.ts
```diff
@@ -47,7 +50,7 @@ export default class AssociatedRolesPage {
         cy.get("input").click();
       });
 
-    cy.getId(this.addAssociatedRolesModalButton).contains("Add").click();
+    cy.findByTestId(this.addAssociatedRolesModalButton).contains("Add").click();
 
     cy.contains("Users in role").click().getId(this.usersPage).should("exist");
```
mfrances17, line 55: missed a getId here

cypress/support/pages/ViewHeaderPage.ts
```diff
@@ -2,7 +2,7 @@ export default class ListingPage {
   private actionMenu = "action-dropdown";
 
   clickAction(action: string) {
-    cy.getId(this.actionMenu).click().getId(action).click();
+    cy.findByTestId(this.actionMenu).click().getId(action).click();
```
mfrances17, line 5: missed a nested getId here

cypress/integration/realm_settings_test.spec.ts
```diff
@@ -74,14 +80,14 @@ describe("Realm settings", () => {
   /*const deleteProvider = (providerName: string) => {
     const url = `/auth/admin/realms/${realmName}/users/*`;
     cy.intercept(url).as("reload");
-    cy.getId("provider-name")
+    cy.findByTestId("provider-name")
       .contains(providerName)
       .parentsUntil(".pf-c-data-list__item-row")
       .find(".pf-c-dropdown__toggle")
       .click()
       .getId(realmSettingsPage.deleteAction)
```
mfrances17, line 88: it's commented out, but still... since you changed the others to findByTestId, might as well change the one on ln 88 so there's no confusion when it's uncommented out in the future

mfrances17 (changes requested):
Found a few `getId` calls that need to be updated to `findByTestId`, found them by seeing which tests were failing in the cypress test suite. You might want to just do a global search in the `/cypress` dir for `getId` in case there are others. Other than those lines that I called out, looks great.

mfrances17:
Also, i looked at the `[data-testid="provider-name-link"]` cypress test failure in `realm_settings_test.spec.ts`, that test is always going to fail if there are multiple providers defined because they are all using the same test id for the link. I would prefix the data-testid in KeyProvidersTab (ln 378) with the component name or something similar to differentiate them.

jonkoops:
Thanks @mfrances17, I totally missed those calls in my find and replace. I've fixed them up so we should be good to go.

> Also, i looked at the `[data-testid="provider-name-link"]` cypress test failure in `realm_settings_test.spec.ts`, that test is always going to fail if there are multiple providers defined because they are all using the same test id for the link. I would prefix the Link's data-testid in KeyProvidersTab (ln 378) with the component name or something similar to differentiate them.

Yeah, this is a weird one. It seems to fail, but not consistently. Either way this is a change that was already here so let's see if we can get the tests to pass.

PR #1170, opened 2021-09-16
SAML ID provider cleanup

mfrances17:
## Motivation
Several fields in the original SAML identity provider were not working in the first pass, to be fixed later. 

## Brief Description
This fixes it so all SAML fields should be working as expected. Any fields that are not working from here on should be filed as issues.

## Verification Steps
Verify that the following fields now work as expected in the SAML ID provider config, as they worked in the old console (set the value in the old console and verify that the new console reads the value, etc):
Single logout service URL
NameID policy format
Principal type
Principal attribute
Want AuthnRequests signed
Signature algorithm
SAML signature key name
Validating X509 certificates
Pass subject
Allowed clock skew
Comparison
AuthnContext ClassRefs
AuthnContext DeclRefs
Stored tokens readable
Hide on login page

## Checklist:

## Additional Notes
Cypress tests coming in separate PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "keycloak-admin-ui",
       "version": "0.0.1",
       "license": "Apache",
       "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "keycloak-admin-ui",
       "version": "0.0.1",
       "license": "Apache",
       "dependencies": {
--- a/src/identity-providers/add/AdvancedSettings.tsx
+++ b/src/identity-providers/add/AdvancedSettings.tsx
@@ -30,7 +30,7 @@ const LoginFlow = ({
   useFetch(
     () => adminClient.authenticationManagement.getFlows(),
     (flows) =>
-      setFlows(flows?.filter((flow) => flow.providerId === "basic-flow")),
+      setFlows(flows.filter((flow) => flow.providerId === "basic-flow")),
     []
   );
 
--- a/src/identity-providers/add/AdvancedSettings.tsx
+++ b/src/identity-providers/add/AdvancedSettings.tsx
@@ -108,8 +108,9 @@ export const AdvancedSettings = ({ isOIDC, isSAML }: AdvancedSettingsProps) => {
       <SwitchField field="storeToken" label="storeTokens" fieldType="boolean" />
       {isSAML && (
         <SwitchField
-          field="config.addReadTokenRoleOnCreate"
+          field="addReadTokenRoleOnCreate"
           label="storedTokensReadable"
+          fieldType="boolean"
         />
       )}
       {!isOIDC && !isSAML && (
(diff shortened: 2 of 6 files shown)


package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "keycloak-admin-ui",
```
jonkoops, line 8: This change is generated by older versions of NPM, but makes no functional difference. Make sure to update to the latest version of NPM and this change should disappear. You can upgrade NPM by running `npm install -g npm@latest`. 

Alternatively using the latest version of Node should also resolve this, if you have NVM tyou can run `nvm install 16 && nvm alias default 16`.

src/identity-providers/add/AdvancedSettings.tsx
```diff
@@ -30,7 +30,7 @@ const LoginFlow = ({
   useFetch(
     () => adminClient.authenticationManagement.getFlows(),
     (flows) =>
-      setFlows(flows?.filter((flow) => flow.providerId === "basic-flow")),
+      setFlows(flows.filter((flow) => flow.providerId === "basic-flow")),
```
jonkoops, line 33: Nice, thanks for cleaning that up!

src/identity-providers/add/DescriptorSettings.tsx
```diff
@@ -45,6 +45,12 @@ const Fields = ({ readOnly }: DescriptorSettingsProps) => {
     name: "config.validateSignature",
   });
 
+  const principalType = useWatch({
+    control: control,
```
jonkoops, line 49: If the value has the same name as the key the part after the colon can be omitted.
```suggestion
    control,
```

src/identity-providers/add/ReqAuthnConstraintsSettings.tsx
```diff
@@ -31,23 +31,23 @@ export const ReqAuthnConstraints = () => {
...
+          name="config.authnContextComparisonType"
           defaultValue={comparisonValues[0]}
           control={control}
           render={({ onChange, value }) => (
             <Select
               toggleId="comparison"
               required
               direction="up"
-              onToggle={() => setSyncModeOpen(!syncModeOpen)}
+              onToggle={() => setComparisonOpen(!comparisonOpen)}
```
jonkoops, line 42: The `onToggle` prop provides a value in the callback to set the open state so we don't have to control this ourselves.
```suggestion
              onToggle={(isExpanded) => setComparisonOpen(isExpanded)}
```

jonkoops (changes requested):

package-lock.json
mfrances17: hmmm, i'm already using v16.1.0 of node i'm surprised that's the issue, but i will try updating, thanks

mfrances17: i don't know why i bother questioning it, you were right - fixed when i updated to the latest npm, thanks!

src/identity-providers/add/DescriptorSettings.tsx
mfrances17: fixed, always forget this one - thanks

src/identity-providers/add/ReqAuthnConstraintsSettings.tsx
mfrances17: always forget this one as well - fixed, ty

src/identity-providers/add/DescriptorSettings.tsx
jonkoops: Yeah, might be worth adding a linting rule for it.

jonkoops (approved):

PR #1178, opened 2021-09-16
Check types using ts-jest and upgrade Jest

jonkoops:
## Motivation
Allows unit tests to be properly type checked, which is a caveat that was introduced by #1168. Also allows us to upgrade to Jest 27 as we are no longer tied to the configuration provided by Snowpack.


agagancarczyk (approved):
Great work @jonkoops !

PR #1183, opened 2021-09-16
Realm settings(login tab): fix switches

### LINKED ISSUE #1158: Realm settings --> Login tab is broken
If I click on any of the toggle buttons on the page I will get a success message.  However, the toggle buttons do not change state and the value does not change in the database.

jenny-s51:
## Motivation

Closes #1158

## Verification Steps

1. Go to Realm Settings -> Login Tab
2. Verify that the all switches work as expected.

## Checklist:

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -126,6 +126,15 @@ describe("Realm settings tests", () => {
       realmSettingsPage.toggleSwitch(realmSettingsPage.rememberMeSwitch);
     });
 
+    it("Check login tab values", () => {
+      sidebarPage.goToRealmSettings();
+      cy.findByTestId("rs-login-tab").click();
+
+      cy.get("#kc-user-reg-switch-off").should("be.visible");
+      cy.get("#kc-forgot-pw-switch-off").should("be.visible");
+      cy.get("#kc-remember-me-switch-off").should("not.be.visible");
+    });
+
     it("Go to email tab", () => {
       sidebarPage.goToRealmSettings();
       cy.findByTestId("rs-email-tab").click();
--- a/src/realm-settings/LoginTab.tsx
+++ b/src/realm-settings/LoginTab.tsx
@@ -5,17 +5,32 @@ import { FormAccess } from "../components/form-access/FormAccess";
 import { HelpItem } from "../components/help-enabler/HelpItem";
 import { FormPanel } from "../components/scroll-form/FormPanel";
 import type RealmRepresentation from "@keycloak/keycloak-admin-client/lib/defs/realmRepresentation";
+import { Controller, useForm } from "react-hook-form";
 
 type RealmSettingsLoginTabProps = {
   save: (realm: RealmRepresentation) => void;
   realm: RealmRepresentation;
+  refresh: () => void;
 };
 
 export const RealmSettingsLoginTab = ({
   save,
   realm,
+  refresh,
 }: RealmSettingsLoginTabProps) => {
   const { t } = useTranslation("realm-settings");
+
+  const form = useForm<RealmRepresentation>({ mode: "onChange" });
+
+  const updateSwitchValue = (
+    onChange: (newValue: boolean) => void,
+    value: boolean,
+    name: string
+  ) => {
+    save({ ...realm, [name as keyof typeof realm]: value });
+    onChange(value);
+    refresh();
+  };
 
   return (
     <PageSection variant="light">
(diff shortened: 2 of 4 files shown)


mfrances17 (changes requested):
It's really odd that the changes are automatically saved when each switch changes - the old console had a save/cancel. Personally I would lob this back at design as to why it was changed  to automatically update on every switch status change. 

Aside from that, in the old console, if you turned on 'User registration', you would get another option 'Email as username' - wondering if the intent was to omit this. There is no such option in the new UI. Likewise with the 'Require SSL' dropdown - was this omitted purposely?

cypress/integration/realm_settings_test.spec.ts
```diff
@@ -19,7 +19,7 @@ describe("Realm settings", () => {
   beforeEach(() => {
     keycloakBefore();
     loginPage.logIn();
-    sidebarPage.goToRealm(realmName);
+    // sidebarPage.goToRealm(realmName);
```
jonkoops, line 22: Some commented out code, I believe this might be causing some test failures.

jonkoops (changes requested):
Looks like the tests are failing.

jenny-s51:
> It's really odd that the changes are automatically saved when each switch changes status - the old console had a save/cancel. Personally I would lob this back at design as to why it was changed to automatically update on every switch status change.

@mfrances17 certainly valid points! We had a team discussion about this in Zulip a few months ago -- [here is the link to the thread](https://keycloak.zulipchat.com/#narrow/stream/251649-dev-new-admin/topic/Realm.20settings.20login.20screen)! 

Based on Haley's response it was decided to proceed with the implementation without the save/cancel buttons. 

> Aside from that, in the old console, the enable/disable of the 'Email as username' switch was controlled by the 'User registration' switch. In the new implementation, 'Email as username' is always enabled regardless of the status of 'User registration' - wondering if this was the intent? Also, the old console had a 'Require SSL' dropdown - was this omitted purposely?

Good questions. @yih-wang could you kindly speak to this?

yih-wang:
@mfrances17 @jenny-s51 

> Aside from that, in the old console, the enable/disable of the 'Email as username' switch was controlled by the 'User registration' switch. In the new implementation, 'Email as username' is always enabled regardless of the status of 'User registration' - wondering if this was the intent?

Yes. The email could be used as a username no matter whether we allow the users self-registration. It's not correct in the old console. So we changed that in the new one.

> Also, the old console had a 'Require SSL' dropdown - was this omitted purposely?

This one is moved to the General tab.

mfrances17 (approved):

mfrances17:
thanks @yih-wang and @jenny-s51, your responses addressed my only concerns. LGTM.

jenny-s51:
@ssilvert nice catch, looking into this. Thank you @mfrances17

jonkoops (approved):
LGTM

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -52,6 +52,19 @@ export const RealmSettingsSection = () => {
...
+  const refresh = () => {
+    setKey(new Date().getTime());
+  };
+
+  useEffect(() => {
+    const update = async () => {
+      const realm = await adminClient.realms.findOne({ realm: realmName });
+      setRealm(realm);
+    };
+    setTimeout(update, 1000);
```
jonkoops, line 66: What is the reason there is a `setTimeout` here with a hardcoded value?

src/realm-settings/RealmSettingsSection.tsx
```diff
@@ -52,6 +52,19 @@ export const RealmSettingsSection = () => {
     useState<ComponentRepresentation[]>();
   const [currentUser, setCurrentUser] = useState<UserRepresentation>();
   const { whoAmI } = useWhoAmI();
+  const [key, setKey] = useState(0);
+
+  const refresh = () => {
+    setKey(new Date().getTime());
```
jonkoops, line 58: Instead of using a date this value could also be incremented.
```suggestion
    setKey(key + 1);
```

jonkoops (changes requested):

src/realm-settings/LoginTab.tsx
```diff
@@ -159,24 +209,32 @@ export const RealmSettingsLoginTab = ({
...
+                  labelOff={t("common:off")}
+                  name="duplicateEmailsAllowed"
+                  isChecked={
+                    form.getValues().duplicateEmailsAllowed &&
+                    !form.getValues().loginWithEmailAllowed &&
+                    !form.getValues().registrationEmailAsUsername
+                  }
+                  onChange={(value) => {
+                    save({ ...realm, duplicateEmailsAllowed: value });
+                    onChange(value);
```
ssilvert, line 230: Do you need a call to refresh() here?

src/realm-settings/LoginTab.tsx
```diff
@@ -35,16 +40,25 @@ export const RealmSettingsLoginTab = ({
...
+                  id="kc-user-reg-switch"
+                  data-testid="user-reg-switch"
+                  name="registrationAllowed"
+                  label={t("common:on")}
+                  labelOff={t("common:off")}
+                  isChecked={value}
+                  onChange={(value) => {
+                    save({ ...realm, registrationAllowed: value });
+                    onChange(value);
+                    refresh();
```
ssilvert, line 58: These three lines are repeated over and over.  Suggest you factor it out into a function.

ssilvert (changes requested):

src/realm-settings/RealmSettingsSection.tsx
jenny-s51: @jonkoops Using the `setTimeout` here was suggested by @edewit, due to the bug @ssilvert noticed above. There is a server-side bug that causes the re-fetching to happen too quickly, and therefore the switches are not updated correctly. @edewit suggested to use `setTimeout` to delay the fetching which solved this.

jonkoops: Oh wow, that is really strange. Perhaps you could add a comment to explain this in the code so people do not get confused in the future?

jonkoops (approved):
LGTM

src/realm-settings/RealmSettingsSection.tsx
jenny-s51: Good idea. Done :)

ssilvert (approved):

PR #1186, opened 2021-09-17
Check types when running Cypress

jonkoops:
## Motivation
Allows Cypress to check files for type-errors as this is no longer performed by Snowpack which was disabled in #1168.


agagancarczyk (approved):
lgtm @jonkoops

PR #1187, opened 2021-09-17
deleted original screens,  works like old console

edewit:
These mapper screens are dyanmic
so introduced components that make up the form

fixes: #1099


src/client-scopes/routes/Mapper.ts
```diff
@@ -7,10 +7,11 @@ export type MapperParams = {
   realm: string;
   id: string;
   mapperId: string;
+  type: string;
```
jonkoops, line 10: Let's keep the order in the interface the same as the path.
```suggestion
  id: string;
  type: string;
  mapperId: string;
```

src/client-scopes/details/MappingDetails.tsx
```diff
@@ -163,199 +169,76 @@ export const MappingDetails = () => {
...
           <ActionGroup>
             <Button variant="primary" type="submit">
               {t("common:save")}
             </Button>
-            <Button variant="link">{t("common:cancel")}</Button>
+            <Button
+              variant="link"
+              component={Link}
+              // @ts-ignore
+              to={toClientScope({ realm, id, type, tab: "mappers" })}
```
jonkoops, line 238: Nice! Great that this is a link now!

src/client-scopes/add/components/components.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ScriptComponent } from "./ScriptComponent";
+
+export type ComponentProps = Omit<ConfigPropertyRepresentation, "type">;
+export const ComponentTypes: ReadonlyArray<string> = [
+  "String",
+  "boolean",
+  "List",
+  "Role",
+  "Script",
+];
```
jonkoops, line 17: You can give this one a constant type so that it can be narrowed down further.
```suggestion
export const ComponentTypes = [
  "String",
  "boolean",
  "List",
  "Role",
  "Script",
] as const;
```

src/client-scopes/add/components/components.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type Components = typeof ComponentTypes[number];
+
+export const COMPONENTS: { [index in Components]: FC<ComponentProps> } = {
```
jonkoops, line 21: We're using `FunctionComponent` in other places, let's keep that consistent.
```suggestion
export const COMPONENTS: {
  [index in Components]: FunctionComponent<ComponentProps>;
} = {
```

src/client-scopes/details/MappingDetails.tsx
```diff
@@ -163,199 +169,76 @@ export const MappingDetails = () => {
...
+              </FormGroup>
+            </>
+          )}
+          <FormProvider {...form}>
+            {mapping?.properties.map((property) => {
+              const componentType = property.type!;
+              if (ComponentTypes.includes(componentType)) {
+                const Component = COMPONENTS[componentType as Components];
+                return <Component key={property.name} {...property} />;
+              } else {
```
jonkoops, line 223: Instead of using `includes()` with a type-cast you could write a function that uses a [type-guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to narrow down the type. For example:

```ts
const isValidComponentType = (value: string): value is Components => value in COMPONENTS;
```

Now you can use this function to narrow down the type without needing to cast it.
```suggestion
              if (isValidComponentType(componentType)) {
                const Component = COMPONENTS[componentType];
                return <Component key={property.name} {...property} />;
              } else {
```

src/client-scopes/details/MapperList.tsx
```diff
@@ -33,17 +34,17 @@ type Row = ProtocolMapperRepresentation & {
...
-export const MapperList = ({ clientScope, refresh }: MapperListProps) => {
+export const MapperList = ({ clientScope, type, refresh }: MapperListProps) => {
   const { t } = useTranslation("client-scopes");
   const adminClient = useAdminClient();
   const { addAlert, addError } = useAlerts();
 
   const history = useHistory();
   const { realm } = useRealm();
-  const url = `/${realm}/client-scopes/${clientScope.id}/mappers`;
+  const url = `/${realm}/client-scopes/${clientScope.id}/${type}/mappers`;
```
jonkoops, line 44: Any chance we can use a router function here instead?

src/client-scopes/add/components/RoleComponent.tsx
```diff
@@ -0,0 +1,207 @@
...
+                      setClientsOpen(false);
+                    }}
+                  >
+                    {createSelectGroup(clients)}
+                  </Select>
+                )}
+              </SplitItem>
+              <SplitItem>
+                <Select
+                  onToggle={() => setRoleOpen(!roleOpen)}
```
jonkoops, line 178:
```suggestion
                  onToggle={(isExpanded) => setRoleOpen(isExpanded)}
```

jonkoops (changes requested):

src/client-scopes/add/components/components.ts
edewit: I tried that first, but that doesn't work when you want to do:
```typescript
const componentType = property.type!;
if (ComponentTypes.includes(componentType)) {
  // only supported compnent types
```
because in this case when you use the `as const` then `compnentType` needs to be one of the types so defies the check

src/client-scopes/add/components/components.ts
edewit: `FC` is just shorthand for `FunctionComponent`, from react:
```typescript
type FC<P = {}> = FunctionComponent<P>;
```
but I'll change it

src/client-scopes/add/components/components.ts
jonkoops: Correct, I have a utility function in [another comment](https://github.com/keycloak/keycloak-admin-ui/pull/1187#discussion_r711106658) that fixes that.

src/client-scopes/details/MappingDetails.tsx
edewit: Yes it's a cast, but it can never fail as Components are all the Components that are in ComponentTypes. And your function is also casting by specifying the return type.

jonkoops (approved):

PR #1199, opened 2021-09-21
fix the name of the fields introduced test

edewit:
fixes: #1182

### DIFF
--- a/src/realm-settings/security-defences/BruteForceDetection.tsx
+++ b/src/realm-settings/security-defences/BruteForceDetection.tsx
@@ -62,100 +62,107 @@ export const BruteForceDetection = ({
           )}
         />
       </FormGroup>
-      {enable && (
-        <>
-          <FormGroup
-            label={t("failureFactor")}
-            labelIcon={
-              <HelpItem
-                helpText="realm-settings-help:failureFactor"
-                forLabel={t("failureFactor")}
-                forID="failureFactor"
-              />
-            }
-            fieldId="failureFactor"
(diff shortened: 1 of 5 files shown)


src/realm-settings/security-defences/BruteForceDetection.tsx
```diff
@@ -62,100 +62,107 @@ export const BruteForceDetection = ({
...
+        label={t("failureFactor")}
+        labelIcon={
+          <HelpItem
+            helpText="realm-settings-help:failureFactor"
+            forLabel={t("failureFactor")}
+            forID="failureFactor"
+          />
+        }
+        fieldId="failureFactor"
+        style={enable ? {} : { display: "none" }}
```
jonkoops, line 75: Instead of using styling to hide these fields why not hide them using conditional logic?

edewit: because that doesn't register the fields so data loaded will not be displayed

mfrances17 (approved):
LGTM, code looks ok and was able to save successfully several times after changing several fields, including `Wait increment`, and no errors and was able to retain values

src/realm-settings/security-defences/BruteForceDetection.tsx
jonkoops: You can configure the forms library to take care of that I believe. I remember @jenny-s51 had to pass some flag to dynamically render the fields in another PR.

PR #1200, opened 2021-09-21
fixes issue with selected locals

edewit:
fixes: #1179


mfrances17 (approved):
LGTM, code looks good and supported locales works perfectly now when navigating away and back.

ssilvert:
@edewit It looks like something in master broke Localization.  When I enable Internationalization,  `Cannot read properties of undefined (reading 'map')`

Can you take a look?  Maybe it makes sense to just add the fix to this PR since I can't try it out without a fix.

edewit:
@ssilvert I can't reproduce that, though I'm still on keycloak 15.0.2

PR #1201, opened 2021-09-21
Realm Settings(Client Policies): Add policies list

jenny-s51:
## Motivation

https://issues.redhat.com/browse/CLUX-262

## Brief Description

Add policies list to client policies

## Verification Steps

1. Go to old admin console.
2. Go to Realm settings -> Client policies -> policies subtab
3. Click "Create" and fill out the form fields. 
4. Save the new policies.
5. Go to the new admin console
6. Go to Realm settings -> Client policies -> policies subtab
7. Verify the policies you created are listed in the table correctly.

## Checklist:

## Additional Notes

- Enabled/disabled switches, and create/update/delete functionality will be implemented once https://github.com/keycloak/keycloak-nodejs-admin-client/pull/258 is merged
- Tests will be written once CRUD functionality can be implemented for policies (also dependent on https://github.com/keycloak/keycloak-nodejs-admin-client/pull/258)

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -129,8 +129,6 @@ describe("Realm settings", () => {
     sidebarPage.goToRealmSettings();
     cy.findByTestId("rs-email-tab").click();
 
-    cy.wait(1000);
-
     realmSettingsPage.addSenderEmail("example@example.com");
 
     realmSettingsPage.toggleCheck(realmSettingsPage.enableSslCheck);
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.1",
       "license": "Apache",
       "dependencies": {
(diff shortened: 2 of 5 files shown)


mfrances17 (changes requested):
Code looks good and testing using the steps provided worked as expected.

One odd thing that should be checked - in the old console, the column name was `Enabled`, and the values were `True` and `False`. In the new console, it looks like it was renamed to `Status`, but it still has the same values of `True` and `False`. To me, that new wording is not natural and a bit weird, e.g. `Status` is `True`. 

I would think if you would want `Status` as the column name, it should be `Enabled` or `Disabled`. Otherwise I would keep the original `Enabled` - `True/False`.

jenny-s51:
Thank you for your feedback @mfrances17 ! I agree; my next PR will include switches instead of those `True` / `False` values. I put "Status" there since that's the column name in the mock, but until then I'm reverting back to "Enabled" per your suggestion. 👍

mfrances17 (approved):
LGTM, thanks @jenny-s51

PR #1207, opened 2021-09-23
Run the Cypress tests on master of Keycloak Server

### LINKED ISSUE #1036: Run Cypress tests against Keycloak master
In order to guarantee that we are compatible with the version of Keycloak that will be shipped we will have to run our Cypress tests against the latest version of Keycloak from master.

jonkoops:
## Motivation
Builds the latest version of Keycloak Server and runs the Cypress tests on top of it. This ensures that we remain compatible with Keycloak before release, as opposed to using the latest stable version, catching regressions in the process.

This also runs the complete integration of the Admin UI and Keycloak as we would for production by building the theme locally and installing it as a dependency when building Keycloak.

Closes #1036

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -7,21 +7,59 @@ on:
 jobs:
   cypress-run:
     runs-on: ubuntu-latest
+    env:
+      ADMIN_UI_PATH: admin-ui
+      KEYCLOAK_SERVER_PATH: keycloak-server
     steps:
-      - uses: actions/checkout@v2
+      - name: Check out Admin UI
+        uses: actions/checkout@v2
+        with:
+          path: ${{ env.ADMIN_UI_PATH }}
 
-      - name: Build Keycloak server
-        run: docker build --tag keycloak .
+      - name: Check out Keycloak Server
+        uses: actions/checkout@v2
+        with:
+          repository: keycloak/keycloak
+          path: ${{ env.KEYCLOAK_SERVER_PATH }}
 
-      - name: Start Keycloak server
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: temurin
+          java-version: 11
+          check-latest: true
+          cache: maven
+
+      # See https://issues.redhat.com/browse/KEYCLOAK-17812
+      - name: Update Maven settings
+        working-directory: ${{ env.KEYCLOAK_SERVER_PATH }}
+        run: cp .github/settings.xml ~/.m2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress.yml
```diff
@@ -7,45 +7,86 @@ on:
...
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "KEYCLOAK_SERVER_BIN=keycloak-server/distribution/server-dist/target/keycloak-$VERSION/bin" >> $GITHUB_ENV
 
-      - name: Start Keycloak server
+      - name: Start Keycloak Server
+        working-directory: ${{ env.KEYCLOAK_SERVER_BIN }}
         run: |
-          docker run --detach --publish 8080:8080 --env KEYCLOAK_USER=admin --env KEYCLOAK_PASSWORD=admin --name keycloak keycloak
-          sleep 10
+          ./add-user-keycloak.sh --user admin --password admin
```
edewit, line 55: Does this command work when the server is not running?

.github/workflows/cypress.yml
```diff
@@ -7,45 +7,86 @@ on:
...
+          ./standalone.sh -Dkeycloak.profile.feature.admin2=enabled &
 
-      - name: Configure Keycloak server
+      - name: Configure Keycloak Server
+        working-directory: ${{ env.KEYCLOAK_SERVER_BIN }}
         run: |
-          docker exec keycloak /opt/jboss/keycloak/bin/kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin
-          docker exec keycloak /opt/jboss/keycloak/bin/kcadm.sh update realms/master -s "adminTheme=keycloak.v2"
+          ./kcadm.sh config credentials --server http://localhost:8080/auth --realm master --user admin --password admin
```
edewit, line 61: wait, is this creating the admin user also?

edewit (approved):

.github/workflows/cypress.yml
edewit: ahh no that is logging in

.github/workflows/cypress.yml
jonkoops: Yes, it writes some information to a JSON file that contains the credentials. Once the server starts it will read this file and create the user.

.github/workflows/cypress.yml
jonkoops: Yes, just a login. The user is created before this.

Issue #1216, opened 2021-09-23
Custom fields for mapper types need cleanup


ssilvert:
1. Go to Client scopes --> Mappers --> Add mapper --> By configuration
2. Select a mapper type such as "Javascript Mapper"
3. This mapper has one custom field called "Script".
4. The help text for this custom field is incorrect.

There are similar issues on other of these custom fields.  For some the help text is wrong,  For others, the field name is not found in the message bundle (example is `name-id-format` for `User Attribute Mapper For NameID` mapper type.

Someone needs to go through all these and make sure the help text and attribute names match the old console.

edewit:
Seem to me the help text of the Javascript Mapper is correct, old console:
![image](https://user-images.githubusercontent.com/51133/134636216-d6db3a03-8dd5-429d-a843-223981396bf9.png)
New console:
![image](https://user-images.githubusercontent.com/51133/134636313-c200105d-9d5c-41bb-a18c-a7a5a50add05.png)

edewit:
oh wait I see it's only part of the text... and let's format the newlines as well

edewit:
there is also a missing label on keycloak itself [KEYCLOAK-19329](https://issues.redhat.com/browse/KEYCLOAK-19329)

PR #1217, opened 2021-09-24
Added missing labels on the client scope mappers

edewit:
fixes: #1216

### DIFF
--- a/src/client-scopes/add/components/ScriptComponent.tsx
+++ b/src/client-scopes/add/components/ScriptComponent.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Fragment } from "react";
 import { Controller, useFormContext } from "react-hook-form";
 import { useTranslation } from "react-i18next";
 import { FormGroup } from "@patternfly/react-core";
@@ -20,7 +20,20 @@ export const ScriptComponent = ({
     <FormGroup
       label={t(label!)}
       labelIcon={
-        <HelpItem helpText={t(helpText!)} forLabel={t(label!)} forID={name!} />
+        <HelpItem
+          helpText={
+            <span>
+              {helpText?.split("\n").map((item, key) => (
+                <Fragment key={key}>
+                  {item}
+                  <br />
+                </Fragment>
+              ))}
+            </span>
+          }
+          forLabel={t(label!)}
+          forID={name!}
+        />
       }
       fieldId={name!}
     >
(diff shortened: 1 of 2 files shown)


edewit:
formatted the help text on the Javascript mapper:
image

src/client-scopes/add/components/ScriptComponent.tsx
```diff
@@ -20,7 +20,20 @@ export const ScriptComponent = ({
     <FormGroup
       label={t(label!)}
       labelIcon={
-        <HelpItem helpText={t(helpText!)} forLabel={t(label!)} forID={name!} />
+        <HelpItem
+          helpText={
+            <span>
+              {helpText?.split("\n").map((item, key) => (
+                <Fragment key={key}>
```
jonkoops, line 27: Instead of using the index as a key perhaps you can use the item itself?
```suggestion
                <Fragment key={item}>
```

src/client-scopes/add/components/ScriptComponent.tsx
```diff
@@ -20,7 +20,20 @@ export const ScriptComponent = ({
...
+                  {item}
+                  <br />
+                </Fragment>
+              ))}
+            </span>
+          }
```
jonkoops, line 33: If you are looking to preserve the white-space, perhaps you can wrap this in a `pre` tag instead?
```suggestion
          helpText={<pre>{t(helpText!)}</pre>}
```

jonkoops (changes requested):

ssilvert:
Group test should be fixed now if you rebase.

src/client-scopes/add/components/ScriptComponent.tsx
edewit: Doesn't really matter as it's a static list

src/client-scopes/add/components/ScriptComponent.tsx
edewit: a `<pre` will use another font and mess up the layout:
image

src/client-scopes/add/components/ScriptComponent.tsx
jonkoops: True, still better not to do it out of convention. I'd like to add an ESLint rule for this at some point as well.

src/client-scopes/add/components/ScriptComponent.tsx
jonkoops: Ahhh yes. We can use some styling here instead by adding `white-space: pre-wrap;` to the containing element.

jonkoops (approved):

PR #1245, opened 2021-09-28
Hide certain elements on the page if client is a realm 

jonkoops:
See the [design](https://marvelapp.com/prototype/5di1ce1/screen/81685254)


edewit (changes requested):
could this be tested with a cypress test?

edewit (approved):

PR #1246, opened 2021-09-28
Small fix on policies tab

edewit:
When a request is active and the user navigates away from the screen that triggered the request we need to cancel that request
and not set the state as that would create memory issues.

This PR changes how the state is set by using `useFetch` instead of setting the sate in the loader.

### DIFF
--- a/src/realm-settings/ProfilesTab.tsx
+++ b/src/realm-settings/ProfilesTab.tsx
@@ -6,6 +6,7 @@ import {
   ButtonVariant,
   Label,
   PageSection,
+  Spinner,
   ToolbarItem,
 } from "@patternfly/react-core";
 import { Divider, Flex, FlexItem, Radio, Title } from "@patternfly/react-core";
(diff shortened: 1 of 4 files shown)


src/realm-settings/ProfilesTab.tsx
```diff
@@ -38,31 +40,33 @@ export const ProfilesTab = () => {
...
-    setTableProfiles(allClientProfiles);
+      const allClientProfiles = globalProfiles?.concat(profiles ?? []);
+      setTableProfiles(allClientProfiles || []);
+    },
+    [key]
+  );
 
-    return allClientProfiles ?? [];
-  };
+  const loader = async () => Promise.resolve(tableProfiles || []);
```
jonkoops, line 69: Same here, this is an async function you should be able to return the value directly.
```suggestion
  const loader = async () => tableProfiles ?? [];
```

jonkoops (changes requested):

jonkoops (approved):

PR #1270, opened 2021-09-29
Add tests for bearer-only client details

jonkoops:
## Motivation
I added some logic to show a tooltip for bearer-only clients that was not yet tested, this PR adds the missing tests.

### DIFF
--- a/cypress/integration/clients_test.spec.ts
+++ b/cypress/integration/clients_test.spec.ts
@@ -258,4 +258,33 @@ describe("Clients test", function () {
       cy.findByTestId("delete-client").should("not.exist");
     });
   });
+
+  describe("Bearer only", () => {
+    const clientId = "bearer-only";
+
+    before(() => {
+      new AdminClient().createClient({
+        clientId,
+        protocol: "openid-connect",
+        publicClient: false,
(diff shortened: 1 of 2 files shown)


agagancarczyk (approved):

cypress/integration/clients_test.spec.ts
```diff
@@ -258,4 +258,33 @@ describe("Clients test", function () {
...
+  describe("Bearer only", () => {
+    const clientId = "bearer-only";
+
+    before(() => {
+      new AdminClient().createClient({
+        clientId,
+        protocol: "openid-connect",
+        publicClient: false,
+        bearerOnly: true,
+      });
```
edewit, line 271: great way to get a bearer only client

edewit (approved):

ssilvert (changes requested):
I found an issue with the bearer only client.  The fix looks easy, but we need a test for the use case so you may want to include both the fix and the new test with this PR.

https://github.com/keycloak/keycloak-admin-ui/issues/1276

jonkoops:
@ssilvert I'll create some separate PR for those issues, I don't want to pull in too much logic into a single PR.

Issue #1114, opened 2021-09-02
[UX Review] The Test button shouldn't report success when the Bind DN/credential is wrong


yih-wang:
Location: User federation -> Create provider

When the Bind type is Simple, and the bind credential isn't been input. The system should report failure instead of success when the user presses the Test button. 

![image](https://user-images.githubusercontent.com/23331885/131820945-9cfe071b-26df-46fb-af59-0fde6f5678f9.png)

Also, when the user inputs a wrong Bind DN, the Test button reports success.

jonkoops:
I don't know if we actually have control over this message is the result of an API call ([see the code](https://github.com/keycloak/keycloak-admin-ui/blob/master/src/user-federation/ldap/LdapSettingsConnection.tsx#L55-L71)).

@yih-wang Can you share with me the form values you are using to reproduce this issue?

yih-wang:
> I don't know if we actually have control over this message is the result of an API call (see the code).

@jonkoops Yes, I think so. The alert works well in the old console.

So the issue is, we used to have two different buttons to test connection & authentication separately in the old console (see the screenshot below). 
![image](https://user-images.githubusercontent.com/23331885/132342887-2c62d7e6-f426-456f-9a6b-89e96927fe25.png)
In the new design, there is only one button left to test all the things. However, in the current implementation, this button only validates the connection, but not the authentication.

To reproduce that, you can type "ldaps://ldap.jumpcloud.com:636" as the Connection URL (to avoid a connection error), and then directly click Test at the bottom of this section. You'll find it successful without validating the Bind DN or credential.

Having said that, I'm kind of thinking that we should still provide two separate buttons in the new console instead of one. During user testing, we found that LDAP is complicated for most of the users so that more validations can give them more confidence. @ssilvert Do you have an opinion on that? Do you think it makes more sense to validate connection and authentication separately, like the screen below?
![image](https://user-images.githubusercontent.com/23331885/132345401-a9be2d6a-dd50-481c-8ea4-985b136b2d34.png)

PR #1275, opened 2021-09-29
Updated the test ldap settings with second button

edewit:
fixes: #1114

### DIFF
--- a/src/client-scopes/add/MapperDialog.test.tsx
+++ b/src/client-scopes/add/MapperDialog.test.tsx
@@ -1,13 +1,19 @@
 /**
  * @jest-environment jsdom
  */
-import { Button } from "@patternfly/react-core";
-import { fireEvent, render, screen } from "@testing-library/react";
-import type { ServerInfoRepresentation } from "@keycloak/keycloak-admin-client/lib/defs/serverInfoRepesentation";
 import React, { useState } from "react";
+import "@testing-library/jest-dom";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { Button } from "@patternfly/react-core";
+
+import type { ServerInfoRepresentation } from "@keycloak/keycloak-admin-client/lib/defs/serverInfoRepesentation";
+import type WhoAmIRepresentation from "@keycloak/keycloak-admin-client/lib/defs/whoAmIRepresentation";
 import { ServerInfoContext } from "../../context/server-info/ServerInfoProvider";
 import serverInfo from "../../context/server-info/__tests__/mock.json";
 import { AddMapperDialog, AddMapperDialogModalProps } from "./MapperDialog";
+import { WhoAmI, WhoAmIContext } from "../../context/whoami/WhoAmI";
+
+import whoami from "../../context/whoami/__tests__/mock-whoami.json";
 
 describe("MapperDialog", () => {
   const Test = (args: AddMapperDialogModalProps) => {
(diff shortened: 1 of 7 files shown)


src/client-scopes/add/MapperDialog.tsx
```diff
@@ -42,21 +44,24 @@ export const AddMapperDialog = (props: AddMapperDialogProps) => {
...
+    .map((mapper) => {
+      const mapperType = protocolMappers.filter(
+        (type) => type.id === mapper.protocolMapper
+      )[0];
+      return {
+        item: mapper,
+        selected: false,
+        cells: [mapper.name, mapperType.helpText],
+      };
+    });
```
jonkoops, line 64: This is quite an expensive operation to do on every render, I would recommend that you put it in a `useMemo()`.

src/client-scopes/add/MapperDialog.tsx
```diff
@@ -118,32 +127,45 @@ export const AddMapperDialog = (props: AddMapperDialogProps) => {
...
+                  <DataListCell style={{ fontWeight: 700 }} key={name}>
+                    {name}
+                  </DataListCell>
+                ))}
+              />
+            </DataListItemRow>
+          </DataListItem>
+          {protocolMappers
+            .sort((a, b) => a.name!.localeCompare(b.name!, whoAmI.getLocale()))
+            .map((mapper) => (
```
jonkoops, line 143: This is quite an expensive operation to do on every render, I would recommend to use `useMemo()` here as well.

src/user-federation/ldap/LdapSettingsConnection.tsx
```diff
@@ -40,6 +40,19 @@ const testLdapProperties: Array<keyof TestLdapConnectionRepresentation> = [
...
 ];
 
+type TestTypes = "testConnection" | "testAuthentication";
+
+const convertFormToSettings = (form: UseFormMethods) => {
+  const settings: TestLdapConnectionRepresentation = {};
+
+  testLdapProperties.forEach((key) => {
+    const value = _.get(form.getValues(), `config.${key}`);
+    settings[key] = _.isArray(value) ? value[0] : "";
```
jonkoops, line 50: You can also use [`Array.isArray()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) here instead of lodash.
```suggestion
    settings[key] = Array.isArray(value) ? value[0] : "";
```

jonkoops (changes requested):

src/client-scopes/add/MapperDialog.tsx
jonkoops: @edewit this is still being done on each render.

jonkoops (changes requested):
Looks like the tests are failing

src/client-scopes/add/MapperDialog.test.tsx
```diff
@@ -1,13 +1,19 @@
 /**
  * @jest-environment jsdom
  */
-import { Button } from "@patternfly/react-core";
+import React, { useState } from "react";
+import "@testing-library/jest-dom";
```
jonkoops, line 5: There is no need to import this from the testing library as this is already included in the Jest setup.

jonkoops (approved):

src/client-scopes/add/MapperDialog.test.tsx
jonkoops: Pushed a commit to fix this

edewit: It didn't work for me without it

PR #1278, opened 2021-09-30
Remove ability to delete a realm client from the overview

jonkoops:
## Motivation
Clients that are intended to be used for authenticating a realm are not allowed to be deleted.


xianli123 (approved):
Great work.

xianli123:
In the client list, the type of the realm-client should be openid-connect.

jonkoops:
> In the client list, the type of the realm-client should be openid-connect.

@xianli123 Yeah, it should also have a nicer human readable name. I'll add this to #1283

agagancarczyk (approved):

PR #1279, opened 2021-09-30
Hide capability config for bearer-only clients

### LINKED ISSUE #1277: Bearer only clients should not have Capability config
Bearer only clients are including the `Capability config` section on the Client settings tab.  This section should not be there and should also be removed from the `Jump to section`.

For regular bearer only clients, it is showing the section and all the settings.  For realm clients in master, it is showing the `Capability config` section header with no fields.

jonkoops:
Closes #1277

### DIFF
--- a/src/components/scroll-form/ScrollForm.tsx
+++ b/src/components/scroll-form/ScrollForm.tsx
@@ -31,29 +31,33 @@ export const ScrollForm: FunctionComponent<ScrollFormProps> = ({
   ...rest
 }) => {
   const { t } = useTranslation("common");
-
   const nodes = Children.toArray(children);
+
   return (
     <Grid hasGutter {...rest}>
       <GridItem span={8}>
-        {sections.map((cat, index) => (
-          <Fragment key={cat}>
-            {!borders && (
-              <ScrollPanel scrollId={spacesToHyphens(cat)} title={cat}>
-                {nodes[index]}
-              </ScrollPanel>
-            )}
-            {borders && (
-              <FormPanel
-                scrollId={spacesToHyphens(cat)}
-                title={cat}
-                className="kc-form-panel__panel"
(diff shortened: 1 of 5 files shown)


xianli123:
Hi @jonkoops, As for bearer-only client, not only the capability configs have been removed, but also the "Login settings" have been removed. In addition the access settings only keep the "Admin URL".  The client-realm also has some changes than the openid-connect client. We may file some different issues to fix them. WDYT? 

Here are the designs FYI. 
Bearer-only clients - https://marvelapp.com/prototype/5di1ce1/screen/81685256
Bearer-only realm-clients - https://marvelapp.com/prototype/5di1ce1/screen/81685255

xianli123 (approved):
The capability configs have been removed.

jonkoops:
> As for bearer-only client, not only the capability configs have been removed, but also the "Login settings" have been removed. In addition the access settings only keep the "Admin URL". The client-realm also has some changes than the openid-connect client. We may file some different issues to fix them. WDYT?

That is peculiar, I will take a look next week to see if I can resolve these under this PR.

src/components/scroll-form/ScrollForm.tsx
```diff
@@ -67,7 +67,11 @@ export const ScrollForm: FunctionComponent<ScrollFormProps> = ({
           >
             {sections.map((cat) => (
               // note that JumpLinks currently does not work with spaces in the href
-              <JumpLinksItem key={cat} href={`#${spacesToHyphens(cat)}`}>
+              <JumpLinksItem
+                key={cat}
+                href={`#${spacesToHyphens(cat)}`}
```
edewit, line 72: could do `spacesTohyphens` once instead of two times

edewit (approved):

PR #1287, opened 2021-10-01
Prefer using arrow functions for callbacks

jonkoops:
## Motivation
Use arrow functions for all callbacks and enforce this as a rule. Came up as a point of discussion between me and @edewit and we thought this might be a good path.


edewit (approved):
Yes this is cool

jenny-s51 (approved):
Nice addition @jonkoops - good to have this. LGTM 👍

agagancarczyk (approved):

Issue #1352, opened 2021-10-14
OIDC mapper: Help text broken


ssilvert:
Go to Identity providers --> OIDC provider --> Edit Identity Provider Mapper

Help text is broken (or doesn't match old console) for the following fields in OIDC mappers/types:
**Advanced Claim to Role:** Claims
**Attribute Importer:** Mapper type
**Advanced Claim to Role:** Claims
**Advanced Claim to Role:** Regex Attribute Values
**Hardcoded Role:** Role
**Claim to Role**: Claim Value
**Claim to Role:** Role

PR #1443, opened 2021-10-28
changed to use mapping components

edewit:
fixes: #1352


mfrances17 (changes requested):
if you add a bad discovery endpoint, and then decide to not use a discovery endpoint e.g. toggle `Use discovery endpoint` to off, the whole app crashes and you need to restart:

discovery-endpoint-to-off

edewit:
closed in favour of #1462

PR #1449, opened 2021-10-29
Fix notification-related cypress failures without waits

mfrances17:
## Motivation
Follow-up to https://github.com/keycloak/keycloak-admin-ui/pull/1438, and provides a fix for the issue in which notifications can stack to a point that they will interfere with attempted actions on a modal. This solution dismisses the notifications instead of adding timed waits.

## Brief Description
See https://github.com/keycloak/keycloak-admin-ui/pull/1438 for a more descriptive explanation of the problem. This solution adds an optional closeNotification parameter to the existing checkNotificationMessage function, so when you call the function to check the content of a notification message and pass true, it will dismiss the notification after checking the text. I chose to make it optional so it won't affect any legacy tests in which this is not an issue.

## Verification Steps
Verify that the ldap mapper test no longer fails locally or in CI.

## Checklist:

## Additional Notes
None

### DIFF
--- a/cypress/support/pages/admin_console/Masthead.ts
+++ b/cypress/support/pages/admin_console/Masthead.ts
@@ -50,9 +50,11 @@ export default class Masthead {
     cy.get("#manage-account").click();
   }
 
-  checkNotificationMessage(message: string) {
+  checkNotificationMessage(message: string, closeNotification?: boolean) {
     cy.contains(message).should("exist");
-
+    if (closeNotification) {
+      cy.get(".pf-c-alert__action").click();
+    }
     return this;
   }
 
--- a/cypress/integration/user_fed_ldap_mapper_test.spec.ts
+++ b/cypress/integration/user_fed_ldap_mapper_test.spec.ts
@@ -151,39 +151,33 @@ describe("User Fed LDAP mapper tests", () => {
 
     listingPage.itemExist(creationDateMapper).deleteItem(creationDateMapper);
     modalUtils.checkModalTitle(mapperDeleteTitle).confirmModal();
-    masthead.checkNotificationMessage(mapperDeletedSuccess);
+    masthead.checkNotificationMessage(mapperDeletedSuccess, true);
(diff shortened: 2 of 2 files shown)


cypress/support/pages/admin_console/Masthead.ts
```diff
@@ -50,9 +50,11 @@ export default class Masthead {
     cy.get("#manage-account").click();
   }
 
-  checkNotificationMessage(message: string) {
+  checkNotificationMessage(message: string, closeNotification?: boolean) {
```
jonkoops, line 53: You can specify a default here as well.
```suggestion
  checkNotificationMessage(message: string, closeNotification = false) {
```

cypress/support/pages/admin_console/Masthead.ts
```diff
@@ -50,9 +50,11 @@ export default class Masthead {
...
     cy.contains(message).should("exist");
-
+    if (closeNotification) {
+      cy.get(".pf-c-alert__action").click();
```
jonkoops, line 56: Should this not be only the notification that matches the message? This way it might close another notification. Something like:
```suggestion
      cy.contains(message).closest(".pf-c-alert__action").click();
```

jonkoops (approved):

cypress/support/pages/admin_console/Masthead.ts
mfrances17: i did go down this exact road of matching the exact message but it was failing... i can take another look

jonkoops: Hmmm, yes this seems to fail quite hard. I wonder why, perhaps it's a sibling element?

mfrances17: not sure, but i did do a bunch of testing with this PR's implementation and it works flawlessly, so i went with it. and there aren't arbitrary waits anymore, so i do think it's a viable solution for the test suite.

Issue #1025, opened 2021-08-19
Identity providers should use useFetch instead of useEffect


mfrances17:
Both the OIDC and SAML providers use the useEffect() hook, but we are trying to standardize on useFetch() within the keycloak admin console UI.

PR #1460, opened 2021-11-01
removed useEffect data load instead uses fetch

edewit:
fixes: #1025

### DIFF
--- a/src/user-federation/UserFederationKerberosSettings.tsx
+++ b/src/user-federation/UserFederationKerberosSettings.tsx
@@ -37,7 +37,7 @@ const KerberosSettingsHeader = ({
   toggleDeleteDialog,
 }: KerberosSettingsHeaderProps) => {
   const { t } = useTranslation("user-federation");
-  const { id } = useParams<{ id: string }>();
+  const { id } = useParams<{ id?: string }>();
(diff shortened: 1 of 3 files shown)


jonkoops (approved):

mfrances17 (changes requested):
i'm no longer able to create a new kerberos user fed provider with these changes..

kerberos

the routing will probably need to change if you want to implement changes like this, i assume. it's looking for the 'new' resource instead of doing the proper routing to add if it's 'new'.

edewit:
@mfrances17 don't know what your issue is, but as you can see I didn't update the routing

mfrances17 (approved):
I'll run a clean and retest, it could be an issue on my end.

mfrances17:
@edewit actually, the kerberos user fed test is failing with these changes, that test has been rock solid to this point. i would verify that you did not break the test.

mfrances17 (changes requested):
Changing from approved to request changes just until i retest

mfrances17:
@edewit yes, this is still failing after a clean build. did you test creating a new Kerberos user fed provider? these changes break it, which is why i assume the test is failing as well. you did not change the route, but the existing route depends on 'new'.

mfrances17 (approved):
LGTM, issue with creating a new kerberos provider is fixed, works great now

PR #1464, opened 2021-11-01
IDPs: go to edit form after save()

### LINKED ISSUE #1440: IDP Mappers: Can edit read-only fields after saving
1. Go to Identity providers --> OIDC provider --> Add Identity Provider Mapper
2. Fill in fields and save
3. You will still be on the "create" screen
4. You are now allowed to edit `Name` and `Mapper Type`, which should be read-only fields after saving.  Also, the ID is not shown.

The simple solution is to do what the old console does, which is to just return to the list after saving.

jenny-s51:
Closes #1440

### DIFF
--- a/src/identity-providers/add/AddMapper.tsx
+++ b/src/identity-providers/add/AddMapper.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo, useState } from "react";
-import { Link, useParams } from "react-router-dom";
+import { Link, useHistory, useParams } from "react-router-dom";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


jonkoops (approved):
LGTM

edewit (changes requested):
test seems to fail

jonkoops (approved):

mfrances17 (approved):
LGTM, changes work as expected. 

Side note... this is tagged as a Post-1.0 Issue, we should be fixing 1.0 issues first before we tackle these.

edewit (approved):

PR #1473, opened 2021-11-02
Use route functions for navigation in authentication

jonkoops:


edewit (approved):
route functions everywhere!

PR #1474, opened 2021-11-02
Use route functions for navigation in groups

jonkoops:


edewit (approved):

jenny-s51 (approved):
Routes are looking good @jonkoops ! Looks like the data-testid for searchGroups no longer exists on the dropdown item so maybe we can use a `.contains()` in the test instead

mfrances17 (changes requested):
LGTM... just verify that the groups test that is failing is not due to the changes in this PR.

jonkoops:
@mfrances17 The tests should pass properly now, could you take another look?

mfrances17 (approved):
LGTM... all group tests are passing - there are other test failures but they look completely unrelated. Approving and merging.

Issue #1128, opened 2021-09-06
Remove unnecessary conditions from code


jonkoops:
In various places there is code that has [unnecessary conditions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md). This leads to a poor implementation and incorrect types and should not occur. 

The following needs to happen before this issue can be closed:
1. Refactor code that is flagged by the `@typescript-eslint/no-unnecessary-condition` rule so it no longer produces warnings (this can be split up in multiple PRs).
2. Change the same rule to  the `error` level so linting fails when unnecessary conditions are introduced.

alevid99:
Hi @jonkoops to resolve one of the warnings related to the groupPickerDialog.
I need to know if is possible to move a child group to "root" to know how to refactor the code.
If this action should be possible a bugFix is needed otherwise we can simply disable the button.

Actually the button in the dialog is enabled but when you click the button nothing happen (you get a success message but you should get an error). On the confirmation callback the code crash (src/groups/GroupTable.tsx:214) the error get catch but due to the check src/groups/GroupTable.tsx:223 no error is raised.

jonkoops:
Hmmm, yeah that sounds like it could be a separate issue. Could you create one and describe the problem there?

alevid99:
I found this issue #1224 that track this problem. I can comment with more details there

ssilvert:
> Hi @jonkoops to resolve one of the warnings related to the groupPickerDialog. I need to know if is possible to move a child group to "root" to know how to refactor the code. If this action should be possible a bugFix is needed otherwise we can simply disable the button.

Move to root should be possible.  So yes, it's a bug.

PR #1476, opened 2021-11-02
refactor: unecessary condition #1128

alevid99:
## Motivation

Resolve warnings for issue #1128 

## Checklist:

### DIFF
--- a/src/client-scopes/add/components/ClientSelectComponent.tsx
+++ b/src/client-scopes/add/components/ClientSelectComponent.tsx
@@ -36,7 +36,6 @@ export const ClientSelectComponent = ({
       };
       if (search) {
         params.clientId = search;
-        params.search = true;
       }
       return adminClient.clients.find(params);
     },
--- a/src/client-scopes/form/ClientScopeForm.tsx
+++ b/src/client-scopes/form/ClientScopeForm.tsx
@@ -130,9 +130,7 @@ export default function ClientScopeForm() {
           clientScopes.type
         );
         history.push(
-          `/${realm}/client-scopes/${scope.id}/${
-            clientScopes.type || "none"
-          }/settings`
+          `/${realm}/client-scopes/${scope.id}/${clientScopes.type}/settings`
         );
       }
       addAlert(t((id ? "update" : "create") + "Success"), AlertVariant.success);
--- a/src/clients/ClientsSection.tsx
+++ b/src/clients/ClientsSection.tsx
@@ -51,7 +51,6 @@ export default function ClientsSection() {
     };
     if (search) {
       params.clientId = search;
-      params.search = true;
     }
     return await adminClient.clients.find({ ...params });
   };
--- a/src/components/multi-line-input/MultiLineInput.tsx
+++ b/src/components/multi-line-input/MultiLineInput.tsx
@@ -15,13 +15,13 @@ export type MultiLine = {
 };
 
 export function convertToMultiline(fields: string[]): MultiLine[] {
-  return (fields && fields.length > 0 ? fields : [""]).map((field) => {
+  return (fields.length > 0 ? fields : [""]).map((field) => {
     return { value: field };
   });
 }
 
 export function toValue(formValue: MultiLine[]): string[] {
-  return formValue?.map((field) => field.value);
+  return formValue.map((field) => field.value);
 }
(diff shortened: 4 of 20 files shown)


src/client-scopes/add/components/ClientSelectComponent.tsx
```diff
@@ -36,7 +36,6 @@ export const ClientSelectComponent = ({
       };
       if (search) {
         params.clientId = search;
-        params.search = true;
```
jonkoops, line 39: What is the reason this was removed? I am not seeing any warnings that this is an unnecessary condition.

src/client-scopes/form/ClientScopeForm.tsx
```diff
@@ -130,9 +130,7 @@ export default function ClientScopeForm() {
           clientScopes.type
         );
         history.push(
-          `/${realm}/client-scopes/${scope.id}/${
-            clientScopes.type || "none"
-          }/settings`
+          `/${realm}/client-scopes/${scope.id}/${clientScopes.type}/settings`
```
jonkoops, line 133: This will actually cause an error as `clientScopes.type` can be an empty string and thus needs a fallback. I think we should provide a default value to the form to fix this.

src/clients/ClientsSection.tsx
```diff
@@ -51,7 +51,6 @@ export default function ClientsSection() {
     };
     if (search) {
       params.clientId = search;
-      params.search = true;
```
jonkoops, line 54: Why was this removed? I don't see any warnings for this one.

src/components/multi-line-input/MultiLineInput.tsx
```diff
@@ -76,7 +76,7 @@ export const MultiLineInput = ({
               onClick={() => append({})}
               tabIndex={-1}
               aria-label={t("common:add")}
-              isDisabled={rest.isDisabled || !currentValues?.[index]?.value}
```
jonkoops, line 79: This might be a false positive as we have [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) disabled.

src/realm-roles/AssociatedRolesTab.tsx
```diff
@@ -67,12 +67,9 @@ export const AssociatedRolesTab = ({
     return [...result];
   };
 
-  const loader = async (first?: number, max?: number, search?: string) => {
+  const loader = async () => {
     const compositeRoles = await adminClient.roles.getCompositeRoles({
       id: parentRole.id!,
-      first: first,
-      max: max!,
-      search: search,
```
jonkoops, line 75: Why were these removed? I don't see a reason to remove the filters here. Although this could be simplified.

```ts
const compositeRoles = await adminClient.roles.getCompositeRoles({
  id: parentRole.id!,
  first,
  max,
  search,
});
```

jonkoops (changes requested):
It's hard to tell which of these changes might break the application as I don't know if the types we are using are all that correct.

src/client-scopes/add/components/ClientSelectComponent.tsx
alevid99: @jonkoops there is an error due to the typescript type of params

src/clients/ClientsSection.tsx
alevid99: @jonkoops there is an error due to the typescript type of params

src/realm-roles/AssociatedRolesTab.tsx
alevid99: @jonkoops there is an error due to the typescript type. All this field are not present on the input type of the method.

jonkoops:
I'm closing this PR as we currently have to focus on other priorities (delivering the first version of the admin console) and thus cannot afford the time to test the changes right now. We intend to fix the issues with this code progressively as we refactor it over time.

@alevid99 I want to thank you a lot for contributing to the project, we appreciate your efforts a lot but we simply do not have the resources to resolve this issue at this point in time. If you are looking for issues to pick, we have a [milestone](https://github.com/keycloak/keycloak-admin-ui/milestone/1) for things we need to get in before our first release which would help us out greatly.

PR #1495, opened 2021-11-04
Realm settings(Client policies -> policies): Add switch for enabling/disabling policy

### LINKED ISSUE #1389: Client policy details does not have enabled/disabled switch
When a client policy is created, it should be enabled by default.  Currently it is created in a disabled state.

Also, there is an enabled/disabled switch in the upper right corner of the details screen.  See https://marvelapp.com/prototype/2b3ghc06/screen/81619372

jenny-s51:
Closes #1389

### DIFF
--- a/cypress/integration/realm_settings_test.spec.ts
+++ b/cypress/integration/realm_settings_test.spec.ts
@@ -14,7 +14,7 @@ const modalUtils = new ModalUtils();
 const realmSettingsPage = new RealmSettingsPage();
 
 describe("Realm settings tests", () => {
-  describe.skip("Realm settings tabs tests", () => {
+  describe("Realm settings tabs tests", () => {
     const realmName = "Realm_" + (Math.random() + 1).toString(36).substring(7);
 
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


mfrances17 (approved):
LGTM, code looks good. Verified that the Enable/Disable switch saves appropriately and that it defaults to enabled.

cypress/integration/realm_settings_test.spec.ts
```diff
@@ -546,42 +546,42 @@ describe("Realm settings tests", () => {
...
     });
 
-    // it("Check navigating between Form View and JSON editor", () => {
-    //   realmSettingsPage.shouldNavigateBetweenFormAndJSONViewPolicies();
-    // });
+    it("Check navigating between Form View and JSON editor", () => {
+      realmSettingsPage.shouldNavigateBetweenFormAndJSONViewPolicies();
+    });
 
-    /*       it("Check saving changed JSON policies", () => {
```
mfrances17, line 581: Actually, you may want to keep this one commented out for now until it can be fixed - it is the only test that seems to be failing for realm settings.

mfrances17 (changes requested):
Sorry, I originally approved, but after checking out the cypress run, I would keep that one 'Check saving changed JSON policies' test commented out for now, until time can be spent on it to fix it. I think this is the only one that is keeping the realm settings suite from passing. After that, this can be merged.

jenny-s51:
@mfrances17 Thank you for your review! Yes you're right - just commented out the JSON policies test and will plan to fix it in a separate PR. 👍

mfrances17 (approved):

PR #1497, opened 2021-11-05
Added default role tab to realm settings

edewit:
Also changed that the default role in realm tab now navigates to the user registration tab off the realm settings.
And introduced a `KeycloakSpinner` to remove some duplicate code.

### DIFF
--- a/src/realm-roles/RealmRoleTabs.tsx
+++ b/src/realm-roles/RealmRoleTabs.tsx
@@ -5,7 +5,6 @@ import {
   ButtonVariant,
   DropdownItem,
   PageSection,
-  Spinner,
   Tab,
   TabTitleText,
 } from "@patternfly/react-core";
@@ -22,6 +21,7 @@ import {
   arrayToAttributes,
   AttributeForm,
 } from "../components/attribute-form/AttributeForm";
+import { KeycloakSpinner } from "../components/keycloak-spinner/KeycloakSpinner";
 import { ViewHeader } from "../components/view-header/ViewHeader";
 import { useConfirmDialog } from "../components/confirm-dialog/ConfirmDialog";
 import { RealmRoleForm } from "./RealmRoleForm";
@@ -320,8 +320,18 @@ export default function RealmRoleTabs() {
 
   const isDefaultRole = (name: string) => realm?.defaultRole!.name === name;
 
-  if (!realm || !role) {
-    return <Spinner />;
+  if (!realm) {
+    return <KeycloakSpinner />;
+  }
+  if (!role) {
+    return (
+      <RealmRoleForm
+        reset={() => reset(role)}
+        form={form}
+        save={save}
+        editMode={false}
+      />
+    );
   }
 
   return (
@@ -394,14 +404,6 @@ export default function RealmRoleTabs() {
               </Tab>
             )}
           </KeycloakTabs>
-        )}
-        {!id && (
-          <RealmRoleForm
-            reset={() => reset(role)}
-            form={form}
-            save={save}
-            editMode={false}
-          />
         )}
       </PageSection>
     </>
--- a/src/realm-roles/RealmRoleTabs.tsx
+++ b/src/realm-roles/RealmRoleTabs.tsx
@@ -318,8 +318,7 @@ export default function RealmRoleTabs() {
     }
   };
 
-  const isDefaultRole = (name: string) =>
-    (realm?.defaultRole! as unknown as RoleRepresentation).name === name;
+  const isDefaultRole = (name: string) => realm?.defaultRole!.name === name;
 
   if (!realm || !role) {
     return <Spinner />;
--- a/src/realm-roles/RealmRoleTabs.tsx
+++ b/src/realm-roles/RealmRoleTabs.tsx
@@ -205,7 +205,7 @@ export default function RealmRoleTabs() {
   });
 
   const dropdownItems =
-    url.includes("AssociatedRoles") && !realm?.defaultRole
+    url.includes("associated-roles") && !realm?.defaultRole
       ? [
           <DropdownItem
             key="delete-all-associated"
@@ -224,7 +224,7 @@ export default function RealmRoleTabs() {
             {t("deleteRole")}
           </DropdownItem>,
         ]
-      : id && realm?.defaultRole && url.includes("AssociatedRoles")
+      : id && realm?.defaultRole && url.includes("associated-roles")
       ? [
           <DropdownItem
             key="delete-all-associated"
@@ -292,11 +292,14 @@ export default function RealmRoleTabs() {
 
   const toAssociatedRoles = () => {
     const to = clientRoleRouteMatch
-      ? toClientRole({ ...clientRoleRouteMatch.params, tab: "AssociatedRoles" })
+      ? toClientRole({
+          ...clientRoleRouteMatch.params,
+          tab: "associated-roles",
+        })
       : toRealmRole({
           realm: realm?.realm!,
           id,
-          tab: "AssociatedRoles",
+          tab: "associated-roles",
         });
     history.push(to);
   };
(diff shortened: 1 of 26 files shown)


jenny-s51 (dismissed):
Great work @edewit ! Associated role inheritance works as expected both in `Realm roles -> associated roles` and `Realm settings -> User settings -> default role`.

I noticed the pagination isn't updating the data correctly, though after some investigating it seems like this can be resolved by simply removing the `isPaginated` prop in the KDT of the `AssociatedRolesTab`.

How would you suggest testing your changes in client-scopes? I was able to view the `AddRoleMappingModal`  by selecting a hardcoded role mapper type for an LDAP provider, but under client scopes I don't see any of these mapping types: `"service-account" | "client-scope" | "role"`.

src/realm-settings/UserRegistration.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useRealm } from "../context/realm-context/RealmContext";
+import { AssociatedRolesTab } from "../realm-roles/AssociatedRolesTab";
+import { KeycloakSpinner } from "../components/keycloak-spinner/KeycloakSpinner";
+
+type Realm = RealmRepresentation & {
+  defaultRole: {
+    id: string;
+    name: string;
+  };
+};
```
jonkoops, line 16: Why do we need this type? Is the type from the admin client incorrect? Can we fix this there as well?

src/realm-roles/routes/ClientRole.ts
```diff
@@ -7,7 +7,7 @@ export type ClientRoleTab =
   | "details"
   | "attributes"
   | "users-in-role"
-  | "AssociateRoles";
+  | "AssociatedRoles";
```
jonkoops, line 10: The casing of this key is different, perhaps we can align it with the rest?

src/realm-roles/RolesList.tsx
```diff
@@ -79,21 +81,25 @@ export const RolesList = ({
...
-  const RoleDetailLink = (role: RoleRepresentation) => (
-    <>
+  const RoleDetailLink = (role: RoleRepresentation) =>
+    role.name !== realm?.defaultRole?.name ? (
       <RoleLink role={role}>{role.name}</RoleLink>
-      {role.name?.includes("default-role") ? (
+    ) : (
+      <>
+        <Link
+          key={role.id}
```
jonkoops, line 90: I don't think we need a key here, unless I am missing some iteration.

src/realm-roles/RealmRoleTabs.tsx
```diff
@@ -284,14 +290,50 @@ export default function RealmRoleTabs() {
...
+      : toRealmRole({
+          realm: realm?.realm!,
+          id,
+          tab: "AssociatedRoles",
+        });
+    history.push(to);
+  };
+
+  const addComposites = async (composites: RoleRepresentation[]) => {
+    const compositeArray = composites;
```
jonkoops, line 311: This assignment seems pretty useless as it creates an alias which is only passed once.

jonkoops (approved):

src/realm-settings/UserRegistration.tsx
edewit: It's already fixed in the new adminClient

src/realm-roles/routes/ClientRole.ts
edewit: it's was already like this... I can change it

src/realm-roles/RolesList.tsx
edewit: right

src/realm-settings/RealmSettingsTabs.tsx
```diff
@@ -432,6 +433,14 @@ export const RealmSettingsTabs = ({
                 </Tab>
               </Tabs>
             </Tab>
+            <Tab
+              eventKey="userRegistration"
+              title={<TabTitleText>{t("userRegistration")}</TabTitleText>}
+              data-testid="rs-clientPolicies-tab"
+              aria-label={t("clientPoliciesTab")}
```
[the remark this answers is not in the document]
edewit: @jenny-s51 just removing the`isPaginated` can't be the solution right as it seems that `first` and `max` are used to find composite roles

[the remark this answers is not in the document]
edewit: > How would you suggest testing your changes in client-scopes? I was able to view the AddRoleMappingModal by selecting a hardcoded role mapper type for an LDAP provider, but under client scopes I don't see any of these mapping types: "service-account" | "client-scope" | "role".

I didn't change anything on client scopes only introduced this new screen

src/realm-settings/UserRegistration.tsx
edewit: can be removed when #1522 is merged

jenny-s51 (commented):
Pagination looks great and everything else works as expected @edewit! Using `isInheritedHidden` on the isPaginated prop was a good idea.

src/realm-roles/routes/ClientRole.ts
```diff
@@ -7,7 +7,7 @@ export type ClientRoleTab =
   | "details"
   | "attributes"
   | "users-in-role"
-  | "AssociateRoles";
+  | "associated-roles";
```
jenny-s51, line 10: Looks like the Cypress test just needs to be updated to reflect this new route

src/realm-settings/RealmSettingsTabs.tsx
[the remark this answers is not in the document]
jenny-s51: Ah I see, yes that makes sense. Thank you for clarifying

src/realm-settings/UserRegistration.tsx
jonkoops: @edewit This seems to be merged now, can we consider this resolved?

jonkoops (approved):

edewit:
@jenny-s51 I don't see the client tests fail?

src/realm-roles/routes/ClientRole.ts
edewit: I don't see the client tests fail?

src/realm-roles/RolesList.tsx
```diff
@@ -72,21 +70,24 @@ export const RolesList = ({
...
+        <Link
+          to={toRealmSettings({ realm: realmName, tab: "userRegistration" })}
+        >
+          {role.name}{" "}
+        </Link>
         <HelpItem
           helpText={t("defaultRole")}
           forLabel={t("defaultRole")}
           forID="kc-defaultRole"
           id="default-role-help-icon"
```
agagancarczyk, line 87: is this id needed? we don't seem to use it in some other components

src/realm-roles/RolesList.tsx
```diff
@@ -72,21 +70,24 @@ export const RolesList = ({
...
+      <>
+        <Link
+          to={toRealmSettings({ realm: realmName, tab: "userRegistration" })}
+        >
+          {role.name}{" "}
+        </Link>
         <HelpItem
           helpText={t("defaultRole")}
           forLabel={t("defaultRole")}
           forID="kc-defaultRole"
```
agagancarczyk, line 86: for accessibility, we use this format for forID={t(`common:helpLabel`, {
                    label: t("defaultRole"),
                  })}

src/realm-roles/RolesList.tsx
edewit: it's used in the css

src/realm-roles/routes/ClientRole.ts
jenny-s51: It looks like we can get the `realm_roles` test to pass by updating this line with `/associated-roles` https://github.com/keycloak/keycloak-admin-ui/blob/main/cypress/support/pages/admin_console/manage/realm_roles/AssociatedRolesPage.ts#L21

agagancarczyk (approved):

jonkoops (approved):

Issue #1444, opened 2021-10-28
isDisabled in <FormAccess> not disabling all fields 


agagancarczyk:
Using isDisabled in FormAccess is not disabling all fields. Once fixed we will need to address it in global executors as per https://marvelapp.com/prototype/h11728e/screen/81618846

ssilvert:
This may be related to https://github.com/keycloak/keycloak-admin-ui/issues/2099

edewit:
This is solved now right?

agagancarczyk:
yes, fixed, thanks @edewit

PR #1499, opened 2021-11-05
Executor details

agagancarczyk:
## Motivation
-> Displaying default values for executors
-> Editing configuration for added executors
-> Breadcrum fix for edit client so that user can navigate to the list of client profiles via a breadcrum
-> Cypress tests

Please note that the following are not implemented due to bugs in the components: 
1. Disabling configs for executors in global profiles as per https://marvelapp.com/prototype/h11728e/screen/81618846 -> New issue https://github.com/keycloak/keycloak-admin-ui/issues/1444
2. Editing profile executors when the executor is a "List" or "MultivaluedList" -> New issue  https://github.com/keycloak/keycloak-admin-ui/issues/1498

### Profiles
1. Go to  /realm-settings/clientPolicies, create a new profile and create a new executor for that profile (only executors with booleans and strings will be editable).
2. Select created executor, verify that the value is correct and edit it to something else.

### Global profiles
1. Select global executor and verify that default configuration values display.
2. These configuration values should be disbaled but this is not implemented yet due to https://github.com/keycloak/keycloak-admin-ui/issues/1444

## Checklist:


src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
...
-        executors,
-      };
+      let editedExecutorConfig = {};
+      if (editMode) {
+        editedExecutorConfig = Object.assign(
+          profileExecutor!.configuration,
+          formValues.config
+        );
+        profileExecutor!.configuration = editedExecutorConfig;
+      }
```
jonkoops, line 115: Since `editedExecutorConfig` is only used in the context of this statement I believe you can shorten this a bit.
```suggestion
      if (editMode) {
        profileExecutor!.configuration = {
          ...profileExecutor!.configuration,
          ...formValues.config,
        };
      }
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
...
+
+      if (editMode) {
+        return {
+          ...profile,
+        };
```
jonkoops, line 120: Is there a need to spread this value? Perhaps it can be returned immediately?
```suggestion
        return profile;
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
...
+      } else {
+        return {
+          ...profile,
+          executors,
+        };
+      }
```
jonkoops, line 126: Since the other branch of this statement returns a value any code that comes after is already like an 'else' clause implicitly. This means you can omit the `else` statement.
```suggestion
      return {
        ...profile,
        executors,
      };
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
...
+    (property) => {
+      const globalDefaultValues = editMode ? property.defaultValue : "";
+
+      return {
+        helpText: property.helpText!,
+        label: property.label!,
+        name: property.name!,
+        defaultValue: globalDefaultValues,
+        type: property.type!,
+      };
```
jonkoops, line 169: If you are copying the properties over perhaps you can use a spread here instead?
```suggestion
      return {
        ...property,
        defaultValue: globalDefaultValues,
      };
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
...
 
+  const globalProfile = globalProfiles.find(
+    (globalProfile) => globalProfile.name === profileName
+  );
+
+  const profileExecutorType = executorTypes?.find(
+    (executor) => executor.id === executorName
+  );
+
+  const editedProfileExecutors = profileExecutorType?.properties.map(
```
jonkoops, line 159: It's not required, but it might be a good idea to give this map operation a type so the returned value is type-checked.
```suggestion
  const editedProfileExecutors = profileExecutorType?.properties.map<ConfigPropertyRepresentation>(
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -117,12 +190,20 @@ export default function ExecutorForm() {
...
+                    label: t("executorTypeHelpText"),
+                  })}
+                />
               ) : undefined
             }
           >
             <Controller
               name="executor"
-              defaultValue=""
+              defaultValue={""}
```
jonkoops, line 206: There is no need for curly braces around props with a literal string value.
```suggestion
              defaultValue=""
```

src/realm-settings/ExecutorForm.tsx
```diff
@@ -178,26 +260,67 @@ export default function ExecutorForm() {
...
+                    {...props}
+                    to={toClientProfile({ realm, profileName })}
+                  />
+                )}
+                data-testid="addExecutor-cancelBtn"
+              >
+                {t("common:cancel")}
+              </Button>
+            </ActionGroup>
+          ) : undefined}
```
jonkoops, line 305: There is no need to use a ternary operation here.
```suggestion
          {!globalProfile && (
            <ActionGroup>
              <Button
                variant="primary"
                onClick={() => handleSubmit(save)()}
                data-testid="addExecutor-saveBtn"
              >
                {editMode ? t("common:save") : t("common:add")}
              </Button>
              <Button
                variant="link"
                component={(props) => (
                  <Link
                    {...props}
                    to={toClientProfile({ realm, profileName })}
                  />
                )}
                data-testid="addExecutor-cancelBtn"
              >
                {t("common:cancel")}
              </Button>
            </ActionGroup>
          )}
```

jonkoops (changes requested):

edewit (approved):

cypress/integration/realm_settings_test.spec.ts
```diff
@@ -537,7 +545,7 @@ describe("Realm settings tests", () => {
     });
   });
 
-  describe("Realm settings client policies tab tests", () => {
+  describe.skip("Realm settings client policies tab tests", () => {
```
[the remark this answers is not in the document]
agagancarczyk: Skipping these because they are failing. I'll let Jenny know that they are failing so that she can take another look at these. Thanks!

src/realm-settings/ExecutorForm.tsx
```diff
@@ -26,19 +26,23 @@ import {
   COMPONENTS,
   isValidComponentType,
 } from "../client-scopes/add/components/components";
-import type ClientPolicyExecutorRepresentation from "@keycloak/keycloak-admin-client/lib/defs/clientPolicyExecutorRepresentation";
+import type { ExecutorParams } from "./routes/Executor";
 
-type ExecutorForm = Required<ClientPolicyExecutorRepresentation>;
+type ExecutorForm = {
```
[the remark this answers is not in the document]
agagancarczyk: ClientPolicyExecutorRepresentation has configuration and I need config.

src/realm-settings/ExecutorForm.tsx
```diff
@@ -55,54 +59,123 @@ export default function ExecutorForm() {
     ClientProfileRepresentation[]
   >([]);
   const [profiles, setProfiles] = useState<ClientProfileRepresentation[]>([]);
-  const form = useForm<ExecutorForm>({ defaultValues });
-  const { control } = form;
+  const form = useForm({ defaultValues });
+  const { control, setValue, handleSubmit } = form;
+  const editMode = executorName ? true : false;
```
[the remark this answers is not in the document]
agagancarczyk: like this! thanks @jonkoops

src/realm-settings/ExecutorForm.tsx
agagancarczyk: yeah, that's true! thanks!

src/realm-settings/ExecutorForm.tsx
agagancarczyk: ah, yeah, thanks!

src/realm-settings/ExecutorForm.tsx
agagancarczyk: true!

src/realm-settings/ExecutorForm.tsx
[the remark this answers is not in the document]
jonkoops: Probably it was hard-coded in other places

jonkoops (approved):
LGTM

PR #1509, opened 2021-11-08
Add user profile tab to realm settings

jonkoops:
## Motivation
Adds the user profile tab to the realm settings and implements the JSON editor as [per design](https://marvelapp.com/prototype/5eed532/screen/82246815).


xianli123:
@jonkoops . Thanks. That looks good to me. One small detail: can we reduce the vertical size of the box so that users can see the **Save** and **Revert** buttons at their first glance instead of scrolling down? WDYT?

yih-wang:
Hi @jonkoops, it looks great! 

The only question I have is when I go to the tab and single-click somewhere in the code, it will default select all the codes, but not let me start editing. Is it the default behavior of this component? Can we change that?

Another question is related to the code part. Users are not allowed to edit the global profiles. I don't know if there's some way to indicate that. Is it possible and intuitive if we disable the global profiles code (code in grey color)? Or just don't allow users to input/edit the related part and provide some inline notifications? Can you give some opinions from your experience when using the code editor?

yih-wang:
> @jonkoops . Thanks. That looks good to me. One small detail: can we reduce the vertical size of the box so that users can see the **Save** and **Revert** buttons at their first glance instead of scrolling down? WDYT?

@xianli123 I can see them on the same page. Maybe it's a resolution issue？

xianli123:
> @xianli123 I can see them on the same page. Maybe it's a resolution issue？

Maybe my browser has a limited visible size. It doesn't matter. Ignore it.

jonkoops:
> The only question I have is when I go to the tab and single-click somewhere in the code, it will default select all the codes, but not let me start editing. Is it the default behavior of this component? Can we change that?

Great catch! This turned out to be the default behavior because of the way my code was written. This also broke other things you'd expect to work out of the box (such as pressing Ctrl+Z to revert a change). I've changed the code so this now works as expected.

I did a little checking and it looks like the global profiles will not be modified when saving them with edits, it simply reverts back to what was there before whilst keeping the edits made to the regular profiles. Although not ideal from a UX perspective it at least looks like the security aspect is taken care of by the API that Keycloak provides.

> Is it possible and intuitive if we disable the global profiles code (code in grey color)?

I looked around and found this related issue https://github.com/microsoft/monaco-editor/issues/953 for the editor we are using. It is not supported directly, but [someone in this thread](https://github.com/microsoft/monaco-editor/issues/953#issuecomment-958736258) created [a plugin](https://www.npmjs.com/package/constrained-editor-plugin) that could allow us to do this. 

> Or just don't allow users to input/edit the related part and provide some inline notifications? Can you give some opinions from your experience when using the code editor?

I think we can do an experiment with the aforementioned plugin to prevent edits to the 'restricted' parts of the configuration and add a message to make them understand why it cannot be edited. I would propose that we do this in a separate issue to reduce the churn on this PR.

WDYT @yih-wang?

mfrances17 (changes requested):
Code looks good, the JSON editor works great, nice job!

Only comment I would have is that you should check for dirty before enabling the Save button - right now it is always enabled. To be consistent, you shouldn't be able to save if nothing changed.

jonkoops:
@xianli123 @yih-wang What do you think about @mfrances17's comment? Is this in line with the UXD expectancy? And if so should we do the same for the 'reset' button if there is nothing to reset?

mfrances17:
it is only for primary save buttons. we do it throughout the admin console UI. also, you have a merge conflict. 👍

mfrances17 (approved):
Save button behavior is being handled as a separate issue here: https://github.com/keycloak/keycloak-admin-ui/issues/1541

Approving and merging.

yih-wang:
@jonkoops The plugin looks amazing. Yes open a new issue makes sense to me. Thanks!

jonkoops:
@yih-wang Awesome, I've created #1545 to implement this.

Issue #1479, opened 2021-11-03
User Fed: group-ldap-mapper has wrong default


ssilvert:
1. Go to User Federation --> LDAP --> LDAP Mappers
2. Create new mapper
3. Choose `group-ldap-mapper` as mapper type

Default value for `Membership user LDAP attribute` should be `uid` instead of `cn`

mfrances17:
@ssilvert this is not true in the old admin console:
![group-ldap-mapper](https://user-images.githubusercontent.com/39063664/140359173-52c86968-f9a4-4735-a614-354074afcc2a.png)

ssilvert:
@mfrances17 I found the discrepancy.  The defaults are different based on whether you are using Active Directory or Red Hat Directory Server.  The default for `Group Object Classes` is also different.  Maybe others as well.

mfrances17:
ahhhh, ok - thank you! i'll fix it and test all the possible scenarios.

PR #1511, opened 2021-11-08
Fix default values for group and role mappers

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1479.

## Brief Description
For both the group-ldap-mapper and role-ldap-mapper user fed LDAP mappers, if the vendor of the LDAP provider is Active Directory, then there should be different default values for Group/Role Object Classes and Membership User LDAP Attribute fields. The other four vendors share the same default values for those two fields. See https://github.com/keycloak/keycloak-admin-ui/issues/1479 for more info.

## Verification Steps
1. Create a User Federation LDAP provider and set the vendor to Active Directory.
2. Go to User Federation > LDAP > LDAP Mappers.
3. Create a new mapper.
4. Choose **group-ldap-mapper** as the mapper type.
5. Verify the following default values: 
    - Group Object Classes: `group`
    - Membership User LDAP Attribute: `cn`
6. Using the same LDAP provider, create a new mapper and choose **role-ldap-mapper** as the mapper type.
7. Verify the following default values: 
    - Role Object Classes: `group`
    - Membership User LDAP Attribute: `cn`
8. Create a new User Federation LDAP provider and set the vendor to any vendor other than Active Directory.
9. Go to User Federation > LDAP > LDAP Mappers.
10. Create a new mapper.
11. Choose **group-ldap-mapper** as the mapper type.
12. Verify the following default values: 
    - Group Object Classes: `groupOfNames`
    - Membership User LDAP Attribute: `uid`
13. Using the same LDAP provider, create a new mapper and choose **role-ldap-mapper** as the mapper type.
14. Verify the following default values: 
    - Role Object Classes: `groupOfNames`
    - Membership User LDAP Attribute: `uid`

## Checklist:

## Additional Notes
None


src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -52,6 +56,32 @@ export const LdapMapperRoleGroup = ({
...
+          );
+          form.setValue("config.membership-user-ldap-attribute[0]", "cn");
+        }
+      } else if (id) {
+        throw new Error(t("common:notFound"));
+      }
+    },
+    []
+  );
+
```
edewit, line 84: This component is only used in `LdapMapperDetails` and there you already query the component wouldn't it be better just to pass in the vendor instead of querying again?

edewit (changes requested):

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -33,12 +34,15 @@ export const LdapMapperRoleGroup = ({
...
   const [clients, setClients] = useState<ClientRepresentation[]>([]);
 
+  let vendor: string | undefined;
   let isRole = true;
   const groupMapper = "group-ldap-mapper";
 
   if (type === groupMapper) {
     isRole = false;
   }
+  const { id } = useParams<{ id: string }>();
```
jonkoops, line 44: Can we use the appropriate type from the respective route here?

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -33,12 +34,15 @@ export const LdapMapperRoleGroup = ({
   const [isClientIdDropdownOpen, setIsClientIdDropdownOpen] = useState(false);
   const [clients, setClients] = useState<ClientRepresentation[]>([]);
 
+  let vendor: string | undefined;
```
jonkoops, line 37: There is no need to define this in the outer scope as this is only used in a single function.
```suggestion
```

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -33,12 +34,15 @@ export const LdapMapperRoleGroup = ({
...
   let isRole = true;
   const groupMapper = "group-ldap-mapper";
 
   if (type === groupMapper) {
     isRole = false;
   }
+  const { id } = useParams<{ id: string }>();
+  const { mapperId } = useParams<{ mapperId: string }>();
```
jonkoops, line 45: I think you can combine these two calls to `useParams` into a single one and destructure both parameters.

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -52,6 +56,32 @@ export const LdapMapperRoleGroup = ({
     []
   );
 
+  useFetch(
+    async () => {
+      if (id) {
+        return await adminClient.components.findOne({ id });
+      }
+      return undefined;
+    },
```
jonkoops, line 65: According to the route type the `id` field is always defined, which means this code can be simplified.
```suggestion
    () => adminClient.components.findOne({ id }),
```

jonkoops (changes requested):

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -52,6 +55,29 @@ export const LdapMapperRoleGroup = ({
     []
   );
 
+  useFetch(
+    async () => {
```
jonkoops, line 59: This can be shortened even further as an async function is a wrapper for a Promise, so in this case the `async` and `await` keywords can be omitted. Also the arrow function return implicitly.

```ts
useFetch(() => adminClient.components.findOne({ id }))
```

jonkoops (approved):

jenny-s51 (approved):
Looks and works great @mfrances17 - it would be nice to clean up the `useFetch` to return implicitly as @jonkoops suggested. Otherwise 👍!

edewit (approved):

PR #1522, opened 2021-11-09
update the admin client to the latest version

edewit:


mfrances17 (changes requested):
Was the removal of the additions to the RolesRepresentation type intentional? Seems unrelated to updating the admin client to the latest version - but regardless, 'name' doesn't exist without adding to the type but is referenced in a couple of places in the component.

edewit:
@mfrances17 the defined types where a workaround because they where not right in the admin client. Now that they are right in the admin client we can / must remove them here

mfrances17:
ah, ok thanks @edewit... maybe i didn't pull in the latest admin client successfully when i tested? i'll retest.

mfrances17 (approved):
LGTM, not getting any errors - suspect i did not have had the latest admin client when originally tested.

PR #1523, opened 2021-11-10
Split realm context into two

jonkoops:
## Motivation
Separates the concerns of the active realm and the realm overview into two separate contexts. This is also a prerequisite for some further improvements I want to make to the active realm context to reduce the amount of API calls.

This also improves memoization of the values the context provides, which will reduce [accidental re-renders](https://medium.com/@ryanflorence/react-context-and-re-renders-react-take-the-wheel-cd1d20663647).

### DIFF
--- a/src/context/RealmsContext.tsx
+++ b/src/context/RealmsContext.tsx
@@ -1,0 +1,58 @@
+import type RealmRepresentation from "@keycloak/keycloak-admin-client/lib/defs/realmRepresentation";
+import { sortBy } from "lodash";
+import React, {
+  createContext,
+  FunctionComponent,
+  useCallback,
+  useMemo,
+  useState,
+} from "react";
+import { RecentUsed } from "../components/realm-selector/recent-used";
+import useRequiredContext from "../utils/useRequiredContext";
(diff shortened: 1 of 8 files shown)


src/context/RealmsContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const [realms, setRealms] = useState<RealmRepresentation[]>([]);
+  const recentUsed = useMemo(() => new RecentUsed(), []);
+
+  function updateRealms(realms: RealmRepresentation[]) {
+    setRealms(sortBy(realms, "realm"));
+    recentUsed.clean(realms.map(({ realm }) => realm!));
+  }
+
+  useFetch(
+    () => adminClient.realms.find(),
```
edewit, line 36: we should add the new `briefRepresentation=true` to this call so that we only get the realm names and make this call faster.

edewit (commented):
we need to add that to the client though

src/context/RealmsContext.tsx
jonkoops: Perhaps we can do this in a separate PR? I don't want to make this too massive.

edewit (approved):

PR #1530, opened 2021-11-10
Prevent Dependabot from offering major updates for React Router breadcrumbs

jonkoops:
Prevents major updates to `use-react-router-breadcrumbs` from being offered since it is not compatible with the version of React Router we are using.


mfrances17 (approved):
LGTM

PR #1532, opened 2021-11-10
Realm settings(Client policies -> conditions): Update client-roles condition to match new design

jenny-s51:
## Motivation

https://marvelapp.com/prototype/2b3ghc06/screen/83236333

## Brief Description

Created a new `MultivaluedRoleComponent` to support selection of multiple client roles as per the updated design (link above). 

Client roles are now listed in a typeaheadMulti select variant. 

The work done in this PR also allows for the `client-roles` condition type to be saved and edited within an existing client policy. 

## Verification Steps

1. Go to Realm settings(Client policies -> profiles -> conditions).
2. Create a new client policy. 
3. Create a condition within the client policy.
4. Select the `client-roles` condition type.
5. Verify that the new multi-select component renders correctly.
6. Select from any of the client roles listed and verify that the data is saved correctly.
7. Click on the `client-roles` condition under "Conditions"
8. Edit the condition by selecting/deselecting client-roles from the dropdown.
9. Go back to the edited condition and verify it was saved successfully.  

## Checklist:

## Additional Notes

1. With the excellent work done in @agagancarczyk 's PRs for client policies, there have already been a couple times that this path has been changed and merged, which breaks client policies. https://github.com/keycloak/keycloak-admin-ui/pull/1499/files#diff-e9bd4be211c473f1940cf2eb43b009ca1a216d07632c521f494d6a2c4dba6171L12

   **It is important that we keep `/edit-profiles` in this path here, otherwise this breaks the client policy form.**


jonkoops:
@jenny-s51 there are some merge conflicts, could you rebase your branch?

src/client-scopes/add/components/MultivaluedRoleComponent.tsx
```diff
@@ -0,0 +1,120 @@
...
+        })
+      );
+
+      return alphabetize(clientRoles.flat());
+    },
+    (clientRoles) => {
+      setClientRoles(clientRoles);
+    },
+    []
+  );
```
ssilvert, line 58: I'm afraid this is going to be a performance problem if there are a large number of clients.  And the only way to fix it will be to add a new endpoint on the server (unless one of the currently planned endpoints will eventually work.  We should probably talk this over with @edewit and @yih-wang.  

For now, the safe thing is to just do it like the old UI, which allows zero or more ordinary text fields.

ssilvert (changes requested):

ssilvert:
@jenny-s51 Also, there are a lot of failed tests.

edewit (changes requested):

src/realm-settings/NewClientPolicyCondition.tsx
```diff
@@ -316,19 +325,34 @@ export default function NewClientPolicyCondition() {
...
+            <FormProvider {...form}>
+              <DynamicComponents
+                properties={conditionProperties}
+                selectedValues={
+                  conditionName === "client-access-type"
+                    ? selectedVals
+                    : conditionName === "client-updater-context"
+                    ? selectedVals?.["update-client-source"]
+                    : []
+                }
```
edewit, line 351: what is this for it doesn't seem to be used at all

agagancarczyk:
> @jenny-s51 Also, there are a lot of failed tests.

Just a few observations: 

- Any tests relying on creating a new realm role will fail until https://github.com/keycloak/keycloak-admin-ui/pull/1497#pullrequestreview-807057020 gets merged in.

- Events search tests were fixed last week, I hope that simple re-run will show them green. If not, I can take another look.

src/client-scopes/add/components/MultivaluedRoleComponent.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const { control } = useFormContext();
+
+  const [clientRoles, setClientRoles] = useState<RoleRepresentation[]>([]);
+
+  const [open, setOpen] = useState(false);
+
+  const fieldName = `config.${convertToHyphens(name!)}`;
+
+  const alphabetize = (rolesList: RoleRepresentation[]) => {
+    return sortBy(rolesList, (role) => role.name?.toUpperCase());
```
jonkoops, line 36: If you want to compare a string and sort it correctly take a look at [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

src/realm-settings/NewClientPolicyCondition.tsx
```diff
@@ -316,19 +325,34 @@ export default function NewClientPolicyCondition() {
...
-                  : []
-              }
-              parentCallback={handleCallback}
-            />
-          </FormProvider>
+
+          {conditionType === "client-roles" ||
+          conditionName === "client-roles" ? (
+            <FormProvider {...form}>
+              <MultivaluedRoleComponent
```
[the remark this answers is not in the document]
jenny-s51: The dynamic components render based on the `type` attribute of the condition in `serverInfo` ... in this case it's `MultivaluedString` which it would render the incorrect dynamic component. This is why I added this conditional here  and implemented the component directly

src/realm-settings/NewClientPolicyCondition.tsx
jenny-s51: This sets the selected values for the condition types which use the `MultivaluedListComponent`

src/realm-settings/NewClientPolicyCondition.tsx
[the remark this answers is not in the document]
edewit: Still it would be good to put this `MultivaluedRoleComponent` in the same folder as the other Dynamic components: `./src/components/dynamic`

src/realm-settings/NewClientPolicyCondition.tsx
```diff
@@ -316,19 +325,33 @@ export default function NewClientPolicyCondition() {
...
+                name={conditionName || conditionProperties[0].name}
+                key={conditionName || conditionProperties[0].name}
+                label={t("clientRoles")}
+                helpText={t("realm-settings-help:clientRolesConditionTooltip")}
+                selectedValues={selectedVals}
```
edewit, line 337: selectedValues are passed in here but `MultivaluedRoleComponent` doesn't use them

src/realm-settings/NewClientPolicyCondition.tsx
edewit: it seems these selectedValues are always an empty array

edewit:
When I just create a couple of conditions and remove and edit after a while I hit a server error:
```
10:39:36,362 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (default task-16) Uncaught server error: java.lang.IllegalArgumentException: Cannot deserialize value of type `java.util.ArrayList<java.lang.Object>` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.keycloak.services.clientpolicy.condition.ClientUpdaterContextCondition$Configuration["update-client-source"]->java.util.ArrayList[0])
	at com.fasterxml.jackson.core.jackson-databind@2.12.1//com.fasterxml.jackson.databind.ObjectMapper._convert(ObjectMapper.java:4314)
	at com.fasterxml.jackson.core.jackson-databind@2.12.1//com.fasterxml.jackson.databind.ObjectMapper.convertValue(ObjectMapper.java:4245)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.services.clientpolicy.ClientPoliciesUtil.getConditionProvider(ClientPoliciesUtil.java:360)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.services.clientpolicy.ClientPoliciesUtil.getEnabledClientPolicies(ClientPoliciesUtil.java:336)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.services.clientpolicy.DefaultClientPolicyManager.doPolicyOperation(DefaultClientPolicyManager.java:69)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.services.clientpolicy.DefaultClientPolicyManager.triggerOnEvent(DefaultClientPolicyManager.java:61)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.protocol.oidc.endpoints.AuthorizationEndpoint.process(AuthorizationEndpoint.java:167)
	at org.keycloak.keycloak-services@16.0.0-SNAPSHOT//org.keycloak.protocol.oidc.endpoints.AuthorizationEndpoint.buildGet(AuthorizationEndpoint.java:106)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.jboss.resteasy.resteasy-jaxrs@3.15.1.Final//org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:138)
	at org.jboss.resteasy.resteasy-jaxrs@3.15.1.Final//org.jboss.resteasy.core.ResourceMethodInvoker.internalInvokeOnTarget(ResourceMethodInvoker.java:546)
	at org.jboss.resteasy.resteasy-jaxrs@3.15.1.Final//org.jboss.resteasy.core.ResourceMethodInvoker.invokeOnTargetAfterFilter(ResourceMethodInvoker.java:435)
	at org.jboss.resteasy.resteasy-jaxrs@3.15.1.Final//org.jboss.resteasy.core.ResourceMethodInvok
```
only way to recover is to restart with new database

src/realm-settings/NewClientPolicyCondition.tsx
```diff
@@ -163,12 +172,12 @@ export default function NewClientPolicyCondition() {
...
-          [Object.keys(configValues)[0]]: Object.values(
-            configValues?.[Object.keys(configValues)[0]]
-          ).map((item) => (item as ItemType).value),
-        };
+      }
+      return {
+        [Object.keys(configValues)[0]]: Object.values(
+          configValues?.[Object.keys(configValues)[0]]
+        ).map((item) => (item as ItemType).value),
+      };
```
edewit, line 180: I think it would be nicer to use the configuration -> type to determinate if field needs conversion instead of creating this "mapping" between condition name and value. The idea behind this from keycloak point of view is that this is dynamic, but this mapping is now static

src/client-scopes/add/components/MultivaluedRoleComponent.tsx
jenny-s51: I've updated the code to use a case-insensitive `orderBy` from lodash! But I think it was doing a case-insensitive sorting before with `sortBy` ... What is the benefit of using `localeCompare` over either of these lodash functions here?

edewit (approved):

src/client-scopes/add/components/MultivaluedRoleComponent.tsx
jonkoops: The advantage is that it will take into account the user's locale. For example, in French you might have accents on vowels, if you sort purely on the numerical representation of a string it will end up at the bottom (instead of logically after the un-accentuated vowel).

Basically it will just match the expectations of the user more, rather than whatever numerical value UTF-8 assigns to the character.

jenny-s51: @jonkoops Thank you for clarifying and for suggesting this! That makes sense - using `localeCompare` is a great idea. Updated 👍

jonkoops (approved):

jonkoops:
@ssilvert Can you leave a new review so we can merge this PR?

ssilvert (approved):
Sorry.  Didn't know this was waiting on me.  @edewit says our back end plans are in line with fixing my concerns later.  So I'm approving.

PR #1533, opened 2021-11-10
Event listeners

agagancarczyk:
## Motivation
- Adding event listeners sub-tab as per https://marvelapp.com/prototype/6ac0aja/screen/83306311
- Adding cypress tests

## Verification Steps
1.  Go to /realm-settings/events tab and verify that the form is as per design. 
2. Select an option from the dropdown and select "Revert". Revert should revert to the original/ default selection if any.
3. Select an option from the dropdown and select "Save". Success alert should display. Verify that saved events are now visible.

## Checklist:


cypress/integration/realm_settings_test.spec.ts
```diff
@@ -118,7 +118,7 @@ describe("Realm settings tests", () => {
       masthead.checkNotificationMessage("Realm successfully updated");
     });
 
-    /* 
+    /*
```
jonkoops, line 121: Instead if leaving this test commented out perhaps we can use `.skip()` instead so it show up in the overview?

jonkoops (approved):

cypress/integration/realm_settings_test.spec.ts
agagancarczyk: sounds good, using .skip instead now - thanks @jonkoops

PR #1537, opened 2021-11-11
Add correct label for Dependabot PRs to config

jonkoops:
## Motivation
Makes sure that Dependabot uses our new labels correctly.


agagancarczyk (approved):
lgtm

PR #1554, opened 2021-11-12
Save Pass Subject correctly and add two missing fields to SAML IDP

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1487.

## Brief Description
Change the `Pass subject` field so it correctly points to the correct API value: `loginHint`.
Added the two missing fields: `Attribute Consuming Service Index` and `Attribute Consuming Service Name`.

## Verification Steps
1. Go to Identity providers > some SAML provider > Provider details.
2. Change the `Pass subject` switch to `ON` (or if it is already `ON`, change it to `OFF`).
3. Add values to the `Attribute Consuming Service Index` and `Attribute Consuming Service Name` fields, and verify their help works as expected.
4. Click `Save`.
5. Verify that the values of the three fields you modified are persisted in both the new and old admin consoles.
6. Change the three values in the old admin console and verify that they are persisted in the new admin console.

## Checklist:

## Additional Notes
None

### DIFF
--- a/src/identity-providers/add/DescriptorSettings.tsx
+++ b/src/identity-providers/add/DescriptorSettings.tsx
@@ -414,7 +414,7 @@ const Fields = ({ readOnly }: DescriptorSettingsProps) => {
         isReadOnly={readOnly}
       />
       <SwitchField
-        field="config.passSubject"
+        field="config.loginHint"
         label="passSubject"
         isReadOnly={readOnly}
       />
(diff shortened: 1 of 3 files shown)


jonkoops (approved):

jenny-s51 (approved):
LGTM! @mfrances17 👍

PR #1559, opened 2021-11-15
filter out unknown mappers

edewit:
fixes: #1478

### DIFF
--- a/src/client-scopes/details/MapperList.tsx
+++ b/src/client-scopes/details/MapperList.tsx
@@ -56,22 +56,30 @@ export const MapperList = ({
     setAddMapperDialogOpen(!addMapperDialogOpen);
   };
 
-  const loader = async () =>
-    Promise.resolve(
-      (mapperList || [])
-        .map((mapper) => {
-          const mapperType = mapperTypes.filter(
-            (type) => type.id === mapper.protocolMapper
-          )[0];
-          return {
-            ...mapper,
-            category: mapperType.category,
-            type: mapperType.name,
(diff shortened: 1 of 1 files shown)


src/client-scopes/details/MapperList.tsx
```diff
@@ -56,22 +56,25 @@ export const MapperList = ({
...
           } as Row;
-        })
-        .sort((a, b) => a.priority - b.priority)
-    );
+        }
+      })
+      .filter((e) => e !== undefined) as Row[];
+
+    return Promise.resolve(list.sort((a, b) => a.priority - b.priority));
+  };
```
jonkoops, line 77: I think this function could be improved a bit, what do you think of this?

```ts
const loader = async () => {
  if (!mapperList) {
    return [];
  }

  const list = mapperList.reduce<Row[]>((rows, mapper) => {
    const mapperType = mapperTypes.find(
      ({ id }) => id === mapper.protocolMapper
    );

    if (!mapperType) {
      return rows;
    }

    return rows.concat({
      ...mapper,
      category: mapperType.category,
      type: mapperType.name,
      priority: mapperType.priority,
    });
  }, []);

  return list.sort((a, b) => a.priority - b.priority);
};
```

jonkoops (changes requested):

src/client-scopes/details/MapperList.tsx
edewit: yep that improves it...

jonkoops (approved):

PR #1560, opened 2021-11-15
Update to PF 4.159.1

jenny-s51:
## Motivation
Update Patternfly to the latest releases so we stay current with the latest changes.

## Brief Description
As part of the Patternfly release process, I perform sanity tests of the new proposed versions on the keycloak admin UI and file issues as needed. This occurs about every 3 weeks. As part of that process, we should update Keycloak to the latest versions on a regular cadence.

## Verification Steps
Use the keycloak admin UI and verify that no visual or behavioral issues have been introduced, and that there are no new console warnings/errors.

## Checklist:

## Additional Notes
In addition to manual sanity testing, the full cypress test suite was run and no new failures were introduced as a result of the PF upgrade.


package.json
```diff
@@ -24,11 +24,11 @@
...
-    "@patternfly/patternfly": "^4.151.4",
-    "@patternfly/react-code-editor": "^4.9.8",
-    "@patternfly/react-core": "^4.168.8",
-    "@patternfly/react-icons": "4.19.8",
-    "@patternfly/react-table": "^4.37.8",
+    "@patternfly/patternfly": "4.159.1",
+    "@patternfly/react-code-editor": "4.16.2",
+    "@patternfly/react-core": "4.175.2",
+    "@patternfly/react-icons": "4.26.2",
+    "@patternfly/react-table": "4.44.2",
```
jonkoops, line 31: Can you use the version ranges like before instead of exact versions?

jonkoops (changes requested):

PR #1561, opened 2021-11-15
Show/hide principal attribute field correctly

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1433.

## Brief Description
Changed watch behavior of the `Principal type` field so that it does not set `Principal attribute` incorrectly.

## Verification Steps
1. Create a new SAML ID provider.
2. Set Principal type to `Attribute [Name]` or `Attribute [Friendly Name]`, fill in a value for `Principal attribute`, and save.
3. Go back to the ID Provider list.
4. Return to the details screen of the SAML ID provider created in step 1.
5. Verify that the `Principal attribute` field is displayed with the value you added in step 2.

## Checklist:

## Additional Notes
None

### DIFF
--- a/src/identity-providers/add/DescriptorSettings.tsx
+++ b/src/identity-providers/add/DescriptorSettings.tsx
@@ -247,7 +247,7 @@ const Fields = ({ readOnly }: DescriptorSettingsProps) => {
         ></Controller>
       </FormGroup>
 
-      {principalType!.includes("ATTRIBUTE") && (
+      {principalType?.includes("ATTRIBUTE") && (
         <FormGroup
           label={t("principalAttribute")}
           labelIcon={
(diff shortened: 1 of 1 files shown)


src/identity-providers/add/DescriptorSettings.tsx
```diff
@@ -248,7 +247,7 @@ const Fields = ({ readOnly }: DescriptorSettingsProps) => {
         ></Controller>
       </FormGroup>
 
-      {principalType.includes("ATTRIBUTE") && (
+      {principalType!.includes("ATTRIBUTE") && (
```
jonkoops, line 250: Using a non-null assertion operator here causes the page to break if you add a new identity provider. Using optional chaining seems to fix the issue.
```suggestion
      {principalType?.includes("ATTRIBUTE") && (
```

jonkoops (changes requested):

src/identity-providers/add/DescriptorSettings.tsx
mfrances17: thanks... fixed

jonkoops (approved):

PR #1570, opened 2021-11-18
Alternative solution for test fix

edewit:
This is an alternative solution to fix the IdP test (#1569), both commits fix this test.

The first commit fixes the test itself by waiting on the request so that the component doesn't get unmounted when it's still loading.
The second is resetting the cancalation token so that adminClient requests no longer use the same token.


jonkoops:
Closing this as we cherry picked work from this branch to #1569

PR #1574, opened 2021-11-18
Downgrade to Node 16

jonkoops:
## Motivation
Node 17 causes issues like failing to run the tests in Cypress 9 and breaks `import.mjs`. To mitigate this we can downgrade to version 16 which is an LTS version and will be supported for a while.


jenny-s51 (approved):
Node 16 looks good @jonkoops !

mfrances17 (approved):
LGTM - all cypress tests passed on CI so this should be good to go

PR #1577, opened 2021-11-19
Realm settings(client policies): add/edit client-updater-source-groups condition

jenny-s51:
## Motivation

https://issues.redhat.com/browse/CLUX-331
https://marvelapp.com/prototype/2b3ghc06/screen/81619359

## Brief Description

- Adds screen, correct fields, and functionality to client policies -> add condition -> client-updater-source-groups condition type
- Supports save and edit functionality

## Verification Steps

1. Go to Realm settings -> Client policies -> Policies subtab
2. Create a new client policy if one doesn't exist yet.
3. Click on the client policy and under Conditions, click "Add condition"
4. Select the "client-updater-source-groups" condition type.
5. Verify that the UI and functionality match the design and work as expected. 
6. Save the client-updater-source-groups condition.
7. Navigate back to the client-updater-source-groups condition type and verify that updating the data works as expected.

## Checklist:

## Additional Notes

Dependent on #1532  - will rebase once merged to simplify review process.

### DIFF
--- a/src/clients/scopes/AddScopeDialog.tsx
+++ b/src/clients/scopes/AddScopeDialog.tsx
@@ -298,6 +298,7 @@ export const AddScopeDialog = ({
           {
             name: "name",
           },
+          { name: "protocol", displayKey: "Protocol" },
           {
             name: "protocol",
             displayKey: "clients:protocol",
--- a/src/components/dynamic/DynamicComponents.tsx
+++ b/src/components/dynamic/DynamicComponents.tsx
@@ -13,7 +13,7 @@ export const DynamicComponents = ({ properties }: DynamicComponentProps) => (
   <>
     {properties.map((property) => {
       const componentType = property.type!;
-      if (isValidComponentType(componentType)) {
+      if (isValidComponentType(componentType) && property.name !== "scopes") {
         const Component = COMPONENTS[componentType];
         return <Component key={property.name} {...property} />;
       } else {
--- a/src/components/dynamic/MultivaluedChipsComponent.tsx
+++ b/src/components/dynamic/MultivaluedChipsComponent.tsx
@@ -1,0 +1,173 @@
+import React, { useState } from "react";
+import { useTranslation } from "react-i18next";
+import { Controller, useFormContext } from "react-hook-form";
+import {
+  Button,
+  Chip,
+  ChipGroup,
+  FormGroup,
+  TextInput,
+} from "@patternfly/react-core";
+
+import { HelpItem } from "../help-enabler/HelpItem";
+import type { ComponentProps } from "./components";
+import { AddScopeDialog } from "../../clients/scopes/AddScopeDialog";
+import { useAdminClient, useFetch } from "../../context/auth/AdminClient";
+import type ClientScopeRepresentation from "@keycloak/keycloak-admin-client/lib/defs/clientScopeRepresentation";
+import { useParams } from "react-router";
+import type { EditClientPolicyConditionParams } from "../../realm-settings/routes/EditCondition";
+import { GroupPickerDialog } from "../group/GroupPickerDialog";
+import type GroupRepresentation from "@keycloak/keycloak-admin-client/lib/defs/groupRepresentation";
+
+export const MultivaluedChipsComponent = ({
+  defaultValue,
+  name,
+}: ComponentProps) => {
+  const { t } = useTranslation("dynamic");
+  const { control } = useFormContext();
+  const { conditionName } = useParams<EditClientPolicyConditionParams>();
+  const adminClient = useAdminClient();
+  const [open, setOpen] = useState(false);
+  const [clientScopes, setClientScopes] = useState<ClientScopeRepresentation[]>(
+    []
+  );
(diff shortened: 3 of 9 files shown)


jenny-s51:
On it - thank you @jonkoops !

src/user/UserForm.tsx
```diff
@@ -145,13 +145,13 @@ export const UserForm = ({
             setOpen(false);
           }}
           onClose={() => setOpen(false)}
-          filterGroups={selectedGroups}
+          filterGroups={selectedGroups.map((group) => group.name) as string[]}
```
jonkoops, line 148: If you are sure that `name` is always present you can use a non-null assertion so there is no need to type-cast.
```suggestion
          filterGroups={selectedGroups.map((group) => group.name!)}
```

src/components/group/GroupPickerDialog.tsx
```diff
@@ -100,9 +100,10 @@ export const GroupPickerDialog = ({
   );
 
   const isRowDisabled = (row?: GroupRepresentation) => {
-    return !![...joinedGroups, ...(filterGroups || [])].find(
-      (group) => group.id === row?.id
-    );
+    return !![
+      ...joinedGroups.map((item) => item.name),
+      ...(filterGroups || []),
+    ].find((group) => group === row?.name);
```
jonkoops, line 106: This code can be cleaned up a bit by using [`.some()` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)instead of `.find()`.

```suggestion
    return [
      ...joinedGroups.map((item) => item.name),
      ...(filterGroups || []),
    ].some((group) => group === row?.name);
```

src/components/dynamic/MultivaluedChipsComponent.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    ok: "users:join",
+                  }}
+                  onConfirm={(groups) => {
+                    onChange([...value, ...groups.map((group) => group.name)]);
+                    setSelectedGroups([...selectedGroups!, ...groups]);
+                    setOpen(false);
+                  }}
+                  onClose={() => {
+                    setOpen(false);
+                    console.log(value);
```
jonkoops, line 115: Stray console log

jonkoops (changes requested):

src/components/group/GroupPickerDialog.tsx
jenny-s51: Ah nice catch! It definitely makes more sense to use `.some()` here

jonkoops (approved):

src/components/dynamic/DynamicComponents.tsx
```diff
@@ -13,7 +13,7 @@ export const DynamicComponents = ({ properties }: DynamicComponentProps) => (
   <>
     {properties.map((property) => {
       const componentType = property.type!;
-      if (isValidComponentType(componentType)) {
+      if (isValidComponentType(componentType) && property.name !== "scopes") {
```
edewit, line 16: This should not be nessary

src/components/dynamic/MultivaluedChipsComponent.tsx
```diff
@@ -0,0 +1,172 @@
...
+  );
+
+  const toggleModal = () => {
+    setOpen(!open);
+  };
+
+  return (
+    <FormGroup
+      label={
+        name === "scopes"
```
edewit, line 53: why not use the "label" field?

src/components/dynamic/MultivaluedChipsComponent.tsx
```diff
@@ -0,0 +1,172 @@
...
+          ? t("realm-settings:clientScopesCondition")
+          : t("common:groups")
+      }
+      id={name === "scopes" ? "expected-scopes" : "expected-groups"}
```
edewit, line 57: why not use the name field?

src/components/dynamic/MultivaluedChipsComponent.tsx
```diff
@@ -0,0 +1,172 @@
...
+      labelIcon={
+        <HelpItem
+          helpText={
+            name === "scopes"
```
edewit, line 61: why not use the `helpText`

src/components/dynamic/MultivaluedChipsComponent.tsx
```diff
@@ -0,0 +1,172 @@
...
+    >
+      <Controller
+        name={`config.${name}`}
+        control={control}
+        defaultValue={[defaultValue]}
+        rules={{ required: true }}
+        render={({ onChange, value }) => {
+          return (
+            <>
+              {open && name === "scopes" && (
```
edewit, line 83: let's not add a lot of extra logic based on the name of the fields, if we need extra ui logic it should use a different component type on the server side

src/clients/scopes/AddScopeDialog.tsx
```diff
@@ -298,6 +298,7 @@ export const AddScopeDialog = ({
           {
             name: "name",
           },
+          { name: "protocol", displayKey: "Protocol" },
```
edewit, line 301: image
Don't think we need two times protocol

Issue #1481, opened 2021-11-03
User Fed: role-ldap-mapper has wrong defaults


ssilvert:
1. Go to User Federation --> LDAP --> LDAP Mappers
2. Create new mapper
3. Choose `role-ldap-mapper` as mapper type

`Role Object Classes` has wrong default.  Should be `groupOfNames` instead of `group`
`User roles retrieve strategy` has three choices.  It should have only two.  See old console.

mfrances17:
The `Role Object Classes` issue will be fixed as part of https://github.com/keycloak/keycloak-admin-ui/pull/1511.

After testing and reproducing this issue, it seems that `User roles retrieve strategy` is another case of the Active Directory vendor having different values that the other 4 vendor types. In this case, there are 3 available choices for AD and just 2 for the other vendors.

Will have a fix soon, just wanted to add context to the issue.

PR #1578, opened 2021-11-19
Fix role-ldap-mapper defaults

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1481.

## Brief Description
As noted in https://github.com/keycloak/keycloak-admin-ui/issues/1481, the `Role Object Classes` issue was fixed as part of https://github.com/keycloak/keycloak-admin-ui/pull/1511.

For the `User roles retrieve strategy` issue, fixed it so the Active Directory vendor has three possible choices vs the other 4 vendor types which only have two available choices.

During testing, noticed that the field itself was not writing/reading correctly - this is because both roles and groups both use the user-roles-retrieve-strategy API value, but have different enums. It works correctly now.

And finally, with these changes, there were just far too many IF conditions within the Select so made one check for isRole and based on that load a different Select.

## Verification Steps
1. Go to User Federation > LDAP > LDAP Mappers.
2. Create new mapper.
3. Choose `role-ldap-mapper` as mapper type.
4. Verify that `Role Object Classes` has the correct defaults for the Active Directory vendor type (`group`) vs the other vendors (`groupOfNames`), and that you can save/read correctly, using the old console if needed.
5. Verify that for the Active Directory vendor type, `User roles retrieve strategy` has three choices, vs the other vendors which  should have only two choices, and that you can save/read correctly, using the old console if needed.
6. Repeat with the `group-ldap-mapper` mapper type. See old console for comparison purposes.

## Checklist:

## Additional Notes
None

### DIFF
--- a/src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
+++ b/src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
@@ -33,6 +33,7 @@ export const LdapMapperRoleGroup = ({
   const [isRetrieveStratDropdownOpen, setIsRetrieveStratDropdownOpen] =
     useState(false);
   const [isClientIdDropdownOpen, setIsClientIdDropdownOpen] = useState(false);
+  const [vendor, setVendor] = useState<string | undefined>();
   const [clients, setClients] = useState<ClientRepresentation[]>([]);
   const { id, mapperId } = useParams<UserFederationLdapMapperParams>();
 
@@ -61,7 +62,7 @@ export const LdapMapperRoleGroup = ({
     },
     (fetchedComponent) => {
       if (fetchedComponent) {
-        const vendor = fetchedComponent.config?.vendor[0];
+        setVendor(fetchedComponent.config?.vendor[0]);
         if (mapperId === "new" && vendor === "ad") {
           form.setValue(
             isRole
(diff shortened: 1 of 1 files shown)


src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -384,70 +385,107 @@ export const LdapMapperRoleGroup = ({
...
+                  LOAD_ROLES_BY_MEMBER_ATTRIBUTE
+                </SelectOption>
+                <SelectOption
+                  key={1}
+                  value="GET_ROLES_FROM_USER_MEMBEROF_ATTRIBUTE"
+                >
+                  GET_ROLES_FROM_USER_MEMBEROF_ATTRIBUTE
+                </SelectOption>
+                <SelectOption
+                  hidden={vendor !== "ad"}
```
jenny-s51, line 428:
```suggestion
                  hidden={vendorType !== "ad"}
```

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -384,70 +385,107 @@ export const LdapMapperRoleGroup = ({
...
+                  LOAD_GROUPS_BY_MEMBER_ATTRIBUTE
+                </SelectOption>
+                <SelectOption
+                  key={1}
+                  value="GET_GROUPS_FROM_USER_MEMBEROF_ATTRIBUTE"
+                >
+                  GET_GROUPS_FROM_USER_MEMBEROF_ATTRIBUTE
+                </SelectOption>
+                <SelectOption
+                  hidden={vendor !== "ad"}
```
jenny-s51, line 478:
```suggestion
                  hidden={vendorType !== "ad"}
```

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
```diff
@@ -61,7 +62,7 @@ export const LdapMapperRoleGroup = ({
     },
     (fetchedComponent) => {
       if (fetchedComponent) {
-        const vendor = fetchedComponent.config?.vendor[0];
+        setVendor(fetchedComponent.config?.vendor[0]);
```
jenny-s51, line 65: `vendor` is undefined on the initial load so the defaultValue always gets set to `groupOfNames`
```suggestion
        const vendor = fetchedComponent.config?.vendor[0];
        setVendorType(vendor);
```

jenny-s51 (changes requested):
`User roles retrieve strategy` is looking great and works as expected. 

It looks like when the vendor type is AD, and either `group-ldap-mapper` or `role-ldap-mapper` is selected, its value is initially undefined and therefore we always get `groupOfNames` in as the defaultValue (but we need `groups`). After some investigating it looks like you can rename the values in the hook to fix this (see comments)!

Otherwise LGTM 🙂

src/user-federation/ldap/mappers/LdapMapperRoleGroup.tsx
mfrances17: thanks @jenny-s51, good catch 👍  - this wouldn't have worked as originally coded, i thought i had tested that. should be fixed now.

jenny-s51 (approved):
Nice work @mfrances17 !

jonkoops (approved):

PR #1595, opened 2021-11-22
Realm settings(client policies): Add switches to client policies table

### LINKED ISSUE #1388: Enabled/Disabled missing from Client policies table
Client policies table should have a status column for enabled/disabled.  See https://marvelapp.com/prototype/2b3ghc06/screen/81581149

jenny-s51:
Closes #1388


ssilvert (approved):
LGTM

src/realm-settings/PoliciesTab.tsx
```diff
@@ -43,7 +44,7 @@ export const PoliciesTab = () => {
   const [code, setCode] = useState<string>();
   const [tablePolicies, setTablePolicies] =
     useState<ClientPolicyRepresentation[]>();
-
+  const form = useForm({ mode: "onChange" });
```
jonkoops, line 47: Can you provide a type for this form's value?

src/realm-settings/PoliciesTab.tsx
```diff
@@ -58,10 +59,62 @@ export const PoliciesTab = () => {
...
+        return policy;
+      }
+
+      const enabled = switchValues[updatedPolicy.name!];
+
+      return {
+        ...policy,
+        enabled: enabled,
+      };
+    }) as ClientPolicyRepresentation[];
```
jonkoops, line 76: Instead of using `as` (which might cast the type invalidly) you can provide a generic of the expected type to `map()`.
```suggestion
    const updatedPolicies = policies?.map<ClientPolicyRepresentation>(
      (policy) => {
        if (policy.name !== updatedPolicy.name) {
          return policy;
        }

        const enabled = switchValues[updatedPolicy.name!];

        return {
          ...policy,
          enabled: enabled,
        };
      }
    );
```

src/realm-settings/PoliciesTab.tsx
```diff
@@ -58,10 +59,62 @@ export const PoliciesTab = () => {
...
+
+    try {
+      await adminClient.clientPolicies.updatePolicy({
+        policies: updatedPolicies,
+      });
+      history.push(`/${realm}/realm-settings/clientPolicies`);
```
jonkoops, line 82: Can we use a route function here instead?

jonkoops (changes requested):

src/realm-settings/PoliciesTab.tsx
jenny-s51: Yes good idea! After changing this I discovered a bug in the function that updates the switch values... investigating...

jonkoops: Very nice! It's always great when TypeScript helps us find possible bugs.

src/realm-settings/PoliciesTab.tsx
```diff
@@ -43,9 +45,14 @@ export const PoliciesTab = () => {
...
   const [tablePolicies, setTablePolicies] =
     useState<ClientPolicyRepresentation[]>();
-
   const refresh = () => setKey(key + 1);
 
+  type EnabledConfigPolicyForm = {
+    [index: string]: boolean;
+  };
+
+  const form = useForm<EnabledConfigPolicyForm>({ mode: "onChange" });
```
jonkoops, line 54: TypeScript has a useful type for these kind of objects called [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) which you could use here. Both options are equally valid though, so it comes down to personal preference.

```suggestion
  const form = useForm<Record<string, boolean>>({ mode: "onChange" });
```

src/realm-settings/PoliciesTab.tsx
```diff
@@ -152,7 +209,7 @@ export const PoliciesTab = () => {
       <Divider />
       {!show ? (
         <KeycloakDataTable
-          key={policies.length}
+          key={key && policies.length}
```
jonkoops, line 212: I think this might break the refresh logic. If `key` is a 'truthy' value (1 or higher) the resulting value will always be `policies.length`.

jonkoops (changes requested):

src/realm-settings/PoliciesTab.tsx
jenny-s51: Ah yes, very cool! Thank you for suggesting this

jonkoops (approved):

PR #1572, opened 2021-11-18
Show exact commit hash of Admin UI in interface

jonkoops:
## Motivation
Shows the exact commit hash of the Admin UI in the user-interface when built with Maven. This allows us to find out the exact version the user is running so we can reliably reproduce bugs and other issues.

The commit hash is shown on the dashboard page below the server information.

A screenshot showing the commit hash

PR #1598, opened 2021-11-23
Include .git directory to fix Docker build

jonkoops:
Since merging #1572 the Docker build is no longer functional, this is because the Maven plugin that was included needs information from the `.git` directory. This PR allows this directory to be passed into the Docker build, which allows it to function again.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-.git
 cypress
 .github


edewit (approved):

agagancarczyk (approved):

jonkoops:
Merging this as it breaks the Docker setup, which is used by some for development.

PR #1600, opened 2021-11-24
Add useToggle hook to manage toggleable state

jonkoops:
## Motivation
A common pattern we are using is combining `useState(false)` with an accompanying function to toggle it's state to `true` and back. For example:

```tsx
const [isOpen, setIsOpen] = useState(false);
const toggleIsOpen = () => setIsOpen(!isOpen);

<button onClick={toggleIsOpen}>Toggle</button>
```

To reduce the amount of boilerplate code needed the new `useToggle()` hook does the same as follows:
```tsx
const [isOpen, toggleIsOpen] = useToggle();

<button onClick={toggleIsOpen}>Toggle</button>
```

It also still preserves the ability to set the value independently, for example:
```tsx
const [isOpen, toggleIsOpen, setIsOpen] = useToggle();

<button onClick={() => setIsOpen(true)}>Open</button>
<button onClick={() => setIsOpen(false)}>Close</button>
<button onClick={toggleIsOpen}>Toggle</button>
```

And allows you to start out with `true` as the default:
```tsx
const [isOpen, toggleIsOpen] = useToggle(true);
```

The hook also comes with some performance enhancements as it memoizes the callback, which can prevent accidental re-renders.

### DIFF
--- a/src/utils/useToggle.ts
+++ b/src/utils/useToggle.ts
@@ -1,0 +1,13 @@
+import { useCallback, useState } from "react";
+
+/**
+ * A hook that allows you toggle a boolean value, useful for toggle buttons, showing and hiding modals, etc.
+ *
+ * @param initialValue The initial value to use, false by default.
+ */
+export default function useToggle(initialValue = false) {
+  const [value, setValue] = useState(initialValue);
+  const toggleValue = useCallback(() => setValue((val) => !val), []);
+
+  return [value, toggleValue, setValue] as const;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "@testing-library/cypress": "^8.0.2",
         "@testing-library/jest-dom": "^5.15.0",
         "@testing-library/react": "^12.1.1",
+        "@testing-library/react-hooks": "^7.0.2",
         "@types/dagre": "^0.7.45",
         "@types/file-saver": "^2.0.4",
         "@types/lodash": "^4.14.177",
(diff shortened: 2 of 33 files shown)


src/utils/useToggle.ts
```diff
@@ -0,0 +1,13 @@
+import { useCallback, useState } from "react";
+
+/**
+ * A hook that allows you toggle a boolean value, useful for toggle buttons, showing and hiding modals, etc.
+ *
+ * @param initialValue The initial value to use, false by default.
+ */
+export default function useToggle(initialValue = false) {
+  const [value, setValue] = useState(initialValue);
+  const toggleValue = useCallback(() => setValue((val) => !val), []);
```
edewit, line 10: Why use `useCallback` when the dependency array is empty?

edewit (commented):
Although this is not a huge save, it's nice

src/utils/useToggle.ts
jonkoops: It's to memoize the function so it is stable and cannot accidentally cause re-renders of components or trigger side-effects.

edewit: Sure, but it will right as it kinda has to when the value changes something will need to rerender? I think in this case where you have no parameters it will not do anything

jonkoops: In this case an empty array is provided so it will be memoized without every changing, the value will still be updated as it is seperate from the callback. The value will also always be up-to-date for the state setter as it uses a callback style-which retrieves the latest value always.

edewit (approved):

PR #1601, opened 2021-11-25
initial version of the password policies

edewit:
[marvel design](https://marvelapp.com/prototype/5ifgiai/screen/75687759)

### DIFF
--- a/cypress/integration/authentication_test.spec.ts
+++ b/cypress/integration/authentication_test.spec.ts
@@ -7,6 +7,7 @@ import DuplicateFlowModal from "../support/pages/admin_console/manage/authentica
 import FlowDetails from "../support/pages/admin_console/manage/authentication/FlowDetail";
 import RequiredActions from "../support/pages/admin_console/manage/authentication/RequiredActions";
 import AdminClient from "../support/util/AdminClient";
+import PasswordPolicies from "../support/pages/admin_console/manage/authentication/PasswordPolicies";
 
 describe("Authentication test", () => {
   const loginPage = new LoginPage();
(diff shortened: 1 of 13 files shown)


mfrances17 (changes requested):
Initial version looks good. I was a little confused by the mechanism, in that as you add policies from the dropdown, it adds the relevant policy to the existing form. Once I got it though, it made sense. 

IMO this is good enough to go in as an early version, but if this were not labeled as 'initial' version, I would request these changes:

- On the empty state page, there needs to be more space between the 'You haven't...' label and the Add policy select.
- The hashing iterations number field needs to be enlarged - the number defaults are so large that when you click +/- you can't even tell if the number is incrementing/decrementing.
 - Help is missing on all the fields.
 - Do we need a required (*) on all of the fields once they are added? Seems weird.
 - Once you've selected all of the policies, you should probably disable the Add policy dropdown, currently it expands to an empty spacer.
 - For Not Username and Not Email, it seems weird to have them show up as switches - i don't believe the switches serve any purpose, they are not selectable and i don't think ever will be.
 - The fields are in **Title Case** but the rest of the UI is in **Sentence case**.
 - This is more of a ? for UX, but there really should be a way to delete all policies instead of clicking the - for each and every one.

jonkoops (approved):
Code looks good to me

mfrances17 (approved):

edewit:
> The fields are in Title Case but the rest of the UI is in Sentence case.
This is because I'm using the text that comes from the server I'll create an issue there to update this.

jonkoops (approved):

PR #1603, opened 2021-11-25
Prune older versions of published snapshots

jonkoops:
Because we keep publishing snapshot versions to Github Actions we are not exceeding our storage limit. This change will automatically prune older snapshot versions so this storage is freed.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,7 +24,17 @@ jobs:
           check-latest: true
           cache: maven
 
-      - name: Build and deploy Admin UI
-        run: mvn --batch-mode deploy
+      - name: Build Admin UI
+        run: mvn --batch-mode install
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -28,3 +28,10 @@ jobs:
         run: mvn --batch-mode deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Prune older published packages
```
jonkoops, line 32: I am not sure if this will work correctly, as the problem is that we keep pushing assets into an existing snapshot version. However if this approach fails I will move it before the publish so that it can prune all SNAPSHOTS before a new version is published.

agagancarczyk (approved):

PR #1628, opened 2021-11-30
Cypress test fixes

mfrances17:
## Motivation
Stabilize the cypress tests so that they more consistently pass locally and on CI.

## Brief Description
- Fixed tests that were failing on some combination of local browser, local headless, and CI. All tests should now pass when run locally on chrome (cypress open) or headless (cypress run) and they're passing on CI.
- Re-enabled screenshots/videos in the config.
- Chunked larger tests (Clients, Identity providers, Realm settings, User Fed LDAP mapper) that were causing the browser to crash due to out of memory issues into smaller tests.
- Added plugin to cypress index.ts so cypress will only save assets (video/screen caps) for tests that fail, instead of recording and storing them for all tests, even ones that pass.

## Verification Steps
Run the tests in any of the 3 modes mentioned above.

## Checklist:
N/A

## Additional Notes
None


edewit (approved):
This is great stuff, very cool @mfrances17

cypress/plugins/index.ts
```diff
@@ -36,6 +38,18 @@ const configurePlugins: Cypress.PluginConfig = (on) => {
...
+    if (results?.video) {
+      // Do we have failures for any retry attempts?
+      const failures = _.some(results.tests, (test: any) => {
+        return _.some(test.attempts, { state: "failed" });
+      });
+      if (!failures) {
+        // delete the video if the spec passed and no tests retried
+        return del(results.video);
+      }
+    }
```
jonkoops, line 51: I think it should be possible to accomplish the same without the need for external libraries.
```suggestion
    if (!results.video) {
      return;
    }

    // Do we have failures for any retry attempts?
    const failures = results.tests.some(({ attempts }) =>
      attempts.some(({ state }) => state === "failed")
    );

    // delete the video if the spec passed and no tests retried
    if (!failures) {
      fs.rmSync(results.video);
    }
```

package.json
```diff
@@ -30,6 +30,7 @@
     "@patternfly/react-icons": "^4.26.4",
     "@patternfly/react-table": "^4.44.4",
     "dagre": "^0.8.5",
+    "del": "^6.0.0",
```
jonkoops, line 33: It think we can use the [File system API](https://nodejs.org/api/fs.html#fsrmsyncpath-options) from Node instead of using a library.

jonkoops (changes requested):
Great work, the tests are looking great! I have a couple of comments around the deleting of video's but other than that I think this is good to go.

jonkoops (approved):
Approving this PR as the feedback can be applied at a later point

agagancarczyk (approved):
Great work @mfrances17 , great that these tests are going to be stable now.

jonkoops:
I'm merging this before our 24h review period agreement so that we can unblock the open PRs.

package.json
mfrances17: Great suggestion, thanks for the help

PR #1634, opened 2021-12-01
Wait for group to be deleted before navigating

### LINKED ISSUE #1625: Deleting a group from Group Details still shows the group
### Describe the bug

Deleting a group from Group Details still shows the group in the list until the page is refreshed and the message shown says "groupDelete"

### Version

15

### Expected behavior

The group should disappear from the list after it is deleted

### Actual behavior

The group only disappears from the list after the page is refreshed

### How to Reproduce?

Delete a group from the Group Details screen

### Anything else?

_No response_

jonkoops:
Closes #1625

### DIFF
--- a/src/groups/GroupsSection.tsx
+++ b/src/groups/GroupsSection.tsx
@@ -48,7 +48,7 @@ export default function GroupsSection() {
       await adminClient.groups.del({
         id: group.id!,
       });
-      addAlert(t("groupDelete"), AlertVariant.success);
+      addAlert(t("groupDeleted", { count: 1 }), AlertVariant.success);
     } catch (error) {
       addError("groups:groupDeleteError", error);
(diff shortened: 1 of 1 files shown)


agagancarczyk (approved):
lgtm

jenny-s51 (changes requested):
Looking good @jonkoops !

Can we change this line https://github.com/keycloak/keycloak-admin-ui/blob/main/src/groups/GroupsSection.tsx#L51

to `addAlert(t("groupDeleted_one"), AlertVariant.success);`? Looks like it wasn't getting the correct localized string before.

jonkoops:
@jenny-s51 Good catch! I added the change, can you take another look?

jenny-s51 (approved):

PR #1639, opened 2021-12-01
Trim names of client, realm role, user and group before saving

### LINKED ISSUE #1622: Can create a Client, Realm role, User and Group with seemingly duplicated name
### Describe the bug

When creating one of those items there is not a trim in the name, creating an item with the same name but spaces before or after the name lets you create the item but in the list it shows it without the extra spaces (trimmed)

### Version

15

### Expected behavior

Creating a new item with a duplicated name but adding spaces before or after shouldn't be allowed, there should be a trim and maybe a case check

### Actual behavior

Creating a new item with a duplicated name but adding spaces before or after is allowed and then when listed they can be confused

### How to Reproduce?

Create a client with the client ID " account" having already the "account" client

### Anything else?

_No response_

jonkoops:
Closes #1622

### DIFF
--- a/src/clients/ClientDetails.tsx
+++ b/src/clients/ClientDetails.tsx
@@ -274,6 +274,9 @@ export default function ClientDetails() {
           webOrigins,
           attributes,
         };
+
+        newClient.clientId = newClient.clientId?.trim();
+
         await adminClient.clients.update({ id: clientId }, newClient);
         setupForm(newClient);
         setClient(newClient);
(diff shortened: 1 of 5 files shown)


jenny-s51 (changes requested):
Trim function works great @jonkoops ! 

Just one small thing - can we update the error message for users when a duplicate user is created?

For clients, roles, and groups, I'm seeing an error message that states it is a duplicate like this: 
`Could not create group Top level group named '123' already exists.`

But for users the error is shown as: `userCreateError`

jonkoops:
@jenny-s51 Good catch! Looks like it was using the wrong key, should be resolved now.

jenny-s51 (approved):

PR #1640, opened 2021-12-01
better form convert

edewit:
We used to setup the react-hook form for eacht screen by iterating the loaded data:

```typescript
const setupForm = (client: ClientRepresentation) => {
    const { redirectUris, webOrigins, ...formValues } = client;
    form.reset(formValues);
    Object.entries(client).map((entry) => {
      if (entry[0] === "redirectUris" || entry[0] === "webOrigins") {
        form.setValue(entry[0], convertToMultiline(entry[1]));
      } else if (entry[0] === "attributes") {
        convertToFormValues(entry[1], "attributes", form.setValue);
      } else {
        form.setValue(entry[0], entry[1]);
      }
    });
  };
```

This would allow us to convert arrays into fields that react hook form can manage or change the way attributes works.
Because attributes would need to be stored as strings and react hook form would interpert them as objects so:

```javascript
{
  attributes: {
    "something.like.this": "value"
  }
}
```

Would turn into:

```javascript
{
  attributes: {
    something: {
      like: {
        this: "value"
      }
    }
  }
}
```

Our convert script would replace the dots with `-` and then later change that back again.
In this PR I'm changing how this all works, making it better:
So instead of iterating yourself each time you need to setup a form we call one function:

```typescript
const setupForm = (client: ClientRepresentation) => {
  convertToFormValues(client, setValue, ["redirectUris", "webOrigins"]);
};
```

That will setup the form take care of `attributes` and transform the `redirectsUris` and `weborigins` fields.
It will not replace the dots in the attributes field so we don't need to change our input names,
but it will convert / flatten it again when we save with:

```typescript
const save = async (client: ClientRepresentation) => {
  const updatedClient = convertFormValuesToObject(client);

//...
```


jonkoops (approved):
I love this change! Consider it approved if the tests pass.

jenny-s51 (approved):
> I love this change! Consider it approved if the tests pass.

+1 @jonkoops - great update @edewit !! Awesome that we can convert the form values like this now 🙂

mfrances17 (approved):
LGTM, this is such a great improvement! 

I'm a little worried about the amount of test failures, was hoping maybe it hadn't been rebased yet but looks like it was.

PR #1641, opened 2021-12-02
Changed the idP mapper form to use the new Dynamic forms

edewit:
This changes the way the idp mappers render based on the types instead of the premade form


src/components/dynamic/GroupComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const GroupComponent = ({ name, label, helpText }: ComponentProps) => {
+  const { t } = useTranslation("dynamic");
+  const [open, setOpen] = useState(false);
+  const { control } = useFormContext();
+
+  return (
+    <Controller
+      name={`config.${convertToHyphens(name!)}`}
+      defaultValue=""
+      typeAheadAriaLabel="Select an action"
```
jonkoops, line 26: This label should be translated.

src/components/dynamic/GroupComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <>
+          {open && (
+            <GroupPickerDialog
+              type="selectOne"
+              text={{
+                title: "dynamic:selectGroup",
+                ok: "common:select",
+              }}
+              onConfirm={(groups) => {
+                onChange(groups.map((g) => g.path)[0]);
```
jonkoops, line 38: If you are always getting this first group, can't we get rid of the map operation here? Something like:
```ts
onChange(groups[0].path);
```

src/components/dynamic/GroupComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+              <HelpItem
+                helpText={t(helpText!)}
+                forLabel={label!}
+                forID={name!}
+              />
+            }
+            fieldId={name!}
+          >
+            <InputGroup>
+              <ChipGroup categoryName={""}>
```
jonkoops, line 58: Can the category name prop be omitted?
```suggestion
              <ChipGroup>
```

src/components/dynamic/GroupComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+                {value && (
+                  <Chip onClick={() => onChange(undefined)}>{value}</Chip>
```
jonkoops, line 60: Should clicking a chip clear the value?

jonkoops (changes requested):

src/identity-providers/add/AddMapper.tsx
```diff
@@ -75,7 +73,7 @@ export default function AddMapper() {
   );
 
   const [mapperTypes, setMapperTypes] =
-    useState<Record<string, IdentityProviderMapperRepresentation>>();
+    useState<Record<string, IdentityProviderMapperTypeRepresentation>>();
   const [mapperType, setMapperType] = useState(
```
jenny-s51, line 77: Looks like these are not currently rendering the fields for the default mapper type.

Line 81: `"advancedAttributeToRole"` can be replaced with `"saml-advanced-role-idp-mapper"` and for line 82 we can use `"hardcoded-user-session-attribute-idp-mapper"` to render the correct fields by default.

The `attributeImporter` type here also cannot be used the way it was used previously since [lines 228 - 231](https://github.com/keycloak/keycloak-admin-ui/pull/1641/files#diff-99eab818b1103a8fb2fb717ae61ca05c92e919aeba6159455533f9624987469dL228-L231) were deleted

jenny-s51 (changes requested):
Excellent refactor @edewit !! Left a couple comments 🙂

src/components/dynamic/GroupComponent.tsx
edewit: no it renders something different if you do

src/components/dynamic/GroupComponent.tsx
edewit: yes

jonkoops (approved):
Code looks good to me, but it looks like the tests are still failing.

PR #1644, opened 2021-12-03
Complete partial import functionality.

ssilvert:
## Motivation
Completed Partial Import

## Brief Description
The import button is now active and it calls the REST endpoint.  Then it displays the results as per:
https://marvelapp.com/prototype/fg26f9b/screen/75624798

## Verification Steps
1. Go to `Realm settings`.
2. Select Partial import form the Action dropdown.
3. Import a previously exported Keycloak json. (Samples test data at `/cypress/fixtures/partial-import-test-data/`. kcexport.json is a good one to look at)
4. Verify that modal presents the user with options as shown in the design, plus multi-realm functionality.
5. Press import and see results.

## Checklist:

- [ ] Help has been implemented (not applicable)
- [X] Unit tests have been created/updated

### DIFF
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -120,7 +120,7 @@ export const PartialImportDialog = (props: PartialImportProps) => {
       setTargetRealm(value[0] || {});
     } else {
       setIsMultiRealm(false);
-      setTargetRealm((value as RealmRepresentation) || {});
+      setTargetRealm(value as RealmRepresentation);
     }
   };
 
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -471,7 +471,7 @@ export const PartialImportDialog = (props: PartialImportProps) => {
   };
 
   const TypeRenderer = (importRecord: { resourceType: string }) => {
-    const typeMap = new Map<string, string>([
+    const typeMap = new Map([
       ["CLIENT", t("common:clients")],
       ["REALM_ROLE", t("common:realmRoles")],
       ["USER", t("common:users")],
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -1,6 +1,5 @@
 import React, { useState, useEffect } from "react";
 import { useTranslation } from "react-i18next";
-import _ from "lodash";
 import {
   Alert,
   Button,
@@ -55,6 +54,15 @@ type CollisionOption = "FAIL" | "SKIP" | "OVERWRITE";
 
 type ResourceChecked = { [k in Resource]: boolean };
 
+const INITIAL_RESOURCES: Readonly<ResourceChecked> = {
+  users: false,
+  clients: false,
+  groups: false,
+  identityProviders: false,
+  realmRoles: false,
+  clientRoles: false,
+};
+
 export const PartialImportDialog = (props: PartialImportProps) => {
   const tRealm = useTranslation("realm-settings").t;
   const { t } = useTranslation("partial-import");
@@ -73,23 +81,11 @@ export const PartialImportDialog = (props: PartialImportProps) => {
   const [importResponse, setImportResponse] = useState<PartialImportResponse>();
   const { addError } = useAlerts();
 
-  const allResourcesUnChecked: Readonly<ResourceChecked> = {
-    users: false,
-    clients: false,
-    groups: false,
-    identityProviders: false,
-    realmRoles: false,
-    clientRoles: false,
-  };
-
-  const [resourcesToImport, setResourcesToImport] = useState<ResourceChecked>(
-    _.cloneDeep(allResourcesUnChecked)
-  );
-
+  const [resourcesToImport, setResourcesToImport] = useState(INITIAL_RESOURCES);
   const [isAnyResourceChecked, setIsAnyResourceChecked] = useState(false);
 
   const resetResourcesToImport = () => {
-    setResourcesToImport(_.cloneDeep(allResourcesUnChecked));
+    setResourcesToImport(INITIAL_RESOURCES);
     setIsAnyResourceChecked(false);
   };
 
@@ -129,9 +125,11 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     checked: boolean,
     event: React.FormEvent<HTMLInputElement>
   ) => {
-    const resource: Resource = event.currentTarget.name as Resource;
-    const copyOfResourcesToImport = _.cloneDeep(resourcesToImport);
-    copyOfResourcesToImport[resource] = checked;
+    const resource = event.currentTarget.name as Resource;
+    const copyOfResourcesToImport = {
+      ...resourcesToImport,
+      [resource]: checked,
+    };
     setResourcesToImport(copyOfResourcesToImport);
     setIsAnyResourceChecked(resourcesChecked(copyOfResourcesToImport));
   };
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -72,7 +72,6 @@ export const PartialImportDialog = (props: PartialImportProps) => {
 
   const [importedFile, setImportedFile] = useState<ImportedMultiRealm>();
   const isFileSelected = !!importedFile;
-  const isMultiRealm = Array.isArray(importedFile) && importedFile.length > 1;
   const [isRealmSelectOpen, setIsRealmSelectOpen] = useState(false);
   const [isCollisionSelectOpen, setIsCollisionSelectOpen] = useState(false);
   const [importInProgress, setImportInProgress] = useState(false);
@@ -135,23 +134,16 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     setResourcesToImport(copyOfResourcesToImport);
   };
 
-  const realmSelectOptions = () => {
-    if (!isMultiRealm) return [];
-
-    const mapper = (realm: RealmRepresentation) => {
-      return (
-        <SelectOption
-          key={realm.id}
-          value={realm}
-          data-testid={realm.id + "-select-option"}
-        >
-          {realm.realm || realm.id}
-        </SelectOption>
-      );
-    };
-
-    return (importedFile as RealmRepresentation[]).map(mapper);
-  };
+  const realmSelectOptions = (realms: RealmRepresentation[]) =>
+    realms.map((realm) => (
+      <SelectOption
+        key={realm.id}
+        value={realm}
+        data-testid={realm.id + "-select-option"}
+      >
+        {realm.realm || realm.id}
+      </SelectOption>
+    ));
 
   const handleCollisionSelect = (
     event: React.ChangeEvent<Element> | React.MouseEvent<Element, MouseEvent>,
@@ -339,7 +331,7 @@ export const PartialImportDialog = (props: PartialImportProps) => {
               <StackItem>
                 <Divider />
               </StackItem>
-              {isMultiRealm && (
+              {Array.isArray(importedFile) && importedFile.length > 0 && (
                 <StackItem>
                   <Text>{t("selectRealm")}:</Text>
                   <Select
@@ -349,7 +341,7 @@ export const PartialImportDialog = (props: PartialImportProps) => {
                     onSelect={(_, value) => handleRealmSelect(value)}
                     placeholderText={targetRealm.realm || targetRealm.id}
                   >
-                    {realmSelectOptions()}
+                    {realmSelectOptions(importedFile)}
                   </Select>
                 </StackItem>
               )}
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -270,10 +270,10 @@ export const PartialImportDialog = (props: PartialImportProps) => {
   const jsonForImport = () => {
     const jsonToImport: PartialImportRealmRepresentation = {
       ifResourceExists: collisionOption,
+      id: targetRealm.id,
+      realm: targetRealm.realm,
     };
 
-    jsonToImport.id = targetRealm.id;
-    jsonToImport.realm = targetRealm.realm;
     if (resourcesToImport["users"]) jsonToImport.users = targetRealm.users;
     if (resourcesToImport["groups"]) jsonToImport.groups = targetRealm.groups;
     if (resourcesToImport["identityProviders"])
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -104,15 +104,14 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     resetInputState();
   }, [props.open]);
 
-  const handleFileChange = (value: object) => {
+  const handleFileChange = (value: ImportedMultiRealm) => {
     resetInputState();
-
     setImportedFile(value);
 
-    if (value instanceof Array && value.length > 0) {
-      setTargetRealm(value[0] || {});
-    } else {
-      setTargetRealm(value as RealmRepresentation);
+    if (!Array.isArray(value)) {
+      setTargetRealm(value);
+    } else if (value.length > 0) {
+      setTargetRealm(value[0]);
     }
   };
 
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -308,19 +308,20 @@ export const PartialImportDialog = (props: PartialImportProps) => {
 
   async function doImport() {
     if (importInProgress) return;
+
+    setImportInProgress(true);
+
     try {
-      const jsonToImport: PartialImportRealmRepresentation = jsonForImport();
-      setImportInProgress(true);
-      const importResults: PartialImportResponse =
-        await adminClient.realms.partialImport({
-          realm: realm,
-          rep: jsonToImport,
-        });
+      const importResults = await adminClient.realms.partialImport({
+        realm,
+        rep: jsonForImport(),
+      });
       setImportResponse(importResults);
     } catch (error) {
-      setImportInProgress(false);
       addError("partial-import:importFail", error);
     }
+
+    setImportInProgress(false);
   }
 
   const importModal = () => {
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -127,11 +127,11 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     event: React.FormEvent<HTMLInputElement>
   ) => {
     const resource = event.currentTarget.name as Resource;
-    const copyOfResourcesToImport = {
+
+    setResourcesToImport({
       ...resourcesToImport,
       [resource]: checked,
-    };
-    setResourcesToImport(copyOfResourcesToImport);
+    });
   };
 
   const realmSelectOptions = (realms: RealmRepresentation[]) =>
--- a/src/realm-settings/PartialImport.tsx
+++ b/src/realm-settings/PartialImport.tsx
@@ -454,21 +454,29 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     return [];
   };
 
-  const ActionLabel = (importRecord: { action: string }) => {
-    let text = t("added");
-    let color: "green" | "orange" | "purple" = "green";
+  const ActionLabel = (importRecord: { action: string; id: string }) => {
+    const actionLabels = new Map([
+      [
+        "ADDED",
+        <Label key={importRecord.id} color="green">
+          {t("added")}
+        </Label>,
+      ],
+      [
+        "SKIPPED",
+        <Label key={importRecord.id} color="orange">
+          {t("skipped")}
+        </Label>,
+      ],
+      [
+        "OVERWRITTEN",
+        <Label key={importRecord.id} color="purple">
+          {t("overwritten")}
+        </Label>,
+      ],
+    ]);
 
-    if (importRecord.action === "SKIPPED") {
-      text = t("skipped");
-      color = "orange";
-    }
-
-    if (importRecord.action === "OVERWRITTEN") {
-      text = t("overwritten");
-      color = "purple";
-    }
-
-    return <Label color={color}>{text}</Label>;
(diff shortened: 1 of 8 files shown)


src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+          rep: jsonToImport,
+        });
+      setImportResponse(importResults);
+    } catch (error) {
+      setImportInProgress(false);
+      addError("partial-import:importFail", error);
+    }
+  }
+
+  const importModal = () => {
```
jonkoops, line 326: Technically this function is a valid React function component, perhaps we should be treating it as such. It doesn't have to leave this file but could be standalone.

```tsx
const ImportModal = () => { ... }

{!importResponse && <ImportModal />}

File: src/realm-settings/PartialImport.tsx [hunk 2]
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+  const importCompleteMessage = () => {
+    return t("importSuccess", {
+      added: importResponse?.added,
+      overwritten: importResponse?.overwritten,
+      skipped: importResponse?.skipped,
+    });
+  };
+
+  const loader = async (first?: number, max?: number) => {
+    let last = first! + max!;
```
jonkoops, line 444: Using the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) (`!`) here can cause exceptions at run-time as the value can be possibly `undefined`. When using this operator you are essentially saying to the compiler that it should not be catching these errors at compile time as you are 'sure' that this is fine.

A better solution here I think would be to provide some default values so that we're not disabling any compile-time checks.
```suggestion
  const loader = async (first = 0, max = 15) => {
    let last = first + max;
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+  const importCompleteMessage = () => {
+    return t("importSuccess", {
+      added: importResponse?.added,
+      overwritten: importResponse?.overwritten,
+      skipped: importResponse?.skipped,
+    });
+  };
+
+  const loader = async (first?: number, max?: number) => {
```
jonkoops, line 443: You can reduce the amount of non-null assertions and optional chaining here by using an early return on the presence of `importResponse` (like already at the end of this function).

```suggestion
  const loader = async (first?: number, max?: number) => {
    // Empty state can't happen as you will get back a result record
    // for each record you submit. This is here to make Typescript happy.
    if (!importResponse) {
      return [];
    }
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+    let last = first! + max!;
+    const length = importResponse?.results.length;
+    if (last > length!) last = length!;
```
jonkoops, line 446: Instead of this statement I think you could also use [`Math.min`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min) to keep the bounds here.

```ts
const last = Math.min(first + max, importResponse.results.length);
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+    if (importResponse?.results)
+      return importResponse.results.slice(first, last);
+
+    // Empty state can't happen as you will get back a result record
+    // for each record you submit. This is here to make Typescript happy.
+    return [];
+  };
```
jonkoops, line 454: If you take all the suggestions above you'll end up with something like this:

```ts
const loader = async (first = 0, max = 15) => {
  if (!importResponse) {
    return [];
  }

  const last = Math.min(first + max, importResponse.results.length);

  return importResponse.results.slice(first, last);
};
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+  const ActionLabel = (importRecord: { action: string }) => {
```
jonkoops, line 456: You can also directly reference the original type here. It's a tad verbose as the type was not extracted in the admin client library.
```suggestion
  const ActionLabel = (importRecord: PartialImportResponse["results"][0]) => {
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+    let text = t("added");
+    let color: "green" | "orange" | "purple" = "green";
```
jonkoops, line 458: You can reference the prop type here directly as well. 
```suggestion
    let color: LabelProps["color"] = "green";
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+    if (importRecord.action === "SKIPPED") {
+      text = t("skipped");
+      color = "orange";
+    }
+
+    if (importRecord.action === "OVERWRITTEN") {
```
jonkoops, line 465: Since these branches are mutually exclusive you might want to use `else if` here.
```suggestion
    } else if (importRecord.action === "OVERWRITTEN") {
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+      text = t("overwritten");
+      color = "purple";
+    }
+
+    return <Label color={color}>{text}</Label>;
+  };
+
+  const TypeRenderer = (importRecord: { resourceType: string }) => {
```
jonkoops, line 473: The original type can be referenced here as well.
```suggestion
  const TypeRenderer = (importRecord: PartialImportResponse["results"][0]) => {
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+    const typeMap = new Map<string, string>([
```
jonkoops, line 474: There is no need to provide a generic type for this make as it can be inferred from the value passed into the constructor.
```suggestion
    const typeMap = new Map([
```

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+      ["REALM_ROLE", t("common:realmRoles")],
+      ["USER", t("common:users")],
+      ["CLIENT_ROLE", t("common:clientRoles")],
+      ["IDP", t("common:identityProviders")],
+    ]);
+
+    return <span>{typeMap.get(importRecord.resourceType)}</span>;
+  };
+
+  const importCompletedModal = () => {
```
jonkoops, line 485: This is essentially a React functional component, perhaps we should use it like such.
```tsx
const ImportCompletedModal = () => { }
  
{importResponse && <ImportCompletedModal />}
```

src/realm-settings/PartialImport.tsx
```diff
@@ -153,7 +159,7 @@ export const PartialImportDialog = (props: PartialImportProps) => {
       );
     };
 
-    return (importedFile as [ImportedRealm]).map(mapper);
+    return (importedFile as [RealmRepresentation]).map(mapper);
```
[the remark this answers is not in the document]
ssilvert: That's true, but TypeScript can't figure it out without the `isArray`.  Using isMultiRealm is more clear.  

However, I do see that I've got the wrong type there.  It should be `RealmRepresentation[]`.  So I'll fix that.

src/realm-settings/PartialImport.tsx
```diff
@@ -233,7 +239,9 @@ export const PartialImportDialog = (props: PartialImportProps) => {
     return targetRealm[resource as NonRoleResource]!.length;
   };
 
-  const clientRolesCount = (clientRoles: { [index: string]: [] }) => {
+  const clientRolesCount = (clientRoles: {
+    [index: string]: RoleRepresentation[];
+  }) => {
```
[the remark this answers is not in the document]
ssilvert: Agree.  That's much better.

src/realm-settings/PartialImport.tsx
ssilvert, line 162:
```suggestion
    return (importedFile as RealmRepresentation[]).map(mapper);
```

src/realm-settings/PartialImport.tsx
ssilvert: That's an interesting way to do it, but unless you feel strongly, I'd rather not rewrite that section of the code.

src/realm-settings/PartialImport.tsx
ssilvert: I'll go one better.  How about no if statement at all?
```
  const ActionLabel = (importRecord: { action: string; id: string }) => {
    const actionLabels = new Map([
      [
        "ADDED",
        <Label key={importRecord.id} color="green">
          {t("added")}
        </Label>,
      ],
      [
        "SKIPPED",
        <Label key={importRecord.id} color="orange">
          {t("skipped")}
        </Label>,
      ],
      [
        "OVERWRITTEN",
        <Label key={importRecord.id} color="purple">
          {t("overwritten")}
        </Label>,
      ],
    ]);

    return actionLabels.get(importRecord.action);
  };
```

src/realm-settings/PartialImport.tsx
ssilvert: I don't understand why I would want or need the original type here.

src/realm-settings/PartialImport.tsx
ssilvert: I like that idea, but I found something interesting.  When I change this one to be a component, the dropdowns inside the modal only contain one item:
`{!importResponse && <ImportModal />}`

Any idea why that would be?  Somehow, the function is treated differently when used as a component.

src/realm-settings/PartialImport.tsx
ssilvert: BTW, I went ahead and added the above code.

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
-      </Stack>
-    </Modal>
+            </>
+          )}
+        </Stack>
+      </Modal>
+    );
+  };
+
+  const importCompleteMessage = () => {
```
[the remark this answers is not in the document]
ssilvert: I think that would look ugly.  I like to keep the markup compact so you can easily see its structure.

src/realm-settings/PartialImport.tsx
ssilvert: See my reply to your same concern on [ImportCompletedModal](https://github.com/keycloak/keycloak-admin-ui/pull/1644#discussion_r765094879).  Weird behavior when I changed this.

src/realm-settings/PartialImport.tsx
ssilvert: I don't understand that one.  Why would I want to do that?

src/realm-settings/PartialImport.tsx
ssilvert: Yes, that's much nicer.

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+          {isFileSelected && targetHasResources() && (
+            <>
               <StackItem>
-                <Text>{t("selectRealm")}:</Text>
+                <Divider />
+              </StackItem>
+              {isMultiRealm && (
+                <StackItem>
+                  <Text>{t("selectRealm")}:</Text>
```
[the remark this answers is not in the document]
ssilvert: I'm not sure that's true in this case.  The colon is not punctuation but rather part of the layout.

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+                    onSelect={handleRealmSelect}
+                    placeholderText={targetRealm.realm || targetRealm.id}
+                  >
+                    {realmSelectOptions()}
+                  </Select>
+                </StackItem>
+              )}
+              <StackItem>
+                <Text>{t("chooseResources")}:</Text>
+                <DataList aria-label="Resources to import" isCompact>
```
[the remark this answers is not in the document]
ssilvert: Done.

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,258 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+          </StackItem>
+          <StackItem>
+            <JsonFileUpload
+              id="partial-import-file"
+              allowEditingUploadedText
+              onChange={handleFileChange}
+            />
+          </StackItem>
+
+          {isFileSelected && targetHasResources() && (
```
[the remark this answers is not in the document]
ssilvert: I think that would be overkill for this.

src/realm-settings/PartialImport.tsx
```diff
@@ -278,107 +285,264 @@ export const PartialImportDialog = (props: PartialImportProps) => {
...
+
+  const importCompleteMessage = () => {
+    return t("importSuccess", {
+      added: importResponse?.added,
+      overwritten: importResponse?.overwritten,
+      skipped: importResponse?.skipped,
+    });
+  };
+
+  const loader = async (first = 0, max = 15) => {
```
edewit, line 444: These defaults have no effect, maybe we should change the loader type

src/realm-settings/PartialImport.tsx
edewit: you are now declaring `imported` record as a new type when you apply the suggestion by @jonkoops the type will be the same as `PartialImportResponse` and not duplicated e.g.:
```typescript
importRecord: {
    action: string;
    resourceType: string;
    resourceName: string;
    id: string;
} 
```

src/realm-settings/messages.ts
```diff
@@ -743,9 +743,16 @@ export default {
     selectIfResourceExists:
       "If a resource already exists, specify what should be done",
     import: "Import",
+    resourcesToImport: "Resources to import",
+    importFail: "Import failed: {{error}}",
     FAIL: "Fail import",
     SKIP: "Skip",
     OVERWRITE: "Overwrite",
+    importSuccess:
+      "{{added}} records added. {{skipped}} records skipped. {{overwritten}} records overwritten.",
```
edewit, line 752: You could split this up into separate parts:
```typescript
importSuccessRoutes_zero: "no records where added",
importSuccessRoutes_one: "one record added.",
importSuccessRoutes_other: "{{count}} records added.",
```
And then use it like this:
```typescript
t("importSuccessRoutes", {
      count: importResponse?.added,
})
```

edewit (changes requested):

src/realm-settings/PartialImport.tsx
ssilvert: Perhaps so.

src/realm-settings/PartialImport.tsx
```diff
@@ -59,15 +58,20 @@ type ResourceChecked = { [k in Resource]: boolean };
...
   const [isFileSelected, setIsFileSelected] = useState(false);
   const [isMultiRealm, setIsMultiRealm] = useState(false);
   const [importedFile, setImportedFile] = useState<ImportedMultiRealm>([]);
   const [isRealmSelectOpen, setIsRealmSelectOpen] = useState(false);
   const [isCollisionSelectOpen, setIsCollisionSelectOpen] = useState(false);
+  const [importInProgress, setImportInProgress] = useState(false);
   const [collisionOption, setCollisionOption] =
     useState<CollisionOption>("FAIL");
-  const [targetRealm, setTargetRealm] = useState<ImportedRealm>({});
+  const [targetRealm, setTargetRealm] = useState<RealmRepresentation>({});
```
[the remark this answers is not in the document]
jonkoops: @ssilvert why was this one marked as resolved? I still see the original code.

src/realm-settings/messages.ts
ssilvert: That's nice.  Fixed.

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
jonkoops: I have to disagree that `isMultiRealm` is the correct solution here, for the reason that this code can compile and still produce run-time errors or other unexpected behaviour. When the `as` keyword is used a subset of the type-checking in the compiler is disabled, essentially you are telling the compiler that you know better than it what type a value is. 

TypeScript does in no way provide any runtime guarantees for these kinds of type-casts as they are simply compiled away in the translation to JavaScript.


For example, if I were to set `isMultiRealm` to `true` because I am a human that can make mistakes this code will produce a runtime exception:

An example of such a runtime exception

This is further exacerbated by the fact that  `isMultiRealm` is not a computed value, but a separate state. Which increases the possibility for errors simply by forgetting to update the state when the `importedFile` state changes. A better solution would be to derive the value of `isMultiRealm` directly from the `importedFile` state:

```ts
  const [importedFile, setImportedFile] = useState<ImportedMultiRealm>([]);
  const isMultiRealm = Array.isArray(importedFile);
```

This guarantees that the value remains correct without the need for manually setting state to do so. However this does not resolve issues of type-casting a value based on other state.

src/realm-settings/PartialImport.tsx
ssilvert: I still don't see why that's better.  You have to know what `[0]` means.  Seems to me that my way is more clear to the reader.

src/realm-settings/PartialImport.tsx
```diff
@@ -114,15 +120,15 @@ export const PartialImportDialog = (props: PartialImportProps) => {
       setTargetRealm(value[0] || {});
     } else {
       setIsMultiRealm(false);
-      setTargetRealm((value as ImportedRealm) || {});
+      setTargetRealm((value as RealmRepresentation) || {});
     }
   };
 
   const handleRealmSelect = (
     event: React.ChangeEvent<Element> | React.MouseEvent<Element, MouseEvent>,
```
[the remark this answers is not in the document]
ssilvert: Good idea.  Changed.

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
jonkoops: True, in this case it might be overkill. Consider it resolved.

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
jonkoops: Fair enough.

src/realm-settings/PartialImport.tsx
jonkoops: It would allow you to use all the fields from this object without creating a duplicate type for it. However it's perhaps not the best option. Consider this one resolved.

src/realm-settings/PartialImport.tsx
jonkoops: Sure I have no preference here, just a note that this is possible.

src/realm-settings/PartialImport.tsx
jonkoops: This is a good improvement, but you could take it a step further and remove the need to keep a `Map` around by using a `switch` statement and returning the values directly. This might also be more performant as it does not allocate the extra elements in memory and is likely more easily optimized by the JIT.

```tsx
const ActionLabel = (importRecord: { action: string; id: string }) => {
  switch (importRecord.action) {
    case "ADDED":
      return (
        <Label key={importRecord.id} color="green">
          {t("added")}
        </Label>
      );
    case "SKIPPED":
      return (
        <Label key={importRecord.id} color="green">
          {t("added")}
        </Label>
      );
    case "OVERWRITTEN":
      return (
        <Label key={importRecord.id} color="purple">
          {t("overwritten")}
        </Label>
      );
  }
};
```

src/realm-settings/PartialImport.tsx
jonkoops: Exactly, however as long as they types share the same fields this should not be a problem per se.

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
ssilvert: isMultiRealm is computed in lines 118-124 and it's just a saved value of what you are suggesting.  If that isn't set correctly you will have other problems.

`const isMultiRealm = Array.isArray(importedFile);` is not correct because a one-element array is not multi-realm.

So to do it your way I would need both:
`if (!isMultiRealm) return [];`
`if (!Array.isArray(importedFile)) {
  return [];
}`

which is redundant.  

I guess it wouldn't hurt though, so if you want me to add that redundant check just to get rid of the cast I'll do it.

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
ssilvert: Sorry.  I didn't mean to mark resolved without an explanation.  I don't see why there is a problem with setting a default value.  Then I never have to worry about doing operations on an undefined value.

src/realm-settings/PartialImport.tsx
jonkoops: Yeah, I think we can keep things like they are for now.

src/realm-settings/PartialImport.tsx
ssilvert: Agreed off line to leave as-is.

jonkoops (approved):

src/realm-settings/PartialImport.tsx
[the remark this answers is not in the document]
jonkoops: I guess if it becomes a problem we can always add a quick fix for it. Let's consider this resolved.

edewit (approved):

PR #1663, opened 2021-12-06
Edit message bundle values

### LINKED ISSUE #684: Realm settings (localization): implement editable rows
https://marvelapp.com/prototype/56d9606/screen/76458141

jenny-s51:
## Motivation

https://issues.redhat.com/browse/CLUX-162
Closes #684 

## Verification Steps

1. Go to Realm Settings -> Localization
2. Enable internationalization
3. Save and add a message bundle to the data table below.
4. Click the edit icon to the right of the key/value pair you added.
5. Verify that the value can be edited, updated, and saved correctly.

## Checklist:

## Additional Notes

Tests will be added in separate PR because #1677 must be fixed first


edewit (changes requested):
I think it's better to not use the `KeycloakDatatable` for this and instead use a custom table. We already do this in other places if we add to much edge cases to this generic component it becomes bloated.

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +100,159 @@ export const LocalizationTab = ({
...
+    } else {
+      newRow = applyCellEdits(newRows[rowIndex], type);
+    }
+    newRows[rowIndex] = newRow;
 
-      return Object.entries(result);
-    } catch (error) {
-      return [];
+    // Update the copy of the retrieved data set so we can save it when the user saves changes
+    if (!invalid && type === RowEditAction.save) {
```
jonkoops, line 232: You could reduce the nesting here by escaping on the inverse of this condition.
```ts
setTableRows(newRows);

if (invalid || type !== RowEditAction.save) {
  return;
}
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +100,159 @@ export const LocalizationTab = ({
...
+      const key = (newRow.cells?.[0] as IRowCell).props.value;
+      const value = (newRow.cells?.[1] as IRowCell).props.value;
+
+      // We only have one editable value, otherwise we'd need to save each
+      try {
+        adminClient.setConfig({
+          requestConfig: { headers: { "Content-Type": "text/plain" } },
+        });
+        adminClient.realms.addLocalization(
```
jonkoops, line 241: This API call needs to be awaited.
```suggestion
        await adminClient.realms.addLocalization(
```

jonkoops (changes requested):

src/realm-settings/LocalizationTab.tsx
jenny-s51: That is a great suggestion - though I tested this and the behavior did not work as expected, since we want to call a different function for each row edit action

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +100,159 @@ export const LocalizationTab = ({
...
+        );
+
+        result = Object.fromEntries(filtered);
+      }
+
+      return { result };
+    },
+    ({ result }) => {
+      setMessageBundles(Object.entries(result).slice(first, first + max + 1));
+      return Object.entries(result).slice(first, first + max + 1);
```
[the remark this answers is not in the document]
jenny-s51: Good idea 👍

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +100,159 @@ export const LocalizationTab = ({
...
+  }, [messageBundles]);
+
+  useFetch(
+    async () => {
+      const params = {
+        first,
+        max,
+      };
+      let result = await adminClient.realms.getRealmLocalizationTexts({
+        ...params,
```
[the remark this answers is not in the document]
jenny-s51: Done! Thank you for your approval on https://github.com/keycloak/keycloak-nodejs-admin-client/pull/326 - when you have a moment, would be great if you could kick off a rerun of the tests as I no longer have write access to this repo.

edewit (approved):

mfrances17 (approved):
LGTM, code looked good and it tested as expected.

src/realm-settings/LocalizationTab.tsx
```diff
@@ -364,37 +553,38 @@ export const LocalizationTab = ({
...
+                  hasIcon={true}
+                  icon={SearchIcon}
+                  isSearchVariant={true}
+                  message={t("common:noSearchResults")}
+                  instructions={t("common:noSearchResultsInstructions")}
+                />
+              )}
+              {messageBundles.length !== 0 && (
+                <Table
+                  aria-label="editable-rows-table"
```
jonkoops, line 576: The `aria-label` attribute should be a human-readable string, in this case it should be localized as well.

src/realm-settings/LocalizationTab.tsx
```diff
@@ -364,37 +553,38 @@ export const LocalizationTab = ({
...
-                {
-                  name: "Value",
-                  cellRenderer: (row) => row[1],
-                },
-              ]}
-            />
+              )}
+              {messageBundles.length === 0 && filter && (
+                <ListEmptyState
+                  hasIcon={true}
```
jonkoops, line 567: Boolean props that are `true` can have their value omitted.
```suggestion
                  hasIcon
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -364,37 +553,38 @@ export const LocalizationTab = ({
...
+                  icon={SearchIcon}
+                  isSearchVariant={true}
```
jonkoops, line 569: Boolean props that are `true` can have their value omitted.
```suggestion
                  isSearchVariant
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -154,9 +321,9 @@ export const LocalizationTab = ({
         realmName: currentRealm!,
       });
       refreshTable();
-      addAlert(t("pairCreatedSuccess"), AlertVariant.success);
+      addAlert(t("addMessageBundleSuccess"), AlertVariant.success);
     } catch (error) {
-      addError("realm-settings:pairCreatedError", error);
+      addError("addMessageBundleError", error);
```
jonkoops, line 326: I believe the key passed here should be a full path as `addError` does  a call to get the translation from the root scope. If we leave it like this I think the message "addMessageBundleError" will be shown in the error message.

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +104,176 @@ export const LocalizationTab = ({
...
+    // Update the copy of the retrieved data set so we can save it when the user saves changes
+
+    if (!invalid && type === RowEditAction.save) {
+      const key = (newRow.cells?.[0] as IRowCell).props.value;
+      const value = (newRow.cells?.[1] as IRowCell).props.value;
+
+      // We only have one editable value, otherwise we'd need to save each
+      try {
+        adminClient.setConfig({
+          requestConfig: { headers: { "Content-Type": "text/plain" } },
```
jonkoops, line 260: This is not necessary anymore as the admin client should set this headers itself (see [here](https://github.com/keycloak/keycloak-nodejs-admin-client/blob/121231eaeb21be54aca9315df82f9f11d91c7742/src/resources/realms.ts#L330)).

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +104,176 @@ export const LocalizationTab = ({
...
+      const params = {
+        first,
+        max,
+      };
+      let result = await adminClient.realms.getRealmLocalizationTexts({
+        ...params,
         realm: realm.realm!,
         selectedLocale:
-          selectMenuLocale || getValues("defaultLocale") || DEFAULT_LOCALE,
+          selectMenuLocale || getValues("defaultLocale") || whoAmI.getLocale,
```
jonkoops, line 126: `getLocale()` is a function, it should be executed.
```suggestion
          selectMenuLocale || getValues("defaultLocale") || whoAmI.getLocale(),
```

src/realm-settings/LocalizationTab.tsx
```diff
@@ -76,31 +104,176 @@ export const LocalizationTab = ({
...
+  const [first, setFirst] = useState(0);
+  const [filter, setFilter] = useState("");
+
+  const refreshTable = () => {
+    setTableKey(tableKey + 1);
+  };
+
+  useFetch(
+    async () => {
+      const params = {
```
jonkoops, line 118: These params can be inlined into the `getRealmLocalizationTexts()` call below.

jonkoops (dismissed):

src/realm-settings/LocalizationTab.tsx
jenny-s51: Nice catch, thank you

PR #1711, opened 2021-12-14
UX Review: Updates to User Address Mapper 

### LINKED ISSUE #1659: [UX Review] Change the fields' order in the User Address mapper
### Describe the bug

In the new UI, I found the order of the fields is different from the current console. We can change the order to align with the current console. In addition, these three labels are confusing. We can refer to the current console to correct them.

**New UI**
Screen Shot 2021-12-06 at 2 40 46 PM

**Proposal**
Screen Shot 2021-12-06 at 2 49 12 PM

### Version

New Admin UI 

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

jenny-s51:
Closes #1659

### DIFF
--- a/src/client-scopes/details/MappingDetails.tsx
+++ b/src/client-scopes/details/MappingDetails.tsx
@@ -1,14 +1,18 @@
 import React, { useState } from "react";
 import { Link, useHistory, useParams, useRouteMatch } from "react-router-dom";
 import { useTranslation } from "react-i18next";
-import { FormProvider, useForm } from "react-hook-form";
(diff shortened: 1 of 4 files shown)


xianli123 (approved):

edewit (commented):
I don't think we should fix this like this. When we start making exceptions for these dynamic forms we'll end up with a lot of conditional rendering. My suggestion is either we change it on the backend or we leave it like it is

PR #1713, opened 2021-12-15
Finish the policy pages of the authentication section

edewit:
see [marvel](https://marvelapp.com/prototype/5ifgiai/screen/75687758)


src/authentication/policies/WebauthnPolicy.tsx
```diff
@@ -0,0 +1,382 @@
...
+            data-testid="webAuthnPolicyRpEntityName"
+            validated={errors.webAuthnPolicyRpEntityName ? "error" : "default"}
+          />
+        </FormGroup>
+        <FormProvider {...form}>
+          <WebauthnSelect
+            name={`${namePrefix}SignatureAlgorithms`}
+            label="webAuthnPolicySignatureAlgorithms"
+            options={SIGNATURE_ALGORITHMS}
+            isMultiSelect
```
mfrances17, line 244: i'm not able to use the `Signature algorithms` drop-down. Something seems to be conflicting with `isMultiSelect` because when i comment it out, the dropdown works and displays all `SIGNATURE_ALGORITHMS` values (but obviously is not multi-select). But when `isMultiSelect` is left in, the Select does not drop down and display the values.

mfrances17 (changes requested):
Just one issue I found with the `Signature algorithms` select not working as expected. Otherwise, code looks good and everything works as expected.

PR #1717, opened 2021-12-16
Add toggle for user profile to realm settings

jonkoops:
## Verification Steps
1. Start the Keycloak server with the `DECLARATIVE_USER_PROFILE` feature enabled (`start.mjs` will do this for you).
2. Go to realm settings and enable the 'User Profile Enabled' switch
3. Verify the 'User Profile' tab is visible

### DIFF
--- a/src/realm-settings/GeneralTab.tsx
+++ b/src/realm-settings/GeneralTab.tsx
@@ -18,6 +18,7 @@ import {
 
 import type RealmRepresentation from "@keycloak/keycloak-admin-client/lib/defs/realmRepresentation";
 import { getBaseUrl } from "../util";
+import useIsFeatureEnabled, { Feature } from "../utils/useIsFeatureEnabled";
 import { useAdminClient } from "../context/auth/AdminClient";
 import { useRealm } from "../context/realm-context/RealmContext";
 import { FormAccess } from "../components/form-access/FormAccess";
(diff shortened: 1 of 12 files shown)


src/realm-settings/GeneralTab.tsx
```diff
@@ -166,6 +168,35 @@ export const RealmSettingsGeneralTab = ({
             )}
           />
         </FormGroup>
+        {isFeatureEnabled(Feature.DeclarativeUserProfile) && (
```
edewit, line 171: shouldn't it be if feature enabled and user profile realm switch enabled?

edewit (approved):

mfrances17 (approved):
LGTM, works as expected, nice job. The approval is assuming that the `Attributes` and `Attributes group` tab contents are incomplete and TBD.

src/realm-settings/GeneralTab.tsx
jonkoops: Yeah, this condition is for the toggle itself not the tab (see [here](https://github.com/keycloak/keycloak-admin-ui/pull/1717/files#diff-385f6ae54f2531fbacfd9d3b9036588968189956cb45d8ccb473f566e58b5707R434-R435))

jonkoops:
> The approval is assuming that the `Attributes` and `Attributes group` tab contents are incomplete and TBD.

Yes, that is correct. The rest of the implementation will follow in a couple of smaller PRs to keep the review workload to a minimum.

PR #1750, opened 2021-12-21
initial version of the scopes screen

edewit:
[marvel](https://marvelapp.com/prototype/5fi097g/screen/83009424)

### DIFF
--- a/src/clients/authorization/ScopeDetails.tsx
+++ b/src/clients/authorization/ScopeDetails.tsx
@@ -1,0 +1,210 @@
+import React, { useState } from "react";
+import { Link, useHistory, useParams } from "react-router-dom";
+import { useForm } from "react-hook-form";
+import { useTranslation } from "react-i18next";
+import {
+  ActionGroup,
+  AlertVariant,
+  Button,
+  ButtonVariant,
+  DropdownItem,
+  FormGroup,
+  PageSection,
+  TextInput,
+  ValidatedOptions,
+} from "@patternfly/react-core";
+
+import type ScopeRepresentation from "@keycloak/keycloak-admin-client/lib/defs/scopeRepresentation";
+import type { ScopeDetailsParams } from "../routes/Scope";
+import { FormAccess } from "../../components/form-access/FormAccess";
+import { HelpItem } from "../../components/help-enabler/HelpItem";
+import { ViewHeader } from "../../components/view-header/ViewHeader";
+import { toClient } from "../routes/Client";
+import { useAdminClient, useFetch } from "../../context/auth/AdminClient";
+import { useAlerts } from "../../components/alert/Alerts";
+import useToggle from "../../utils/useToggle";
+import { DeleteScopeDialog } from "./DeleteScopeDialog";
+
+export default function ScopeDetails() {
+  const { t } = useTranslation("clients");
+  const { id, scopeId, realm } = useParams<ScopeDetailsParams>();
+  const history = useHistory();
+
+  const adminClient = useAdminClient();
+  const { addAlert, addError } = useAlerts();
+
+  const [deleteDialog, toggleDeleteDialog] = useToggle();
+  const [scope, setScope] = useState<ScopeRepresentation>();
+  const { register, errors, reset, handleSubmit } =
+    useForm<ScopeRepresentation>({
+      mode: "onChange",
+    });
+
+  useFetch(
+    async () => {
+      if (scopeId) {
+        const scope = await adminClient.clients.getAuthorizationScope({
+          id,
+          scopeId,
+        });
+        if (!scope) {
+          throw new Error(t("common:notFound"));
+        }
+        return scope;
+      }
+    },
+    (scope) => {
+      setScope(scope);
+      reset({ ...scope });
+    },
+    []
+  );
+
+  const save = async (scope: ScopeRepresentation) => {
+    try {
+      if (scopeId) {
+        await adminClient.clients.updateAuthorizationScope(
+          { id, scopeId },
+          scope
+        );
+      } else {
+        await adminClient.clients.createAuthorizationScope(
+          { id },
+          {
+            name: scope.name!,
+            displayName: scope.displayName,
+            iconUri: scope.iconUri,
+          }
+        );
+        history.push(toClient({ realm, clientId: id, tab: "authorization" }));
+      }
+      addAlert(
+        t((scopeId ? "update" : "create") + "ScopeSuccess"),
+        AlertVariant.success
+      );
+    } catch (error) {
+      addError("clients:scopeSaveError", error);
+    }
+  };
+
+  return (
+    <>
+      <DeleteScopeDialog
+        clientId={id}
+        open={deleteDialog}
+        toggleDialog={toggleDeleteDialog}
+        selectedScope={scope}
+        refresh={() =>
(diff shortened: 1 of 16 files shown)


src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+    async () => {
+      const params = {
+        first,
+        max,
+        deep: false,
+      };
+      const scopes = (await adminClient.clients.listAllScopes({
+        ...params,
+        id: clientId,
+      })) as ExpandableScopeRepresentation[];
```
jonkoops, line 75: This cast is not accurate as it's missing the required `isExpanded` field. Might be better to pass a generic into the `map()` down below.

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { ListEmptyState } from "../../components/list-empty-state/ListEmptyState";
+import useToggle from "../../utils/useToggle";
+import { DeleteScopeDialog } from "./DeleteScopeDialog";
+
+type ScopesProps = {
+  clientId: string;
+};
+
+export type ExpandableScopeRepresentation = ScopeRepresentation & {
+  resources: ResourceRepresentation[] | undefined;
```
jonkoops, line 43: The `resources` property already exists on `ScopeRepresentation`, I think this can be removed.
```suggestion
```

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+  permissions: PolicyRepresentation[] | undefined;
```
jonkoops, line 44: If a field is optional you can use the following shorthand.
```suggestion
  permissions?: PolicyRepresentation[];
```

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+  const [key, setKey] = useState(0);
+  const refresh = () => setKey(key + 1);
+
+  const [max, setMax] = useState(10);
+  const [first, setFirst] = useState(0);
+
+  useFetch(
+    async () => {
+      const params = {
```
jonkoops, line 67: These parameters should be in-lined in the call below, if they are missing in the admin client they should be added there.

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+        ...params,
+        id: clientId,
+      })) as ExpandableScopeRepresentation[];
+      await Promise.all(
+        scopes.map(async (s) => {
+          s.resources = await adminClient.clients.listAllResourcesByScope({
+            id: clientId,
+            scopeId: s.id!,
+          });
+          s.permissions = await adminClient.clients.listAllPermissionsByScope({
```
jonkoops, line 82: These API calls should happen in parallel, now they are waiting for each other.

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+            id: clientId,
+            scopeId: s.id!,
+          });
+        })
+      );
+      return scopes;
+    },
+    (scopes) =>
+      setScopes(scopes.map((scope) => ({ ...scope, isExpanded: false }))),
```
jonkoops, line 91: This mapping procedure should happen before in the fetching logic so that the type is immediately correct.

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+        first,
+        max,
+        deep: false,
+      };
+      const scopes = (await adminClient.clients.listAllScopes({
+        ...params,
+        id: clientId,
+      })) as ExpandableScopeRepresentation[];
+      await Promise.all(
+        scopes.map(async (s) => {
```
jonkoops, line 77:
```suggestion
        scopes.map<Promise<ExpandableScopeRepresentation>>(async (s) => {
```

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+          s.resources = await adminClient.clients.listAllResourcesByScope({
```
jonkoops, line 78: This assignment mutates the object directly, in a `map()` operation a new version should be returned that does not mutate the original reference. Same goes for `permissions`.

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+            scopeId: s.id!,
+          });
+        })
+      );
+      return scopes;
+    },
+    (scopes) =>
+      setScopes(scopes.map((scope) => ({ ...scope, isExpanded: false }))),
+    [key]
+  );
```
jonkoops, line 93: If you incorporate all my suggestions from before you'll end up with this: 
```tsx
useFetch(
  async () => {
    const scopes = await adminClient.clients.listAllScopes({
      first,
      max,
      deep: false,
      id: clientId,
    });

    return await Promise.all(
      scopes.map<Promise<ExpandableScopeRepresentation>>(async (scope) => {
        const options = { id: clientId, scopeId: scope.id! };
        const [resources, permissions] = await Promise.all([
          adminClient.clients.listAllResourcesByScope(options),
          adminClient.clients.listAllPermissionsByScope(options),
        ]);

        return {
          ...scope,
          resources,
          permissions,
          isExpanded: false,
        };
      })
    );
  },
  (scopes) => setScopes(scopes),
  [key]
);
```

src/clients/authorization/Scopes.tsx
```diff
@@ -0,0 +1,285 @@
...
+                  </Td>
+                  <Td>
+                    <PermissionsRenderer row={scope} />
+                  </Td>
+                  <Td
+                    actions={{
+                      items: [
+                        {
+                          title: t("common:delete"),
+                          onClick: async () => {
```
jonkoops, line 208:
```suggestion
                          onClick () => {
```

jonkoops (changes requested):

src/clients/authorization/Scopes.tsx
edewit: it doesn't need to be, but it could

src/clients/authorization/Scopes.tsx
edewit: it doesn't matter that these objects are mutated at this point as they are not yet part of react scope, but I like your change

src/clients/authorization/ScopeDetails.tsx
```diff
@@ -0,0 +1,210 @@
...
+      setScope(scope);
+      reset({ ...scope });
+    },
+    []
+  );
+
+  const save = async (scope: ScopeRepresentation) => {
+    try {
+      if (scopeId) {
+        await adminClient.clients.updateAuthorizationScope(
```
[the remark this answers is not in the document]
edewit: no the form still has the values that the user entered

src/clients/authorization/Scopes.tsx
jonkoops: This has not been resolved yet.

jonkoops (approved):
One comment, otherwise LGTM

src/clients/authorization/Scopes.tsx
edewit: that is right as the PR for changing the types is still pending, I'll change it in another PR

PR #1820, opened 2022-01-03
Update Masthead tests

Aboullos:
## Motivation
Updating Masthead tests in order to make it more clear and avoid a possible failure scenario of  the`isMobileMode` method

## Brief Description
- Change the name of the method `isAdminConsole`, that could imply the return of a boolean, to `checkIsAdminConsole` to imply the assert check.
- Modify `isMobileMode` method to check the window size instead of just returning a variable (there are different scenarios where the mobile mode could change without updating the variable)

## Checklist:

### DIFF
--- a/cypress/support/pages/admin_console/Masthead.ts
+++ b/cypress/support/pages/admin_console/Masthead.ts
@@ -5,9 +5,8 @@ export default class Masthead {
 
   private userDrpDwn = "#user-dropdown";
   private userDrpDwnKebab = "#user-dropdown-kebab";
-  private isMobile = false;
 
-  isAdminConsole() {
+  checkIsAdminConsole() {
     cy.get(this.logoBtn).should("exist");
     cy.get(this.userDrpDwn).should("exist");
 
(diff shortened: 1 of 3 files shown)


cypress/support/pages/admin_console/Masthead.ts
```diff
@@ -7,19 +7,24 @@ export default class Masthead {
...
 
   get isMobileMode() {
+    if (window.parent[0].innerWidth >= 1024) {
+      this.isMobile = false;
+    } else {
+      this.isMobile = true;
+    }
+
     return this.isMobile;
   }
```
jonkoops, line 25: If this is always the logic to determine if we are mobile or not we can get rid of the `isMobile` field entirely and return the result of the statement instead.

```suggestion
  get isMobileMode() {
    return window.parent[0].innerWidth >= 1024;
  }
```

jonkoops (changes requested):

cypress/support/pages/admin_console/Masthead.ts
Aboullos: Updated

jonkoops (approved):
LGTM

PR #1822, opened 2022-01-04
Add overview for attributes groups to realm settings

jonkoops:
Adds the overview for the attributes groups to the realm settings (as per [design](https://marvelapp.com/prototype/5eed532/screen/82246800)). Does not include the form to create a new attributes group to keep review size small.

To test, follow these steps:

1. Go to realm settings and enable the 'User Profile Enabled' switch
2. Save the realm settings
3. Use the old UI to create an attributes group (Realm Settings :arrow_right: User Profile :arrow_right: Attribute Groups)
4. Visit the same page in the new UI and verify it can be deleted from the context menu.

### DIFF
--- a/cypress/integration/user_profile_tab.spec.ts
+++ b/cypress/integration/user_profile_tab.spec.ts
@@ -1,0 +1,54 @@
+import ListingPage from "../support/pages/admin_console/ListingPage";
+import RealmSettingsPage from "../support/pages/admin_console/manage/realm_settings/RealmSettingsPage";
+import SidebarPage from "../support/pages/admin_console/SidebarPage";
+import LoginPage from "../support/pages/LoginPage";
+import AdminClient from "../support/util/AdminClient";
+import { keycloakBefore } from "../support/util/keycloak_before";
+import ModalUtils from "../support/util/ModalUtils";
+
+const loginPage = new LoginPage();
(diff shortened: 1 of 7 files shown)


jonkoops:
Tests have not been added yet, but I will add them in a subsequent commit.

cypress/integration/user_profile_tab.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach(() => {
+    keycloakBefore();
+    loginPage.logIn();
+    sidebarPage.goToRealm(realmName);
+    sidebarPage.goToRealmSettings();
+  });
+
+  describe("Attribute groups", () => {
+    it("deletes an attributes group", () => {
+      cy.wrap(null).then(() =>
```
edewit, line 41: Nice already doing it new style

edewit (approved):
Looks great!

cypress/integration/user_profile_tab.spec.ts
jonkoops: Yeah, this was a lifesaver! Most of the test is now just the login step before each case, which I am sure we can get rid of at some point.

edewit: yep see #1828

PR #1828, opened 2022-01-05
now only logs in per test suite for client tests

edewit:
future PR will change other integration tests to do the same


jonkoops (approved):

mfrances17 (approved):
LGTM - this is great work! so much smarter and more streamlined. we should see a measurable performance increase when it is applied throughout the entire test suite.

PR #1835, opened 2022-01-07
Adds the permissions tab to the Authorization screen under clients

edewit:
Permissions tab with create for resource based permssions scope-based comming in a sperate PR also search is still a work in progress

### DIFF
--- a/src/clients/authorization/permissions.css
+++ b/src/clients/authorization/permissions.css
@@ -1,0 +1,5 @@
+.pf-c-button.disabled {
+  color: var(--pf-c-button--disabled--Color);
+  background-color: var(--pf-c-button--disabled--BackgroundColor);
+  --pf-c-button--after--BorderWidth: 0;
+}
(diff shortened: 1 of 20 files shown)


src/clients/authorization/permissions.css
```diff
@@ -0,0 +1,5 @@
+.pf-c-button.disabled {
```
mfrances17, line 1: is this really needed? it should probably be renamed to make it unique since styles are global.

mfrances17 (changes requested):
Functionality works great... nice job. 

Added one review comment regarding a style, and also I would expect the associated policy row to be indented under the permission, since it is a drop-down, as it is in the old console. Currently it's outdented and doesn't look quite right:
indent

src/clients/authorization/permissions.css
edewit: yes I can't make the buttons actually disabled as that would make the tooltip not work, but your right I should make it a unique name what was I thinking

mfrances17 (approved):
Both issues were addressed, nice work

PR #1839, opened 2022-01-07
Realm roles: fixes bug that prevents fields from being populated

### LINKED ISSUE #1838: Realm roles: role name is cleared when attributes are added
### Describe the bug

There are a few bugs in the realm roles section:

- When a new realm role is created, it takes you to the "Details" tab -- the name and description in this form do not get populated with the data for the new realm role.
- When an attribute is added to the a realm role, the `name` attribute is set to `""` at `save`. If you navigate back to the roles list, the link is no longer clickable because the realm name has been replaced with an empty string.

### Version

16.1

### Expected behavior

Details tab should be populated with the field values from the server, role name should not be cleared on attributes save

### How to Reproduce?

See above

### Anything else?

Was working on updates to the `AttributesInput` component in another issue, and discovered these bugs. Investigating.

jenny-s51:
Closes #1838 

Finishing up a PR that will add selectable functionality to [AttributeInput](https://github.com/keycloak/keycloak-admin-ui/blob/0bbd4ddad1b0fe3eeac9942b88539933f2e66c3a/src/components/attribute-input/AttributeInput.tsx) -- however while testing, I discovered the data could not be saved correctly due to this bug.

After a bunch of investigating, this seems like it's a timing issue.

### DIFF
--- a/src/realm-roles/RealmRoleTabs.tsx
+++ b/src/realm-roles/RealmRoleTabs.tsx
@@ -90,21 +90,19 @@ export default function RealmRoleTabs() {
         throw new Error(t("common:notFound"));
       }
 
+      setRealm(realm);
+
       if (role) {
         const convertedRole = convert(role);
         setRole(convertedRole);
-        setTimeout(() => {
-          Object.entries(convertedRole).map((entry) => {
-            setValue(entry[0], entry[1]);
-          });
-        }, 5);
+        Object.entries(convertedRole).map((entry) => {
+          setValue(entry[0], entry[1]);
+        });
 
         if (url.includes("/attributes")) {
           setValue("attributes", convertedRole.attributes);
         }
       }
-
-      setRealm(realm);
     },
     [key, url]
   );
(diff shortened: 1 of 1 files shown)


src/realm-roles/RealmRoleTabs.tsx
```diff
@@ -93,14 +93,20 @@ export default function RealmRoleTabs() {
...
-        Object.entries(convertedRole).map((entry) => {
-          setValue(entry[0], entry[1]);
-        });
+        setTimeout(() => {
+          Object.entries(convertedRole).map((entry) => {
+            setValue(entry[0], entry[1]);
+          });
+        }, 5);
+
+        if (url.includes("/attributes")) {
```
edewit, line 102: This seem like a workaround also the timout of 5 milliseconds I don't see why that would be needed

edewit (changes requested):

src/realm-roles/RealmRoleTabs.tsx
jenny-s51: Ah you're right, reordered the hook and the conditional and this seemed to resolve it.

jonkoops (approved):

edewit (approved):

PR #1840, opened 2022-01-07
Adds selectable attribute functionality

jenny-s51:
## Motivation

Selectable attribute functionality is necessary for the Clients -> Evaluate screens. This PR allows the user to set a prop on `AttributeForm` called `isKeySelectable`, which would render a dropdown for the key of attribute pairs. 
Towards https://marvelapp.com/prototype/5eg9ic7/screen/83188830

## Verification Steps

1. Go to `RealmRoleTabs.tsx`.
2. Find the `AttributesForm` component on line 385.
3. Add the prop `isKeySelectable`
4. Add the prop `selectableItems={["a", "b", "c"]}`
5. Go to Realm Roles.
6. Create a new realm role, save it, then go to the "Attributes" tab.
4. Verify that for all attribute keys, you are seeing a Select dropdown with the values you entered.
5. Add/save some attributes and verify that the data is saved and shown correctly.


src/clients/utils.ts
```diff
@@ -19,3 +19,49 @@ export const getProtocolName = (t: TFunction<"clients">, protocol: string) => {
...
+
+export const defaultContextAttributes = [
+  {
+    key: "custom",
+    name: "Custom Attribute...",
+    custom: true,
+  },
+  {
+    key: "kc.identity.authc.method",
+    name: "Authentication Method",
```
edewit, line 31: These texts need to come from a message bundle

src/components/attribute-input/AttributeInput.tsx
```diff
@@ -55,22 +97,110 @@ export const AttributeInput = ({ name }: AttributeInputProps) => {
...
             </Td>
             <Td>
-              <TextInput
-                id={`${attribute.id}-value`}
-                name={`${name}[${rowIndex}].value`}
-                ref={register()}
-                defaultValue={attribute.value}
-                data-testid="attribute-value-input"
-              />
+              {getAttributeValues(rowIndex)?.length ? (
```
edewit, line 161: I think it would have been nicer if this dropdown version was separate from the other attribute inputs as it is used in only one place

src/realm-roles/RealmRoleTabs.tsx
```diff
@@ -93,14 +93,20 @@ export default function RealmRoleTabs() {
       if (role) {
         const convertedRole = convert(role);
         setRole(convertedRole);
-        Object.entries(convertedRole).map((entry) => {
-          setValue(entry[0], entry[1]);
-        });
+        setTimeout(() => {
```
edewit, line 96: can't be that this is needed

src/components/attribute-input/AttributeInput.tsx
jenny-s51: Makes sense, reverted back to the separated dropdown

src/clients/utils.ts
jenny-s51: Working on this

edewit (approved):

PR #1823, opened 2022-01-04
Realm settings: Return user to Policies subtab 

jenny-s51:
## Motivation

@_**Haley Wang|246396** [said](https://keycloak.zulipchat.com/#narrow/pm-with/246396,327337-pm/near/265912135):
> Can the cancel here navigate back to the policies list instead of the profiles list?

Screen Shot 2022-01-04 at 10 57 41 AM

This has been a repeated subject of discussion, as the current implementation of tabs does not support bookmarkable subtabs. @ssilvert recently brought this up while reviewing the code for client policies, and @yih-wang also requested a change here.

## Brief Description

The PR includes a workaround for navigating back to the "Client Policies" subtab.

## Verification Steps

1. Go to Client Policies -> Policies subtab.
2. Click "Create client policy"
3. Click the "Cancel" button.
4. Verify that the UI takes you to the client policies list, rather than the client profiles list.
5. Repeat steps 1-2 and add a new client policy.
6. Go back to the policies list, and click on the new policy.
7. Click the "Client Policies" link in the breadcrumb, and verify that the UI takes you back to the client policies list.

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] User-visible strings are using the react-i18next framework (useTranslation)
- [x] Help has been implemented
- [x] axe report has been run and resulting a11y issues have been resolved
- [x] Unit tests have been created/updated

PR #1852, opened 2022-01-11
Add RoutableTabs component for routed tabs

jonkoops:
## Motivation
Introduces a new `RoutableTabs` component that allows (nested) tabs to be routed by means of the path of a path. This component is intended as a replacement for `KeycloakTabs`, which requires too much configuration and does not support nested tabs properly.

This is required in order to facilitate returning to nested tabs after filling in a form related to that tab (for example #1823).

### DIFF
--- a/cypress/integration/realm_settings_events_test.spec.ts
+++ b/cypress/integration/realm_settings_events_test.spec.ts
@@ -197,7 +197,7 @@ describe("Realm settings events tab tests", () => {
 
   it("Realm header settings", () => {
     sidebarPage.goToRealmSettings();
-    cy.get("#pf-tab-securityDefences-securityDefences").click();
+    cy.findByTestId("rs-security-defenses-tab").click();
     cy.findByTestId("headers-form-tab-save").should("be.disabled");
     cy.get("#xFrameOptions").clear().type("DENY");
     cy.findByTestId("headers-form-tab-save").should("be.enabled").click();
@@ -207,7 +207,7 @@ describe("Realm settings events tab tests", () => {
 
   it("Brute force detection", () => {
     sidebarPage.goToRealmSettings();
-    cy.get("#pf-tab-securityDefences-securityDefences").click();
+    cy.findAllByTestId("rs-security-defenses-tab").click();
     cy.get("#pf-tab-20-bruteForce").click();
 
     cy.findByTestId("brute-force-tab-save").should("be.disabled");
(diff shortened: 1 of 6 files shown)


edewit (approved):
:+1: best we can do with the limits set by PF

jenny-s51 (changes requested):
Great collaboration on this exciting update @jonkoops @edewit ! 

When I go to Realm Settings, I am unable to select any of the realm settings tabs at the moment -- the "General" tab always remains active when I click on a different tab.

jonkoops:
@jenny-s51 I think you might be running an older version of PatternFly, can you run `npm install` and restart your dev server?

jenny-s51:
Thanks @jonkoops ! The tabs are looking good and working as expected in `RealmSettings`. 

To test the nesting, I replaced the `Tabs` component that wrapped the Client Policies subtabs with `RoutableTabs`, then I added an optional `tab` param to `clientPoliciesParams` in `clientPoliciesRoute`. Then I passed `routableTab` into the `profiles` tab like this: 
```
                  {...routableTab({
                    to: toClientPolicies({
                      realm: realmName,
                      tab: "profiles",
                    }),
                    history,
                  })}
```

Is this how the component is expected to be used?

This is the result I'm getting, though we would expect that the tabs above should still be visible/selectable. How did you guys check that the nesting works as expected? I'd like to try the same on my end 🙂 

Screen Shot 2022-01-11 at 3 27 54 PM

jonkoops:
@jenny-s51 We did some isolated testing on a new route to test out the nesting. Thinking about this specific overview, we might have to create more specific routes to ensure we're not accidentally rendering other routes. I'll see if I can get an implementation working.

jonkoops:
@jenny-s51 I've pushed some work to make sure nested tabs remain active, looks like my initial code was not properly optimized for it. 

@edewit could you also review this new code?

edewit (approved):
:+1:

jenny-s51 (approved):
Looking good @jonkoops ! I added a `subTab` attribute to the `toRealmSettings` route, e.g. 
```
             {...routableTab({
                to: toRealmSettings({
                  realm: realmName,
                  tab: "clientPolicies",
                  subTab: "policies",
                }),
                history,
              })}
```

Thank you for looking into the nested tabs issue - the test case I used initially was still not showing the tabs, but when done this way, the behavior works as expected. 👍

jonkoops:
Nice, I am happy to hear it's working. I still see a couple of rough edges in my own testing with nested tabs, but I think they could be resolved. I'll merge this PR after the tests pass and then take another look.

PR #1858, opened 2022-01-12
Fixes localization 500 error

### LINKED ISSUE #1857: Enable Internationalization causes crash
### Describe the bug

1. Go to Realm settings --> Localization
2. Try to enable Internationalization
3. Observe the crash

### Version

17.0.0-SNAPSHOT

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

jenny-s51:
Closes #1857 

Previously there was a `try/catch` block in the loader that prevented this error. The loader was recently refactored to a `useFetch` in #1663 to allow for message bundle functionality.

### DIFF
--- a/src/realm-settings/LocalizationTab.tsx
+++ b/src/realm-settings/LocalizationTab.tsx
@@ -409,17 +409,20 @@ export const LocalizationTab = ({
                       isOpen={supportedLocalesOpen}
                       placeholderText={t("selectLocales")}
                     >
-                      {themeTypes.login![0].locales.map(
-                        (locale: string, idx: number) => (
-                          <SelectOption
-                            selected={value.includes(locale)}
-                            key={`locale-${idx}`}
-                            value={locale}
-                          >
-                            {t(`allSupportedLocales.${locale}`)}
-                          </SelectOption>
+                      {
+                        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+                        themeTypes.login![0].locales?.map(
+                          (locale: string, idx: number) => (
+                            <SelectOption
+                              selected={value.includes(locale)}
+                              key={`locale-${idx}`}
+                              value={locale}
+                            >
+                              {t(`allSupportedLocales.${locale}`)}
+                            </SelectOption>
+                          )
                         )
-                      )}
+                      }
                     </Select>
                   )}
                 />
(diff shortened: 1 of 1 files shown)


src/realm-settings/LocalizationTab.tsx
```diff
@@ -115,13 +115,17 @@ export const LocalizationTab = ({
...
+        .getRealmLocalizationTexts({
+          first,
+          max,
+          realm: realm.realm!,
+          selectedLocale:
+            selectMenuLocale ||
+            getValues("defaultLocale") ||
+            whoAmI.getLocale(),
+        })
+        .catch(() => []);
```
ssilvert, line 128: You are swallowing the error here.  Can you just avoid the error instead?

ssilvert (changes requested):
@jenny-s51 This does not fix the problem I see.  I just click on the Internationalization switch and see the crash.  No request is ever made to the server.

jenny-s51:
@ssilvert I'm no longer seeing a crash and am able to toggle/save the internationalization switch!

Erik implemented a similar change a few months ago to address this issue with the same kind of behavior https://github.com/keycloak/keycloak-admin-ui/commit/6f29839b0620cf075a3e4cabf45b48cac818417b#diff-6a013ef3899e308f424b7fd2f79cee59df8f9f2989832ece8ccf14d4081aa2ab

@edewit was there a reason that it needed to be implemented like this?

ssilvert:
@jenny-s51 I figured out why I'm seeing a crash and you are not.  A bug has been introduced into Keycloak that removes all locales if you set an alternate themes directory in standalone.xml (which is what I always do).

There is still a bug in your code, but it turns out to be an edge case.  You can simulate the Keycloak bug I am seeing by telling Keycloak that you aren't using any locales.

Go into themes/base on your keycloak server.  For each theme type (account, admin, email, and login), comment out the locales in theme.properties.  Properties files use the `#` symbol for comments, so it should look like this:

`# locales=ca,cs,da,de,en,es,fr,hu,it,ja,lt,nl,no,pl,pt-BR,ru,sk,sv,tr,zh-CN`

Then restart your server and you will see the crash.  The problem is that the server will tell you there are no locales, which is actually a valid state.  But your code is not handling it.

jenny-s51:
@ssilvert thank you for your helpful repro steps! Ah ok, yes I'm seeing an `undefined` error since there aren't any locales set. Investigating...

ssilvert (approved):
Looks good.

PR #1860, opened 2022-01-13
LDAP mapper test fix

mfrances17:
## Motivation
Now that we are getting more consistent cypress test suite passes, the LDAP mapper tests stand out as being still a bit flakey and seem to sporadically cause an otherwise passing test suite to fail. 

## Brief Description
After watching the vids of the failing ldap mapper tests, the issue is that when the final mapper is deleted, the timing causes the modal confirm button to be disconnected from the DOM. This is a common issue with cypress and although there are a few community-proposed workarounds, there is no cypress officially supported solution. In my testing, adding a force to the modal button click seems like the easiest resolution to the issue. 

To test these changes, I ran the mapper tests locally (UI and headless) 10 times, and then ran the mapper tests on CI 5 times, and then the full suite on CI an additional 3 times. I could not get either of the mapper tests to fail with these changes.

## Verification Steps
The mapper tests now consistently pass on CI so no verification should be needed, but you can re-run them on CI and/or run the tests locally to verify if you choose.

## Checklist:

## Additional Notes
None

### DIFF
--- a/cypress/support/util/ModalUtils.ts
+++ b/cypress/support/util/ModalUtils.ts
@@ -6,8 +6,8 @@ export default class ModalUtils {
   private cancelModalBtn = "#modal-cancel";
   private closeModalBtn = ".pf-c-modal-box .pf-m-plain";
 
-  confirmModal() {
-    cy.get(this.confirmModalBtn).click();
+  confirmModal(force: boolean = false) {
+    cy.get(this.confirmModalBtn).click({ force: force });
 
     return this;
   }
--- a/cypress/support/util/ModalUtils.ts
+++ b/cypress/support/util/ModalUtils.ts
@@ -6,8 +6,8 @@ export default class ModalUtils {
   private cancelModalBtn = "#modal-cancel";
   private closeModalBtn = ".pf-c-modal-box .pf-m-plain";
 
-  confirmModal(force: boolean = false) {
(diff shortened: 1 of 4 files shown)


edewit (approved):

cypress/support/util/ModalUtils.ts
```diff
@@ -6,8 +6,8 @@ export default class ModalUtils {
   private cancelModalBtn = "#modal-cancel";
   private closeModalBtn = ".pf-c-modal-box .pf-m-plain";
 
-  confirmModal() {
-    cy.get(this.confirmModalBtn).click();
+  confirmModal(force: boolean = false) {
```
jonkoops, line 9: If you provide a default argument TypeScript can infer the type from the value. This means you can omit the type definition.

```suggestion
  confirmModal(force = false) {
```

cypress/support/util/ModalUtils.ts
```diff
@@ -6,8 +6,8 @@ export default class ModalUtils {
...
+    cy.get(this.confirmModalBtn).click({ force: force });
```
jonkoops, line 10: When you pass a value with the same name as the key you can omit the value.
```suggestion
    cy.get(this.confirmModalBtn).click({ force });
```

jonkoops (approved):
Awesome, this test has been very unreliable indeed. Thanks for the fix @mfrances17!

PR #1862, opened 2022-01-14
Make 'User Profile' tabs routable

jonkoops:
This will allow us to return to the correct path when navigating back from a nested form.

### DIFF
--- a/src/components/routable-tabs/RoutableTabs.tsx
+++ b/src/components/routable-tabs/RoutableTabs.tsx
@@ -13,17 +13,23 @@ import React, {
 } from "react";
 import { useLocation } from "react-router-dom";
 
+// TODO: Remove the custom 'children' props and type once the following issue has been resolved:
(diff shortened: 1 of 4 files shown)


edewit (approved):
`defaultLocation` could also be the first tab, right?

jonkoops:
> `defaultLocation` could also be the first tab, right?

It could be any location you'd like really.

jenny-s51 (approved):
Tabs are looking great @jonkoops ! The `defaultLocation` prop is an excellent addition. Just wanted to confirm since I'm seeing a blank screen on the `Attributes` tab ... is the `Attributes` tab still a WIP?

jonkoops:
@jenny-s51 That is correct, the 'Attributes' tab still needs to be implemented.

jonkoops:
I'm merging this PR even with some of the caveats it has, I've created a separate bug to resolve these issues (#1892).

PR #1884, opened 2022-01-18
Added the policy screens

edewit:
implementation of the design [marvel](https://marvelapp.com/prototype/2c39260g/screen/83080815) still missing the search / filter that
will come in a later PR, but it has all the input screens for the different policy types


src/clients/authorization/policy/Role.tsx
```diff
@@ -0,0 +1,164 @@
...
+            </Button>
+          </>
+        )}
+      />
+      {selectedRoles.length > 0 && (
+        <TableComposable>
+          <Thead>
+            <Tr>
+              <Th>{t("roles")}</Th>
+              <Th>{t("common:required")}</Th>
```
jenny-s51, line 114: The column title here should just be "Required" but it becomes `Required Fields` using this message bundle key, which causes it to wrap like this:

Screen Shot 2022-01-19 at 3 17 07 PM

src/clients/authorization/policy/PolicyDetails.tsx
```diff
@@ -0,0 +1,229 @@
...
+  client: Client,
+  "client-scope": ClientScope,
+  group: Group,
+  regex: Regex,
+  role: Role,
+  time: Time,
+  js: JavaScript,
+} as const;
+
+const isValidComponentType = (value: string): boolean => value in COMPONENTS;
```
jenny-s51, line 58: Currently if I try to make a policy of type `user`, an error gets thrown: `Danger alert:Not a supported user!` since `user` doesn't exist within this components object. 

According to the design `user` and `client` policy types follow the same pattern -- maybe the `Client` component can be reused here

src/clients/authorization/policy/Role.tsx
```diff
@@ -0,0 +1,164 @@
...
+    <FormGroup
+      label={t("roles")}
+      labelIcon={
+        <HelpItem
+          helpText="clients-help:policyRoles"
+          fieldLabelId="clients:roles"
+        />
+      }
+      fieldId="roles"
+      helperTextInvalid={t("requiredRoles")}
```
jenny-s51, line 65: Looks like this needs a corresponding message bundle value! e.g. "Please add at least one role."

jenny-s51 (changes requested):
Awesome work on these policy type screens @edewit! 💪 Left some comments

edewit:
Thanks @jenny-s51 for the thorough review

jenny-s51 (approved):

PR #1890, opened 2022-01-18
added permission search

edewit:
added the search bar of the authorization search permissions


src/clients/authorization/Permissions.tsx
```diff
@@ -166,10 +170,12 @@ export const AuthorizationPermissions = ({ clientId }: PermissionsProps) => {
     return <KeycloakSpinner />;
   }
 
+  const noData = permissions.length === 0;
+  const searching = !isEqual(search, {});
```
jonkoops, line 174: I think you could accomplish the same without using Lodash by doing the following:
```suggestion
  const searching = Object.keys(search).length !== 0;
```

src/clients/authorization/SearchDropdown.tsx
```diff
@@ -69,24 +108,46 @@ export const SearchDropdown = ({ types }: SearchDropdownProps) => {
...
-                {types?.map((type) => (
-                  <SelectOption
-                    selected={type.type === value.type}
-                    key={type.type}
-                    value={type}
-                  >
-                    {type.name}
-                  </SelectOption>
-                ))}
+                {[
```
jonkoops, line 116: This is perhaps a place you could use a fragment? Or does that bork the control from PatternFly?

jonkoops (approved):

src/clients/authorization/SearchDropdown.tsx
edewit: No, but we have the rule not to have a single child in a fragment right?

jonkoops: In this case it would not be a single child as there are possibly multiple options mapped below.

jonkoops: Why is this an array of JSX elements in this case? What prevents us from doing:

```jsx
<Select>
  <SelectOption>Default value</SelectOption>
  {otherOptions.map((name) => (<SelectOption>{name}</SelectOption>))
</Select>
```

edewit: Types prevent us from doing that.. seems that patternfly only wants one child

edewit: how about:
```typescript
  const typeOptions = (value: string) => [
    <SelectOption key="empty" value="">
      {t("allTypes")}
    </SelectOption>,
    ...(types || []).map((type) => (
      <SelectOption
        selected={type.type === value}
        key={type.type}
        value={type.type}
      >
        {type.name}
      </SelectOption>
    )),
  ];

// ...

            <Select
                toggleId="type"
                onToggle={toggleType}
                onSelect={(event, value) => {
                  event.stopPropagation();
                  onChange(value);
                  toggleType();
                }}
                selections={value || t("allTypes")}
                variant={SelectVariant.single}
                aria-label={t("common:type")}
                isOpen={typeOpen}
              >
                {typeOptions(value)}
            </Select>
```

Issue #378, opened 2021-02-18
KEYCLOAK-16710


ssilvert:
Make sure this is addressed in new console.
https://issues.redhat.com/browse/KEYCLOAK-16710

karvir26:
When we can expect new GUI admin console available and its version? Any tentative timelines?

jonkoops:
Not at this point.

ssilvert:
@edewit Is this fixed?  If so, please close.

edewit:
no the PR is still open

PR #1916, opened 2022-01-21
handle error code < 500 in error handler

edewit:
also:
fixes: #378

### DIFF
--- a/src/context/RealmsContext.tsx
+++ b/src/context/RealmsContext.tsx
@@ -7,7 +7,7 @@ import React, {
   useMemo,
   useState,
 } from "react";
-import type { AxiosError } from "axios";
+import axios from "axios";
 
 import { RecentUsed } from "../components/realm-selector/recent-used";
 import useRequiredContext from "../utils/useRequiredContext";
--- a/src/context/RealmsContext.tsx
+++ b/src/context/RealmsContext.tsx
@@ -7,6 +7,8 @@ import React, {
   useMemo,
   useState,
 } from "react";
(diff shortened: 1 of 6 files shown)


src/context/RealmsContext.tsx
```diff
@@ -33,7 +35,18 @@ export const RealmsProvider: FunctionComponent = ({ children }) => {
...
+    async () => {
+      try {
+        return await adminClient.realms.find({ briefRepresentation: true });
+      } catch (error) {
+        const e = error as Error | AxiosError;
+        if ("response" in e && e.response && e.response.status < 500) {
+          return [];
+        } else {
+          throw error;
+        }
```
jonkoops, line 47: You can use `axios.isAxiosError()` here to determine the error is indeed an `AxiosError`. You can also omit the `else` clause since the `if` branch terminates with a `return` statement. The `Error` type can also be omitted, as it's fine to re-throw a value of the type `unknown`.

```suggestion
        if (
          axios.isAxiosError(error) &&
          error.response &&
          error.response.status < 500
        ) {
          return [];
        }

        throw error;
```

jonkoops (changes requested):

jonkoops (approved):
LGMT

jonkoops:
@edewit Looks like there are still some test failures.

jonkoops:
@edewit looks like there is one more test failure in the masthead, but otherwise LGTM.

mfrances17 (approved):
LGTM

PR #1937, opened 2022-01-24
Clients(Authentication): Adds "Export" tab

jenny-s51:
## Motivation

https://marvelapp.com/prototype/78j1b69/screen/83085244

## Brief Description

Adds "Export" subtab to Authentication settings in Clients

## Verification Steps

1. Go to Clients
2. Create a new OIDC client and make sure "Client authentication" and "Authorization" switches are both enabled
3. Save the client and go to the Authorization -> Export subtab
4. Verify that the "Download" and "Copy" buttons work as expected

## Checklist:


src/clients/ClientDetails.tsx
```diff
@@ -600,6 +601,14 @@ export default function ClientDetails() {
                       reset={() => setupForm(client)}
                     />
                   </Tab>
+                  <Tab
+                    id="Export"
```
jonkoops, line 605: Any reason this `id` is starting with a capital? Never seen that before.
```suggestion
                    id="export"
```

src/clients/authorization/AuthorizationExport.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useParams } from "react-router-dom";
+import FileSaver from "file-saver";
+import { prettyPrintJSON } from "../../util";
+import "./authorization-details.css";
+import { useAlerts } from "../../components/alert/Alerts";
+
+export const AuthorizationExport = () => {
+  const { t } = useTranslation("clients");
+  const adminClient = useAdminClient();
+  const { clientId } = useParams<{ clientId: string }>();
```
jonkoops, line 24: Instead of providing a new type, is there perhaps an existing type that can be used for the route params?

src/clients/authorization/AuthorizationExport.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { addAlert, addError } = useAlerts();
+
+  const [code, setCode] = useState<string>();
+  const [authorizationDetails, setAuthorizationDetails] = useState("");
+
+  useFetch(
+    async () =>
```
jonkoops, line 31: The function already returns a Promise, so there is no need to mark this as `async`.
```suggestion
    () =>
```

src/clients/authorization/AuthorizationExport.tsx
```diff
@@ -0,0 +1,99 @@
...
+      adminClient.clients.exportResource({
```
jonkoops, line 32: This method in the admin client is missing a proper return type, can you add one there?

src/clients/authorization/AuthorizationExport.tsx
```diff
@@ -0,0 +1,99 @@
...
+        id: clientId,
+      }),
+
+    (authDetails) => {
+      setCode(JSON.stringify(authDetails, null, 2));
+      setAuthorizationDetails(authDetails);
```
jonkoops, line 38: Looking at your code `authorizationDetails` is already a string, so why does it need to be stringified again in `code`?

src/clients/authorization/AuthorizationExport.tsx
```diff
@@ -0,0 +1,99 @@
...
+            data-testid="authorization-export-download"
+            onClick={() => exportAuthDetails()}
+          >
+            {t("common:download")}
+          </Button>
+          <Button
+            data-testid="authorization-export-copy"
+            variant="secondary"
+            onClick={() => {
+              navigator.clipboard.writeText(code!);
```
jonkoops, line 89: According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) this method returns a Promise. Make sure that you await it's result and also handle any errors.

jonkoops (changes requested):

src/clients/authorization/AuthorizationExport.tsx
jenny-s51: It looks like `authorizationDetails` should be of type `resourceServerRepresentation`, therefore the stringify is needed here! Updated the default hook value to use the correct type.

src/clients/ClientDetails.tsx
jenny-s51: Nice catch, thanks - fixed!

jonkoops (approved):
Changes are looking good to me, @jenny-s51 could you install the latest Admin Client to see if the code is compatible with the types you added in https://github.com/keycloak/keycloak-nodejs-admin-client/pull/408?

jonkoops:
Looks like some tests are failing, but I don't think they are related to the changes in this PR. Should we merge this?

PR #1954, opened 2022-01-25
make all sub tabs routeable tabs

edewit:
closes: #1507, closes: #1204

### DIFF
--- a/src/realm-settings/ClientProfileForm.tsx
+++ b/src/realm-settings/ClientProfileForm.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { Fragment, useEffect, useMemo, useState } from "react";
 import {
   ActionGroup,
   AlertVariant,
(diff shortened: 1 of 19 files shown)


jenny-s51 (changes requested):
Subtab/route cleanup is looking good @edewit ! 

A couple of small things: 

In `PageNav.tsx`, the `path` for the `RealmSettings` component is set to `/realm-settings` without a value for tab. If you click on "Realm Settings" in the left nav, it renders a "not found" page because there is no tab set by default.

For the "Client Policies" breadcrumb link, I am seeing that the policy and/or profile name still remains in the URL once clicked, though the user is returned to the correct table view, i.e.
Screen Shot 2022-01-26 at 11 44 41 AM

Otherwise realm settings tabs look great! No longer seeing the console error 👍

edewit:
@jenny-s51 ahh good find fixed (#30e7659) that first point (the no tab route). As for the second I don't think I can fix that as there is an extra level `#/master/realm-settings/clientPolicies/fapi-1-baseline/edit-profile` goes to edit of `fapi-1-baseline` and the breadcrumb will just remove one level in this case `edit-profile` maybe we can change the url not to have this `edit-profile` part? Otherwise I don't think it's a huge problem as it does navigate to the right page.

jenny-s51 (approved):

PR #1861, opened 2022-01-13
Clients: Authorization -> Evaluate tab

jenny-s51:
## Motivation

https://marvelapp.com/prototype/5eg9ic7/screen/82302108

## Brief Description

Adds fields/selectable functionality to Authentication -> Evaluate tab

## Verification Steps

1. Go to Clients 
2. Create a new OIDC client and make sure "Client authentication" and "Authorization" switches are both enabled
3. Save the client and go to the Authorization -> Evaluate subtab
4. Verify that the fields are correct and work/populate as expected 

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] User-visible strings are using the react-i18next framework (useTranslation)
- [x] Help has been implemented
- [x] axe report has been run and resulting a11y issues have been resolved
- [x] Unit tests have been created/updated

## Additional Notes

Evaluate call and action group methods will be done in a separate PR

PR #1961, opened 2022-01-26
Clients(Authorization): Evaluate tab updates

jenny-s51:
## Motivation

Extends work done on #1861 - these fixes are necessary in order to implement results screen https://marvelapp.com/prototype/5eg9ic7/screen/83188838

## Brief Description

Fixes `evaluate` call which now POSTs successfully, adds 'revert' functionality to form, fixes bug in realm role attributes and  resolves some issues with data types in `AttributeInput`

## Verification Steps

1. Create a OIDC client with Client Authentication and Authorization switches enabled.
2. Go to the new client -> Authorization -> Evaluate
3. Fill in form data and verify that the `evaluate` call POSTs successfully

## Checklist:

### DIFF
--- a/src/clients/authorization/AuthorizationEvaluate.tsx
+++ b/src/clients/authorization/AuthorizationEvaluate.tsx
@@ -28,6 +28,18 @@ import { defaultContextAttributes } from "../utils";
 import type ResourceRepresentation from "@keycloak/keycloak-admin-client/lib/defs/resourceRepresentation";
 import { useParams } from "react-router-dom";
 import type ScopeRepresentation from "@keycloak/keycloak-admin-client/lib/defs/scopeRepresentation";
+import type { KeyValueType } from "../../components/attribute-form/attribute-convert";
+
+interface EvaluateFormInputs
+  extends Omit<ResourceEvaluation, "context" | "resources"> {
+  applyToResource: boolean;
+  alias: string;
+  authScopes: string[];
+  context: {
+    attributes: Record<string, string>[];
+  };
+  resources: Record<string, string>[];
+}
 
 export type AttributeType = {
   key: string;
@@ -42,9 +54,18 @@ type ClientSettingsProps = {
   clients: ClientRepresentation[];
   clientName?: string;
   save: () => void;
-  reset: () => void;
   users: UserRepresentation[];
   clientRoles: RoleRepresentation[];
+};
+
+export type AttributeForm = Omit<
+  EvaluateFormInputs,
+  "context" | "resources"
+> & {
+  context: {
+    attributes?: KeyValueType[];
+  };
+  resources?: KeyValueType[];
 };
 
 export const AuthorizationEvaluate = ({
@@ -52,10 +73,9 @@ export const AuthorizationEvaluate = ({
   clientRoles,
   clientName,
   users,
-  reset,
 }: ClientSettingsProps) => {
-  const form = useFormContext<ResourceEvaluation>();
-  const { control } = form;
+  const form = useFormContext<EvaluateFormInputs>();
+  const { control, reset, trigger } = form;
   const { t } = useTranslation("clients");
   const adminClient = useAdminClient();
   const realm = useRealm();
@@ -90,15 +110,27 @@ export const AuthorizationEvaluate = ({
     []
   );
 
-  const evaluate = (formValues: ResourceEvaluation) => {
+  const evaluate = async () => {
+    if (!(await trigger())) {
+      return;
+    }
+    const formValues = form.getValues();
+    const keys = formValues.resources.map(({ key }) => key);
     const resEval: ResourceEvaluation = {
       roleIds: formValues.roleIds ?? [],
+      clientId: selectedClient ? selectedClient.id! : clientId,
       userId: selectedUser?.id!,
+      resources: resources.filter((resource) => keys.includes(resource.name!)),
       entitlements: false,
-      context: formValues.context,
-      resources: formValues.resources,
-      clientId: selectedClient?.id!,
+      context: {
+        attributes: Object.fromEntries(
+          formValues.context.attributes
+            .filter((item) => item.key || item.value !== "")
+            .map(({ key, value }) => [key, value])
+        ),
+      },
     };
+
     return adminClient.clients.evaluateResource(
       { id: clientId!, realm: realm.realm },
       resEval
@@ -128,7 +160,10 @@ export const AuthorizationEvaluate = ({
             fieldId="client"
           >
             <Controller
-              name="client"
+              name="clientId"
+              rules={{
+                validate: (value) => value.length > 0,
+              }}
               defaultValue={clientName}
               control={control}
               render={({ onChange, value }) => (
@@ -140,7 +175,7 @@ export const AuthorizationEvaluate = ({
                     onChange((value as ClientRepresentation).clientId);
                     setClientsDropdownOpen(false);
                   }}
-                  selections={value}
+                  selections={selectedClient === value ? value : clientName}
                   variant={SelectVariant.typeahead}
                   aria-label={t("client")}
                   isOpen={clientsDropdownOpen}
@@ -171,6 +206,9 @@ export const AuthorizationEvaluate = ({
           >
             <Controller
               name="userId"
+              rules={{
+                validate: (value) => value.length > 0,
+              }}
               defaultValue=""
               control={control}
               render={({ onChange, value }) => (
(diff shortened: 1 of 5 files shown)


src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -296,11 +328,23 @@ export const AuthorizationEvaluate = ({
               helperTextInvalid={t("common:required")}
               fieldId={name!}
             >
-              <AttributeInput
-                selectableValues={resources.map((item) => item.name!)}
-                resources={resources}
-                isKeySelectable
-                name="resources"
+              <Controller
```
edewit, line 331: Why would you need a `<Controller` around this control, isn't it already part of the "form"

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -28,6 +28,16 @@ import { defaultContextAttributes } from "../utils";
...
 import type ScopeRepresentation from "@keycloak/keycloak-admin-client/lib/defs/scopeRepresentation";
+import type { KeyValueType } from "../../components/attribute-form/attribute-convert";
+
+interface EvaluateFormInputs
+  extends Omit<ResourceEvaluation, "context" | "resources"> {
+  applyToResource: boolean;
+  alias: string;
+  authScopes: string[];
+  context: { [key: string]: string };
+  resources: { [key: string]: string }[];
```
jonkoops, line 39: You can use a [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) type here.
```suggestion
  context: Record<string, string>;
  resources: Record<string, string>[];
```

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -90,19 +98,37 @@ export const AuthorizationEvaluate = ({
...
-      context: formValues.context,
-      resources: formValues.resources,
-      clientId: selectedClient?.id!,
-    };
-    return adminClient.clients.evaluateResource(
-      { id: clientId!, realm: realm.realm },
-      resEval
-    );
+  const evaluate = async () => {
+    if (await trigger()) {
```
jonkoops, line 102: You can reduce nesting here by using an [early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8).

```ts
if (!(await trigger())) {
  return;
}

// The rest of the code.
```

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -90,19 +98,37 @@ export const AuthorizationEvaluate = ({
...
+      const formValues = form.getValues();
+      const resEval: ResourceEvaluation = {
+        roleIds: formValues.roleIds ?? [],
+        clientId: selectedClient ? selectedClient.id! : clientId,
+        userId: selectedUser?.id!,
+        resources: resources.filter((resource) =>
+          form
+            .getValues()
+            .resources.map((item) => item.key)
```
jonkoops, line 111: This logic is repeated for every resource you are iterating, it might be a good idea to store it in a variable. You also should not have to call `getValues()` again as there is already a variable that contains this called `formValues` above.

```ts
const keys = formValues.resources.map(({ key }) => key);
```

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -90,19 +98,37 @@ export const AuthorizationEvaluate = ({
...
+        entitlements: false,
+        context: {
+          attributes: formValues.context.length
+            ? {
+                [Object.entries(formValues.context).map(
+                  (item) => (item[1] as unknown as KeyValueType).key
+                ) as unknown as string]: Object.entries(formValues.context).map(
+                  (item) => (item[1] as unknown as KeyValueType).value
+                )[0],
+              }
```
jonkoops, line 123: This code is very hard to parse, but looking at the type casting this seems to entirely change the type of `formValues.context`. Why does this need to happen?

I'd like to see this code without type-casting, as this disables certain compiler checks that can prevent errors during runtime.

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -296,11 +328,23 @@ export const AuthorizationEvaluate = ({
...
+              <Controller
+                name="applyToResource"
+                defaultValue=""
+                control={control}
+                render={() => (
+                  <AttributeInput
+                    selectableValues={
+                      resources
+                        .map((item) => [{ name: item.name }])
+                        .flat() as AttributeType[]
```
jonkoops, line 340: There is no need to call `flat()` here or to type-cast the items if you map it directly to an object instead of an array with a generic type.

```ts
resources.map<AttributeType>((item) => ({
  name: item.name!,
}))
```

According to the `AttributeType` type you will also have to provide a `key` which is missing here. Because this was being type-casted this error was suppressed by the compiler.

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -390,35 +434,34 @@ export const AuthorizationEvaluate = ({
...
+            data-testid="authorization-revert"
+            variant="link"
+            onClick={() => reset()}
+          >
+            {t("common:revert")}
+          </Button>
+          <Button
+            data-testid="authorization-revert"
+            variant="primary"
+            onClick={() => reset}
```
jonkoops, line 459: You probably want to call the `reset()` function here.
```suggestion
            onClick={() => reset()}
```

src/components/attribute-input/AttributeInput.tsx
```diff
@@ -56,7 +58,7 @@ export const AttributeInput = ({
     if (!fields.length) {
       append({ key: "", value: "" });
     }
-  }, []);
+  }, [reset]);
```
jonkoops, line 61: Why does this effect now depend on `reset`? It doesn't look like this dependency is used inside of the effect at all.

jonkoops (changes requested):

src/clients/authorization/AuthorizationEvaluate.tsx
jenny-s51: Good call - thanks @jonkoops

src/clients/authorization/AuthorizationEvaluate.tsx
jenny-s51: Nice catch, I've updated this logic - the submitted types should be correct and casting is no longer necessary

src/clients/authorization/AuthorizationEvaluate.tsx
jenny-s51: Thanks! Removed 👍

src/components/attribute-input/AttributeInput.tsx
jenny-s51: The reset sets `fields.length` to 0 -- if it is not a dependency, the field is removed rather than being set to empty

src/clients/authorization/AuthorizationEvaluate.tsx
jonkoops: Awesome, nice work!

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -90,15 +110,31 @@ export const AuthorizationEvaluate = ({
     []
   );
 
-  const evaluate = (formValues: ResourceEvaluation) => {
+  console.log(form.getValues());
```
jonkoops, line 113: Stray console log statement here.
```suggestion
```

src/clients/authorization/AuthorizationEvaluate.tsx
```diff
@@ -90,15 +110,31 @@ export const AuthorizationEvaluate = ({
...
+      resources: resources.filter((resource) => keys.includes(resource.name!)),
       entitlements: false,
-      context: formValues.context,
-      resources: formValues.resources,
-      clientId: selectedClient?.id!,
+      context: {
+        attributes: Object.fromEntries(
+          formValues.context.attributes
+            .filter((item) => item.key || item.value !== "")
+            .map(Object.values)
```
jonkoops, line 130: Not sure if this works correctly, if I split this code out then `atrributes` gets an `any` type. You might want to make this a bit more explicit.
```suggestion
            .map(({ key, value }) => [key, value])
```

src/components/attribute-input/AttributeInput.tsx
```diff
@@ -52,11 +54,13 @@ export const AttributeInput = ({
     name,
   });
 
+  console.log("huh", fields.length);
```
jonkoops, line 57: Another stray console statement here.
```suggestion
```

jonkoops (changes requested):

jonkoops (approved):

jenny-s51:
All Cypress tests passing locally ✅

PR #1965, opened 2022-01-27
added default group tab to user registration

edewit:
[marvel](https://marvelapp.com/prototype/7h0bjj5/screen/75047135)


src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+  useFetch(
+    () => adminClient.realms.getDefaultGroups({ realm }),
+    (groups) => {
+      setDefaultGroups(groups);
+      setKey(key + 1);
+    },
+    [load]
+  );
+
+  const loader = () => Promise.resolve(defaultGroups!);
```
jonkoops, line 57: Having mechanism in place that does fetching and having multiple different states to handle stages of data loading seems rather complex. Perhaps it would be better to move the side-effect of storing the data in this loader?

```ts
const loader = async () => {
  const groups = await adminClient.realms.getDefaultGroups({ realm });
  setDefaultGroups(groups);
  return groups;
};
```

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  const removeGroup = async () => {
+    try {
+      for (const group of selectedRows) {
+        await adminClient.realms.removeDefaultGroup({
```
jonkoops, line 62: These calls will not happen in parallel, but perhaps they can?

```ts
await Promise.all(
  selectedRows.map((group) =>
    adminClient.realms.removeDefaultGroup({
      realm,
      id: group.id!,
    })
  )
);
```

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+      setSelectedRows([]);
+    } catch (error) {
+      addError("realm-settings:groupRemoveError", error);
+    }
+    reload();
+  };
+
+  const addGroups = async (groups: GroupRepresentation[]) => {
+    try {
+      for (const group of groups) {
```
jonkoops, line 80: Same here, I think this can be done in parallel.

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+            toggleGroupPicker();
+          }}
+          onClose={toggleGroupPicker}
+        />
+      )}
+      {enabled && (
+        <Popover
+          bodyContent={
+            <Trans i18nKey="realm-settings-help:defaultGroups">
+              Default groups allow you...
```
jonkoops, line 129: Instead of adding text here you can also add an empty placeholder.
```suggestion
              {" "}
```

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  </DropdownItem>,
+                ]}
+              />
+            </ToolbarItem>
+          </>
+        }
+        actions={[
+          {
+            title: t("common:remove"),
+            onRowClick: async (group: GroupRepresentation) => {
```
jonkoops, line 193: Nothing `async` is happening here.
```suggestion
            onRowClick: (group: GroupRepresentation) => {
```

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -0,0 +1,228 @@
...
+            displayKey: "groups:path",
+          },
+        ]}
+        emptyState={
+          <ListEmptyState
+            hasIcon
+            message={t("noDefaultGroups")}
+            instructions={
+              <Trans i18nKey="realm-settings:noDefaultGroupsInstructions">
+                Default groups allow you...
```
jonkoops, line 216: Instead of adding text here you can also add an empty placeholder.
```suggestion
                {" "}
```

jonkoops (changes requested):

src/realm-settings/DefaultGroupsTab.tsx
edewit: that is not a good idea now when the component gets unmounted while the request is still going on you are then still setting state, but in a component that is unmounted and thus creating a memory leak

jonkoops: It's not that much of an issue, after the state is set the garbage collector will see there are no longer any references to it an clean it up.

Ideally though the Keycloak table would just allow us to pass data directly to it instead.

edewit: It is a big deal, you'll get an error if you do that and you application will slowly start to run less smooth. That is why we have the `useFetch` hook that will cancel the request and only set the state if the component is still mounted

jonkoops: Ahhh, the unmounted component state update. Yeah I forgot about that one, ok let's leave it as is.

jonkoops (approved):

src/realm-settings/DefaultGroupsTab.tsx
```diff
@@ -190,10 +194,10 @@ export const DefaultsGroupsTab = () => {
         actions={[
           {
             title: t("common:remove"),
-            onRowClick: async (group: GroupRepresentation) => {
+            onRowClick: (group: GroupRepresentation) => {
               setSelectedRows([group]);
               toggleRemoveDialog();
-              return false;
+              return Promise.resolve(false);
```
jonkoops, line 200: `onRowClick` expects a Promise? I wasn't expecting that.

edewit: yes

PR #1990, opened 2022-01-31
Replace Lodash with it's ESM variant

jonkoops:
Replaces `lodash` with `lodash-es` and replaces all default imports with named imports. This helps reduce the final bundle size as we don not use all of the functions provided by Lodash. Also adds an ESLint rule to enforce the usage of named imports.


jonkoops:
Marking this as a draft as some changes are needed to the Jest setup to make this work properly.

jonkoops:
Issues have been resolved, so I am marking this as ready for review.

edewit (approved):

Issue #1500, opened 2021-11-05
[UX review] Priority config is missing for the User fed


yih-wang:
The priority config used to belong to each provider is replaced with centralized management in the card view. The behavior is totally the same with the order management of the IDP list.

Design can be found here - https://marvelapp.com/prototype/8e5ej5g

yih-wang:
Hi @mfrances17, I didn't find the issue you created related to this topic so I just create a new one. Sorry if it's duplicated.

PR #2005, opened 2022-02-04
Add priorities to User Federation

mfrances17:
## Motivation
Adds priority configuration to user fed providers. Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1500.

## Brief Description
Uses a drag and drop modal to change priority of existing user federation providers. Adds an optional lower button/link to the ViewHeader component, which in this case is being used to add the Manage priorities link. Also, user fed cards are now sorted alphabetically.

## Verification Steps
1. Go to User Fed and create at least two providers.
2. Select the Manage priorities link, the names should be in alphabetical order by default. 
3. Drag and drop one or more providers to create a new order, then click Save.
4. Exit and return to verify the priority order is retained. 
5. Go to the old console and verify that the order is correct in the Priority column in User Fed.

## Checklist:

## Additional Notes
Could not get actual drag and drop to work in cypress, so the test is limited to just invoking the modal, verifying the original order, and saving.

### DIFF
--- a/src/user-federation/ManagePriorityDialog.tsx
+++ b/src/user-federation/ManagePriorityDialog.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
-import { sortBy } from "lodash-es";
+import _ from "lodash";
 import {
   AlertVariant,
   Button,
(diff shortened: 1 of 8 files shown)


src/user-federation/ManagePriorityDialog.tsx
```diff
@@ -0,0 +1,147 @@
+import React, { useState } from "react";
+import { useTranslation } from "react-i18next";
+import _ from "lodash";
```
jonkoops, line 3:
```suggestion
import { sortBy } from "lodash-es";
```

jonkoops (approved):
Aside from the linter error around Lodash this PR look good to me. The tests are failing, perhaps a rebase would be in order.

jonkoops (approved):

PR #2038, opened 2022-02-09
Add componentId to LDAP provider test connection/authorization

mfrances17:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/1391.

## Brief Description
Added componentId to 'Test connection' and 'Test authentication' calls so the payload is the same as the old console. Verified that it works with LDAP server (see verification steps). Also fixed several broken help IDs in the LDAP provider form, I suspect they were the victim of an overzealous search/replace for user --> users.

## Verification Steps
1. Go to User Fed and create a new LDAP provider.
2. Add the following (leave everything else as the default):
- **Connection URL**: `ldap://ldap.forumsys.com:389`
- **Users DN**: `dc=example,dc=com`
3. Click **Test connection**. It should say that it succeeded, and the payload in the Network call should include componentId.
4. Add the following (leave everything else as the default):
- **Authentication Type**: `Simple`
- **Bind DN**: `cn=read-only-admin,dc=example,dc=com`
- **Bind Credential**: `password`
5. Click **Test Authentication**. It should say that it succeeded, and the payload in the Network call should include componentId. 

## Checklist:

## Additional Notes
None

### DIFF
--- a/src/user-federation/custom/CustomProviderSettings.tsx
+++ b/src/user-federation/custom/CustomProviderSettings.tsx
@@ -95,8 +95,8 @@ export default function CustomProviderSettings() {
             label={t("consoleDisplayName")}
             labelIcon={
               <HelpItem
-                helpText="users-federation-help:consoleDisplayNameHelp"
-                fieldLabelId="users-federation:consoleDisplayName"
+                helpText="user-federation-help:consoleDisplayNameHelp"
+                fieldLabelId="user-federation:consoleDisplayName"
               />
             }
             helperTextInvalid={t("validateName")}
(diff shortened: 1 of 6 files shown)


jonkoops (approved):

jenny-s51 (commented):
The `testLDAPConnection` calls are looking great with the `componentId` attribute @mfrances17 ! 

While I was testing I noticed a couple things! I was only able to test this by first creating the LDAP provider on the master branch and then switching over to this branch. It looks like the `id` attribute is now being set to `""` upon creation of the LDAP provider. If I try to create a new LDAP provider on this branch, I get a "Page not found error" when I try to click on the card since the `id` is used in the route. 

Screen Shot 2022-02-10 at 1 50 10 PM

mfrances17:
Thanks for catching that @jenny-s51... the newest code should work as expected.

mfrances17:
All cypress tests passed locally

Issue #1900, opened 2022-01-20
[UX Review] Client policies > profile > executor details: the changes cannot be saved


yih-wang:
### Describe the bug

- I tried to change the configuration of the executor. The changes won't be saved whether I changed them during creation or edit.
- Breadcrumb is not correct in the details page. Should be "Realm settings > Client policies > [profile name] > Executor details"
- Breadcrumb is not correct in the creation page. Should be "Realm settings > Client policies > [profile name] > Add executor"

![image](https://user-images.githubusercontent.com/23331885/150306451-95d0b794-9b57-4757-b339-7fe2ae8dd3ad.png)

### Version

15.0.2

edewit:
this is also fixed with https://github.com/keycloak/keycloak-admin-ui/pull/2070

Issue #1898, opened 2022-01-20
[UX Review] Client policies > global profile details: all the settings should be read-only


yih-wang:
### Describe the bug

Users are not permitted to change the global profiles. So all the settings fields should be read-only/disabled.

### Version

15.0.2

### Expected behavior

![image](https://user-images.githubusercontent.com/23331885/150304112-43c0d4a7-351f-4ad4-9be5-42449615713a.png)

### Actual behavior

![image](https://user-images.githubusercontent.com/23331885/150304080-6718ca12-577d-4173-a210-968a3fcb494c.png)

PR #2070, opened 2022-02-11
Made form readonly and fixed initialisation

edewit:
The screen is now readonly for global profiles, but the breadcrumb is not fixed as this is a global issue

fixes: #1898
fixes: #1900

### DIFF
--- a/src/components/form-access/FormAccess.tsx
+++ b/src/components/form-access/FormAccess.tsx
@@ -41,6 +41,12 @@ export type FormAccessProps = FormProps & {
    * @type {boolean}
    */
   unWrap?: boolean;
+
+  /**
+   * Overwrite the fineGrainedAccess and make form regardless of access rights.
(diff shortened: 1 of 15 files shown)


src/components/form-access/FormAccess.tsx
```diff
@@ -41,6 +41,12 @@ export type FormAccessProps = FormProps & {
    * @type {boolean}
    */
   unWrap?: boolean;
+
+  /**
+   * Overwrite the fineGrainedAccess and make form regardless of access rights.
+   * @type {boolean}
```
jonkoops, line 47: Although it was added before already using JSDoc to annotate a type in a TypeScript file seems kind of pointless.

```suggestion

jonkoops (approved):

jenny-s51 (approved):
LGTM - all Cypress tests pass locally ✅

PR #2092, opened 2022-02-14
fixed test if value is set

edewit:
fixes: #1386


jenny-s51 (approved):
LGTM. All Cypress tests pass locally

PR #2105, opened 2022-02-15
Update Keycloak development server to version 17.0.0

jonkoops:
Updates both the Docker and scripted development server to the latest version of Keycloak. This does not yet move us to the Quarkus version, for this a separate issue has been created (#2104).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ COPY . .
 
 RUN mvn -f /app/keycloak-theme/pom.xml install
 
-FROM quay.io/keycloak/keycloak:16.1.0
+FROM quay.io/keycloak/keycloak:17.0.0-legacy
 COPY --from=builder /app/keycloak-theme/target/classes /opt/jboss/keycloak/themes/keycloak.v2
 
 EXPOSE 8080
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY . .
 
 RUN mvn -f /app/keycloak-theme/pom.xml install
 
-FROM quay.io/keycloak/keycloak:16.1.0
+FROM quay.io/keycloak/keycloak:17.0.0-legacy
```
edewit, line 8: why is it called legacy?

jonkoops: This is the WildFly based distribution, whereas the new Quarkus distribution does not have a suffix.

edewit: shouldn't we switch to the quarkus version instead?

jonkoops: No, that would require we change some of our scripts. An issue as been created to address this later.

edewit (approved):

Dockerfile
edewit: ahh I see the `add-user` script is no longer there

jonkoops: Yes, that is one of the changes.

Issue #2130, opened 2022-02-21
Users -> Credentials tab -> Drag & drop of credentials within expandable row is not working


agagancarczyk:
### Describe the bug

Drag & drop of credentials within an expandable row (within the same group) is not working. 

The expanded row closes unexpectedly when trying to do drag & drop of credentials within the expendable row. And then when trying to expand it again the error occurs "Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."

### Version

17.0.0

### Expected behavior

User should be able to drag & drop credentials within the expandable row as per https://marvelapp.com/prototype/a9b33db/screen/83342106

### Actual behavior

The expanded row closes unexpectedly when trying to do drag & drop. And then when trying to expand it again the error occurs "Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."

### How to Reproduce?

1. In the old console, add 2 or 3 "webauth register paswordless" credentilas.
2. In the new console, go to Users -> Credentials tab and expand the grouped "webauth register paswordless" group
3. Try to drag and drop "webauth register paswordless" credentials.

PR #2131, opened 2022-02-21
User Credentials -> Draggable user credentials

agagancarczyk:
## Motivation
1. Make the rows in the Credentials tab draggable as per https://marvelapp.com/prototype/a9b33db/screen/83342106. 
2. Add more tests.

### In the old console, add credentials as follows: 

1. To add "webauth register passwordless" credential -> While in Master realm, go to Users menu item, select Details tab, select "Webauthn Register Passwordless" from Required User actions dropdown, save, logout. (If you don't have "Webauthn register passwordless" option in dropdown, go to Authentication sidebar menu, select "Required Actions", select "Register" btn in the right top corner, select "Webauthn register password", tick enable and default action -> you should now be able to select it from Required User actions dropdown on Users/Details tab).

Login, select "Register" btn in Security Key Registration dialog, select "This device" and leave the default label "WebAuthn Authenticator (Default Label)", select ok btn.

2. To add "otp" credential, go to Manage Account in the top right corner, Account Security, select Signing in Device Activity, select Signing In and Set up Authenticator Application, logout, login.

### To test this feature:

1. Go to Users -> select user -> select Credentials tab 
2. Drag and drop credentials to change their position
3. There is a bug open for drag & drop of credentials within an expandable row #2130

All tests are passing locally: 
Screenshot 2022-02-21 at 09 29 52

## Checklist:


src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+          draggingToItemIndex
+        );
+        move(tempItemOrder);
+        setState({
+          ...state,
+          draggingToItemIndex,
+          tempItemOrder,
+        });
+      }
+    }
```
jonkoops, line 273: Since the `if` clause terminates the function with a return statement, you can omit the `else` branch.

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+      !curListItem ||
+      (bodyRef.current && !bodyRef.current.contains(curListItem)) ||
+      curListItem.id === state.draggedItemId
+    ) {
+      return null;
+    } else {
+      const dragId = curListItem.id;
+      const draggingToItemIndex = Array.from(
+        bodyRef.current?.children || []
+      ).findIndex((item) => item.id === dragId);
```
jonkoops, line 259: If the previous suggestion is applied you can remove the null checks here.
```suggestion
      const draggingToItemIndex = Array.from(bodyRef.current.children).findIndex(
        (item) => item.id === dragId
      );
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+  const onDragOver = (evt: React.DragEvent) => {
+    evt.preventDefault();
+    const td = evt.target as HTMLTableCellElement;
+    const curListItem = td.closest("tr");
+    if (
+      !curListItem ||
+      (bodyRef.current && !bodyRef.current.contains(curListItem)) ||
+      curListItem.id === state.draggedItemId
+    ) {
+      return null;
```
jonkoops, line 254: I think you can return here without a value.
```suggestion
      return;
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+          state.draggedItemId,
+          draggingToItemIndex
+        );
+        move(tempItemOrder);
+        setState({
+          ...state,
+          draggingToItemIndex,
+          tempItemOrder,
+        });
+      }
```
jonkoops, line 272: You can reduce the nesting of this statement by using an early return.
```suggestion
    if (draggingToItemIndex === state.draggingToItemIndex) {
      return;
    }
    const tempItemOrder = moveItem(
      itemOrder,
      state.draggedItemId,
      draggingToItemIndex
    );
    move(tempItemOrder);
    setState({
      ...state,
      draggingToItemIndex,
      tempItemOrder,
    });
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+    const tr = evt.target as HTMLTableRowElement;
+    tr.classList.remove(styles.modifiers.ghostRow);
+    tr.setAttribute("aria-pressed", "false");
+    setState({
+      ...state,
+      draggedItemId: "",
+      draggingToItemIndex: -1,
+      dragging: false,
+    });
+  };
```
jonkoops, line 286: Are you sure `target` is always an table row? Unlike [`currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) it is possible for [`target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) to be different depending on where the event originated.

Instead of using a type-cast you could do a run-time check to see if it is the element you expected it to be.

```suggestion
  const onDragEnd = ({ target }: React.DragEvent) => {
    if (!(target instanceof HTMLTableRowElement)) {
      return;
    }
    target.classList.remove(styles.modifiers.ghostRow);
    target.setAttribute("aria-pressed", "false");
    setState({
      ...state,
      draggedItemId: "",
      draggingToItemIndex: -1,
      dragging: false,
    });
  };
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+    if (isValidDrop(evt)) {
+      onDragFinish(state.draggedItemId, state.tempItemOrder);
+    } else {
+      onDragCancel();
+    }
+  };
+
+  const onDragOver = (evt: React.DragEvent) => {
+    evt.preventDefault();
+    const td = evt.target as HTMLTableCellElement;
```
jonkoops, line 247: Are you sure `target` is always an table cell? Unlike [`currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) it is possible for [`target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) to be different depending on where the event originated.

Instead of using a type-cast you could do a run-time check to see if it is the element you expected it to be.

```ts
if (!(target instanceof HTMLTableCellElement)) {
  return;
}
```

Or if you don't care what kind of element it is:
```ts
if (!(target instanceof HTMLElement)) {
  return;
}
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+      ...state,
+      draggedItemId: "",
+      draggingToItemIndex: -1,
+      dragging: false,
+    });
+  };
+
+  const onDragFinish = async (_dragged: string, newOrder: string[]) => {
+    const dragged = _dragged.split(",")[0];
+    const keys = groupedUserCredentials.map((e) => e.value.map((c) => c.id));
```
jonkoops, line 290: You can make this code a bit more readable by dropping the single letter variables and using destructuring instead.
```suggestion
    const keys = groupedUserCredentials.map(({ value }) =>
      value.map(({ id }) => id)
    );
```

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+    const oldIndex = keys.findIndex((el) => el.join().includes(dragged));
+    const newIndex = newOrder.findIndex((el) => el.includes(dragged));
+    const times = newIndex - oldIndex;
+
+    try {
+      for (let index = 0; index < Math.abs(times); index++) {
```
jonkoops, line 296: Any change these API calls can be made in parallel by using `Promise.all()`?

src/user/UserCredentials.tsx
```diff
@@ -205,17 +370,42 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
                 <Th>{t("type")}</Th>
                 <Th>{t("userLabel")}</Th>
                 <Th>{t("data")}</Th>
                 <Th />
                 <Th />
               </Tr>
             </Thead>
-            <Tbody>
+            <Tbody
+              ref={bodyRef}
```
jonkoops, line 382: You could probably get rid of this ref since all event listeners added here will have a `currentTarget` that should be equal to the body element.

src/user/UserCredentials.tsx
```diff
@@ -205,17 +370,42 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+            >
               {groupedUserCredentials.map((groupedCredential, rowIndex) => (
                 <Fragment key={`table-${groupedCredential.key}`}>
-                  <Tr>
+                  <Tr
+                    id={groupedCredential.value
+                      .map((credential) => {
+                        return credential.id;
+                      })
+                      .toString()}
```
jonkoops, line 394: This code could be a little smaller.
```suggestion
                    id={groupedCredential.value.map(({ id }) => id).toString()}
```

src/user/UserCredentials.tsx
```diff
@@ -205,17 +370,42 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+                    onDrop={onDrop}
+                    onDragEnd={onDragEnd}
+                    onDragStart={onDragStart}
+                  >
+                    <Td
+                      draggableRow={{
+                        id: `draggable-row-${groupedCredential.value.map(
+                          (credential) => {
+                            return credential.id;
+                          }
```
jonkoops, line 405: This code could be a little smaller.
```suggestion
                        id: `draggable-row-${groupedCredential.value.map(
                          ({ id }) => id
                        )}`,
```

jonkoops (changes requested):

src/user/UserCredentials.tsx
```diff
@@ -256,8 +447,21 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
                   </Tr>
                   {groupedCredential.isExpanded &&
                     groupedCredential.value.map((credential) => (
-                      <Tr key={`child-key-${credential.id}`}>
+                      <Tr
+                        key={`child-key-${credential.id}`}
```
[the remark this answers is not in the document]
agagancarczyk: Thanks @jonkoops , but it seems to be required. Removing it is causing "Missing "key" prop for element in iteratoreslint[react/jsx-key](https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-key.md)"

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+  };
+
+  const moveItem = (arr: string[], i1: string, toIndex: number) => {
+    const fromIndex = arr.indexOf(i1);
+    if (fromIndex === toIndex) {
+      return arr;
+    }
+    const temp = arr.splice(fromIndex, 1);
+    arr.splice(toIndex, 0, temp[0]);
+    return arr;
```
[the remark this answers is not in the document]
agagancarczyk: very nice! Thx

src/user/UserCredentials.tsx
agagancarczyk: The else seems to be required. Otherwise, drag & drop doesn't work.

src/user/UserCredentials.tsx
```diff
@@ -150,6 +158,163 @@ export const UserCredentials = ({ user }: UserCredentialsProps) => {
...
+  const onDrop = (evt: React.DragEvent) => {
+    if (isValidDrop(evt)) {
+      onDragFinish(state.draggedItemId, state.tempItemOrder);
+    } else {
+      onDragCancel();
+    }
+  };
+
+  const onDragOver = (evt: React.DragEvent) => {
+    evt.preventDefault();
```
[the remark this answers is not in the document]
agagancarczyk: Good question. This is however needed for drag & drop to work.

src/user/UserCredentials.tsx
agagancarczyk: The else seems to be required. Otherwise, drag & drop doesn't work. Therefore this change can't be made.

src/user/UserCredentials.tsx
agagancarczyk: This ref is required somehow. Otherwise, drag & drop is not going to work.

src/user/UserCredentials.tsx
agagancarczyk: Thanks @jonkoops - this works too.

src/user/UserCredentials.tsx
agagancarczyk: thanks @jonkoops event is needed here so I can't change it to target. Otherwise, drag & drop doesn't work.

src/user/UserCredentials.tsx
agagancarczyk: Here I'm trying to be consistent with the similar implementations as in FlowDetails and RequiredActions.

src/user/UserCredentials.tsx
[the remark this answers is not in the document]
jonkoops: Strange, but not a problem I guess.

jonkoops (approved):

PR #2133, opened 2022-02-22
more robust links out from events

edewit:
fixes: #1887

### DIFF
--- a/src/events/ResourceLinks.tsx
+++ b/src/events/ResourceLinks.tsx
@@ -1,0 +1,122 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import { Tooltip } from "@patternfly/react-core";
+
+import type AdminEventRepresentation from "@keycloak/keycloak-admin-client/lib/defs/adminEventRepresentation";
+import { useRealm } from "../context/realm-context/RealmContext";
+import { toClient } from "../clients/routes/Client";
+import { toGroups } from "../groups/routes/Groups";
+import { toClientScope } from "../client-scopes/routes/ClientScope";
+import { toUser } from "../user/routes/User";
+import { toRealmRole } from "../realm-roles/routes/RealmRole";
+import { toFlow } from "../authentication/routes/Flow";
+
+type ResourceLinkProps = {
+  event: AdminEventRepresentation;
+};
+
+const MAX_TEXT_LENGTH = 38;
+const Truncate = ({
+  text,
+  children,
+}: {
+  text?: string;
+  children: (text: string) => any;
+}) => {
+  const definedText = text || "";
+  const needsTruncation = definedText.length > MAX_TEXT_LENGTH;
+  const truncatedText = definedText.substring(0, MAX_TEXT_LENGTH);
+  return (
+    <>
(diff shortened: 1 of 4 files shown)


src/events/ResourceLinks.tsx
```diff
@@ -0,0 +1,122 @@
...
+  text?: string;
+  children: (text: string) => any;
+}) => {
+  const definedText = text || "";
+  const needsTruncation = definedText.length > MAX_TEXT_LENGTH;
+  const truncatedText = definedText.substring(0, MAX_TEXT_LENGTH);
+  return (
+    <>
+      {needsTruncation && (
+        <Tooltip content={text}>{children(truncatedText + "...")}</Tooltip>
```
jonkoops, line 32: There is a special character that you can use for the three dots.
```suggestion
        <Tooltip content={text}>{children(truncatedText)}&hellip;</Tooltip>
```

src/events/ResourceLinks.tsx
```diff
@@ -0,0 +1,122 @@
...
+}: {
+  text?: string;
+  children: (text: string) => any;
+}) => {
+  const definedText = text || "";
+  const needsTruncation = definedText.length > MAX_TEXT_LENGTH;
+  const truncatedText = definedText.substring(0, MAX_TEXT_LENGTH);
+  return (
+    <>
+      {needsTruncation && (
```
jonkoops, line 31: This could be a tenary which eliminates the need to wrap the return value in a fragement.

src/events/ResourceLinks.tsx
```diff
@@ -0,0 +1,122 @@
...
+  event: AdminEventRepresentation;
+};
+
+const MAX_TEXT_LENGTH = 38;
+const Truncate = ({
+  text,
+  children,
+}: {
+  text?: string;
+  children: (text: string) => any;
```
jonkoops, line 24: Is there a better return type we can use here than `any`?

src/events/ResourceLinks.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return "";
+};
+
+export const ResourceLink = ({ event }: ResourceLinkProps) => {
+  const { realm } = useRealm();
+  return (
+    <Truncate text={event.resourcePath}>
+      {(text) => (
+        <>
+          {isLinkable(event) && (
```
jonkoops, line 110: This could also be a ternary so the condition does not have to be repeated and the outer fragment can be removed.

jonkoops (changes requested):

src/events/ResourceLinks.tsx
edewit: This way the `...` is no longer part of the href

jonkoops: Ahh, yeah you are right. I updated the suggestion to use the unicode character instead.

jonkoops (approved):
LGTM. Running the tests to see if we can merge this.

mfrances17:
LGTM, all local tests passed for me as well ✔️

mfrances17 (approved):

PR #2141, opened 2022-02-23
Run Cypress tests on Chrome when running headlessly

jonkoops:
This more accurately reflects what browser our actual users are running. Another advantage is that this will use a recent version of Chrome, whereas Cypress ships with a version of Electron that is locked to the Cypress version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "start:rh-sso": "THEME_NAME=rh-sso snowpack dev",
     "test": "jest",
     "start:cypress": "cypress open",
-    "start:cypress-tests": "cypress run",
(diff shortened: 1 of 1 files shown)


jenny-s51 (approved):
LGTM - runs on Chrome locally with no issues 👍

Aboullos (approved):
LGTM, execution time seems to be more or less the same

jonkoops:
Thanks for checking @jenny-s51 @Aboullos!

jonkoops:
Merging this as it seems to make some tests less flaky and I need it for some of my other test work.

PR #2172, opened 2022-03-01
Disable test retries for Cypress

jonkoops:
Disables the test retries for Cypress in order to make it more apparent which tests are flaky. The idea is to keep this PR open and rebased and provide the actual test fixes in other PRs until we see a green CI here consistently.


edewit (approved):

jonkoops:
I've ran the tests on the CI 5 times and every time it passed, so I think we can go ahead and merge this change.

PR #2177, opened 2022-03-02
split out the key based into a sperate component

edewit:
it's made it's base component to hard to reuse and was only used once

### DIFF
--- a/src/clients/authorization/KeyBasedAttributeInput.tsx
+++ b/src/clients/authorization/KeyBasedAttributeInput.tsx
@@ -40,6 +40,7 @@ type AttributeInputProps = {
 };
 
 type ValueInputProps = {
+  name: string;
   rowIndex: number;
   attribute: any;
   selectableValues?: AttributeType[];
@@ -47,6 +48,7 @@ type ValueInputProps = {
 };
 
 const ValueInput = ({
+  name,
   rowIndex,
   attribute,
   selectableValues,
@@ -226,6 +228,7 @@ export const KeyBasedAttributeInput = ({
               />
             </Td>
             <ValueInput
+              name={name}
               attribute={attribute}
               rowIndex={rowIndex}
               selectableValues={selectableValues}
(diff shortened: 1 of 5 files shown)


src/clients/authorization/KeyBasedAttributeInput.tsx
```diff
@@ -0,0 +1,268 @@
...
+              aria-label={t("roles:addAttributeText")}
+              id="plus-icon"
+              variant="link"
+              className="kc-attributes__plus-icon"
+              onClick={() => {
+                append({ key: "", value: "" });
+                setIsKeyOpenArray([...isKeyOpenArray, false]);
+              }}
+              icon={<PlusCircleIcon />}
+              isDisabled={!watchLastValue}
```
jenny-s51, line 258: Currently this is always disabled even if a value is typed into the attribute value input field.

src/clients/authorization/KeyBasedAttributeInput.tsx
```diff
@@ -0,0 +1,268 @@
...
+                      >
+                        {attribute.name}
+                      </SelectOption>
+                    ))}
+                  </Select>
+                )}
+              />
+            </Td>
+            <ValueInput
+              attribute={attribute}
```
jenny-s51, line 229:
```suggestion
              name={name}
              attribute={attribute}
```

jenny-s51 (changes requested):
Good idea splitting this up into a separate component @edewit !

I tested this in the Authorization -> Evaluate tab of Clients and noticed that whatever was typed into the "value" field wasn't being sent with the data. 

Passing the `name` prop to `ValueInput` seems to fix it!

src/clients/authorization/KeyBasedAttributeInput.tsx
edewit: good catch...

jonkoops (changes requested):

jonkoops (changes requested):
Changes look good, but the tests are failing. @edewit can you take a look?

jenny-s51 (approved):

jonkoops (approved):

PR #2182, opened 2022-03-03
catch 403 when only query-clients is allowed

edewit:
fixes: #2149


jonkoops (approved):

ssilvert (approved):
I think this is fine for now.  I'm trying to come up with a more holistic approach that combines role-based and fine-grained authentication so we will be able to more easily manage what should be disallowed, read-only, or hidden.

It will be at least a couple of weeks before I have a concrete proposal though.

mfrances17 (approved):
LGTM

PR #2195, opened 2022-03-04
default value and required field for generate keys

edewit:
fixes: #1715

### DIFF
--- a/src/clients/keys/GenerateKeyDialog.tsx
+++ b/src/clients/keys/GenerateKeyDialog.tsx
@@ -1,6 +1,11 @@
 import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
-import { Control, Controller, useForm } from "react-hook-form";
+import {
+  Controller,
+  FormProvider,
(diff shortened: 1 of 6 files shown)


mfrances17 (dismissed):
Code looks good, tested well. Only one comment - Key alias now defaults to client name by default as requested in the issue, but not when you choose to import. Shouldn't that be the case there as well?

jonkoops (changes requested):
Changes look good, but some of the tests are failing. Can you take a look @edewit?

edewit:
> Code looks good, tested well. Only one comment - Key alias now defaults to client name by default as requested in the issue, but not when you choose to import. Shouldn't that be the case there as well?

I don't know @xianli123, what about this case?

jonkoops (approved):

xianli123:
Hi @edewit @jonkoops @mfrances17 , the default value of "Key alias" should be the client name when users generate the new key. Key alias will be an input field when a user imports a certificate.

Issue #1718, opened 2021-12-16
Enabled and Disable features incorrect


ssilvert:
### Describe the bug

On Master realm home page, Enabled/Disable features do not match old UI's Server Info.  For instance, DOCKER is listed as both "Enabled" and "Disabled".  MAP_STORAGE is experimental and enabled but is shown as disabled.  Etc.

### Version

15.1.0

### Anything else?

Source is https://github.com/keycloak/keycloak/discussions/9119#discussioncomment-1823140

edewit:
This is because the `disabledFeatures` of the server info is not correct if I look at the Profile class in keycloak I see the following list with features:
`AUTHORIZATION, ACCOUNT2, ACCOUNT, ADMIN, ADMIN2, DOCKER, IMPERSONATION, OPENSHIFT, SCRIPTS, TOKEN, UPLOAD, WEB, CLIENT, CIBA, MAP, PAR, DECLARATIVE, DYNAMIC`
When I startup with no features enabled and look at the `disabledFeatures` in the server info I get the following:
`ADMIN2, ADMIN_FINE_GRAINED_AUTHZ, DECLARATIVE_USER_PROFILE, DOCKER, DYNAMIC_SCOPES, MAP_STORAGE, OPENSHIFT_INTEGRATION, SCRIPTS, TOKEN_EXCHANGE, UPLOAD_SCRIPTS`
So this means we can never tell if `CIBA` is on or not?

PR #2196, opened 2022-03-04
improved the enable / disabled feature

edewit:
list but still broken see: #1718


src/dashboard/Dashboard.tsx
```diff
@@ -56,15 +56,30 @@ const EmptyDashboard = () => {
   );
 };
 
+const ALL_FEATURES = [
```
jonkoops, line 59: If you combine all the values in `serverInfo.profileInfo` do you not get the same result? Since the server can always add new features I don't think we should be hard-coding this.

jonkoops (changes requested):

PR #2259, opened 2022-03-14
Make theme deployable from archive

jonkoops:
Changes the build so the theme can be [deployed as an archive](https://www.keycloak.org/docs/latest/server_development/index.html#deploying-themes). This allows the theme to be added into the Keycloak installation as a JAR, without the need to extract it.

This also helps eliminate the need to build the Keycloak server and theme together, which can simplify our integration tests and possibly reduce the time needed to run the tests in a future change.

Merge together with: https://github.com/keycloak/keycloak/pull/11281

### DIFF
--- a/keycloak-theme/README.md
+++ b/keycloak-theme/README.md
@@ -1,25 +1,13 @@
-# Keycloak Admin Console V2 Maven Build
+# Keycloak Theme (Maven Build)
 
-The Maven build prepares the admin console to be deployed as a theme on the Keycloak server.
+This directory contains the Maven build for the Keycloak theme. It allows the theme to be built as a JAR which can be included when running the Keycloak server.
 
-# Build Instructions
+## Building
 
 ```bash
 mvn install
 ```
 
-# Deployment
+## Deployment
 
-The jar created with `mvn install` needs to be deployed to a Maven repository. From there, it will become part of the Keycloak server build.
-
-For development, you can also just copy the contents of `./target/classes` to `<keycloak server>/themes/keycloak.v2`. Then restart the server.
-
-# To Run
-
-Until New Admin Console becomes the default, you will need to start Keycloak server like this:
-
-```bash
-$> bin/standalone.sh -Dkeycloak.profile.feature.admin2=enabled
-```
-
-Then go to `Realm Settings --> Themes` and set Admin Console Theme to `keycloak.v2`.
+First build the this repository with the instructions above, then [build the Keycloak sever](https://github.com/keycloak/keycloak/blob/main/docs/building.md). Start the Keycloak server and navigate to `Realm Settings` ➡️ `Themes` and set admin theme to `keycloak.v2`.
--- a/keycloak-theme/src/main/resources/META-INF/keycloak-themes.json
+++ b/keycloak-theme/src/main/resources/META-INF/keycloak-themes.json
@@ -1,0 +1,10 @@
+{
+  "themes": [
+    {
+      "name": "keycloak.v2",
+      "types": [
+        "admin"
+      ]
+    }
+  ]
+}
--- a/keycloak-theme/.gitignore
+++ b/keycloak-theme/.gitignore
@@ -1,4 +1,3 @@
-src/
 node/
 # https://github.com/github/gitignore/blob/master/Maven.gitignore
 target/
--- a/keycloak-theme/pom.xml
+++ b/keycloak-theme/pom.xml
@@ -119,17 +119,6 @@
 
         <plugins>
             <plugin>
-                <artifactId>maven-clean-plugin</artifactId>
-                <version>3.2.0</version>
-                <configuration>
-                    <filesets>
-                        <fileset>
-                            <directory>src</directory>
-                        </fileset>
-                    </filesets>
-                </configuration>
-            </plugin>
-            <plugin>
                 <groupId>com.github.eirslett</groupId>
                 <artifactId>frontend-maven-plugin</artifactId>
                 <version>1.12.1</version>
@@ -144,6 +133,9 @@
                         <goals>
                             <goal>npm</goal>
                         </goals>
+                        <configuration>
+                            <arguments>ci</arguments>
+                        </configuration>
                     </execution>
                     <execution>
                         <id>npm run build</id>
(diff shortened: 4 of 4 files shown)


keycloak-theme/src/main/resources/META-INF/keycloak-themes.json
```diff
@@ -0,0 +1,10 @@
+{
+  "themes": [
+    {
+      "name": "admin-ui",
```
ssilvert, line 4: If this is the way we will integrate with the server from now on then I believe the name here should be `keycloak.v2`.

If it's only used for testing we can name it anything we want.

jonkoops: This is just a quick PoC looking for feedback, so your input is very welcome. Let's make it `keycloak-v2` as this name needs to match the directory from what I can see in the docs.

In the future we can expose more variants here when we also pull in the `profile` project into the repo.

edewit:
I don't fully understand this change atm the admin ui is already a jar file in the quarkus distribution.

edewit (approved):

stianst:
> I don't fully understand this change atm the admin ui is already a jar file in the quarkus distribution.

admin-ui is not really a JAR right now, but rather just an archive with files that are extracted, then packaged into keycloak-themes JAR. The change here is that admin-ui is now a proper JAR that is added to the Keycloak server as a dependency directly/

stianst:
This looks like it will work, but wouldn't be installed by copying to the providers directory. Rather, we'd add a dependency on org.keycloak:keycloak-admin-ui to https://github.com/keycloak/keycloak/blob/main/quarkus/runtime/pom.xml, which means it would be included inside `lib/lib/main` alongside the current `...keycloak-themes...jar`.

The dependencies in `lib` is not something you can currently change easily as it is baked into the Quarkus dist by more than just being a jar in a directory. So, not sure how you'd deploy a custom build, but not sure if that's really needed.

.github/workflows/cypress.yml
```diff
@@ -35,28 +35,45 @@ jobs:
...
-        run: mvn clean install --batch-mode --file keycloak-theme/pom.xml
+      - name: Build theme
+        working-directory: ${{ env.ADMIN_UI_PATH }}/keycloak-theme
+        run: mvn clean install --batch-mode
+
+      - name: Determine theme path
+        working-directory: ${{ env.ADMIN_UI_PATH }}/keycloak-theme
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "THEME_JAR_PATH=admin-ui/keycloak-theme/target/keycloak-admin-ui-$VERSION.jar" >> $GITHUB_ENV
```
stianst, line 47: This isn't needed I'd say
```suggestion
      - name: Determine theme path
        working-directory: ${{ env.ADMIN_UI_PATH }}/keycloak-theme
        run: |
          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
          echo "THEME_JAR_PATH=admin-ui/keycloak-theme/target/keycloak-admin-ui-$VERSION.jar" >> $GITHUB_ENV

```

.github/workflows/cypress.yml
```diff
@@ -35,28 +35,45 @@ jobs:
...
           tar -xvf quarkus/dist/target/keycloak-$VERSION.tar.gz -C quarkus/dist/target/
-          echo "KEYCLOAK_SERVER_BIN=keycloak-server/quarkus/dist/target/keycloak-$VERSION/bin" >> $GITHUB_ENV
+          echo "KEYCLOAK_SERVER_DIST=keycloak-server/quarkus/dist/target/keycloak-$VERSION" >> $GITHUB_ENV
 
-      - name: Start Keycloak Server
-        working-directory: ${{ env.KEYCLOAK_SERVER_BIN }}
-        run: ./kc.sh start-dev --features=admin2,admin-fine-grained-authz,declarative-user-profile &> ~/server.log &
+      - name: Copy Admin UI theme
+        run: cp ${{ env.THEME_JAR_PATH }} ${{ env.KEYCLOAK_SERVER_DIST }}/providers/admin-ui.jar
+
```
stianst, line 61: I don't think this will actually work as admin-ui jar is already on the classpath, this basically results in adding the same theme/jar twice on the classpath.

Doesn't matter though, you can just remove this as you already have the latest built admin-ui included in the Keycloak dist above.

```suggestion
```

keycloak-theme/README.md
```diff
@@ -1,25 +1,13 @@
...
-# To Run
-
-Until New Admin Console becomes the default, you will need to start Keycloak server like this:
-
-```bash
-$> bin/standalone.sh -Dkeycloak.profile.feature.admin2=enabled
-```
-
-Then go to `Realm Settings --> Themes` and set Admin Console Theme to `keycloak.v2`.
+Copy the JAR created by the build into the `providers` directory of your Keycloak installation and start the server. Then navigate to `Realm Settings` ➡️ `Themes` and set admin theme to `keycloak.v2`.
```
stianst, line 13: This should probably be changed to first build admin-ui with Maven, then build Keycloak with Maven. I don't think you can actually copy the admin-ui jar to providers dir as it would result in it being twice on the classpath. If needed we could look to see if there's some solution to this though.

.github/workflows/cypress.yml
```diff
@@ -35,28 +35,45 @@ jobs:
...
-        working-directory: ${{ env.KEYCLOAK_SERVER_BIN }}
-        run: ./kc.sh start-dev --features=admin2,admin-fine-grained-authz,declarative-user-profile &
+      - name: Copy Admin UI theme
+        run: cp ${{ env.THEME_JAR_PATH }} ${{ env.KEYCLOAK_SERVER_DIST }}/providers/admin-ui.jar
+
+      - name: Start Keycloak server
+        working-directory: ${{ env.KEYCLOAK_SERVER_DIST }}/bin
+        run: |
+          ./kc.sh start-dev --features=admin2,admin-fine-grained-authz,declarative-user-profile &
+          sleep 20s
```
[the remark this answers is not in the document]
jonkoops: Ok, let's do this in a follow up PR :+1:

.github/workflows/cypress.yml
jonkoops: Good call, removed this logic.

jonkoops:
I have made a PR (https://github.com/keycloak/keycloak/pull/11281) on the Keycloak repo to get rid of the extraction logic, and include this change as a regular dependency. We can merge it after this PR has been merged.

jonkoops:
Note the CI will fail, as it expects Keycloak to install the JAR as per https://github.com/keycloak/keycloak/pull/11281. This is expected and should not hold up a merge.

stianst (approved):

PR #2272, opened 2022-03-15
Remove ability to run dev server through Docker

jonkoops:
The instructions for running the Docker container are outdated and as far as I know nobody is actually running the application through Docker during development. Let's remove this as it's just a maintenance burden at this point.


agagancarczyk (approved):
LGTM @jonkoops

mfrances17 (approved):
👍

PR #2278, opened 2022-03-16
Create attribute - delete validator

agagancarczyk:
## Motivation
Added deleting validator

## Verification Steps
1. Go to  realm-settings, user profile tab
2. Click create attribute btn, add any 2 validators and remove 1 of them and verify that removed validator is not listed in the validations table
3. Click save

## Checklist:


src/realm-settings/NewAttributeSettings.tsx
```diff
@@ -142,7 +142,16 @@ export default function NewAttributeSettings() {
         : profileConfig.scopeRequired,
     };
 
-    const validations = profileConfig.validations;
+    const validations = profileConfig.validations?.reduce(
+      (prevValidations: any, currentValidations: any) => {
+        prevValidations[currentValidations.name] =
+          currentValidations.config.length === 0
+            ? {}
+            : currentValidations.config;
```
edewit, line 150: So if I read this correctly if the array is empty make it a empty object otherwise leave the array? Is that really nessary?

edewit (commented):
:+1:

src/realm-settings/NewAttributeSettings.tsx
agagancarczyk: Thanks Erik, yes, this is correct, BE is expecting {} not []. Validators with empty config are defined as config=[] in Validators.ts

edewit (approved):

PR #2294, opened 2022-03-18
don't test ldap when form isn't valid

edewit:
fixes: #1776

### DIFF
--- a/src/user-federation/kerberos/KerberosSettingsRequired.tsx
+++ b/src/user-federation/kerberos/KerberosSettingsRequired.tsx
@@ -70,6 +70,8 @@ export const KerberosSettingsRequired = ({
           }
           fieldId="kc-console-display-name"
           isRequired
+          validated={form.errors.name ? "error" : "default"}
+          helperTextInvalid={form.errors.name?.message}
         >
           {/* These hidden fields are required so data object written back matches data retrieved */}
           <TextInput
@@ -109,10 +111,8 @@ export const KerberosSettingsRequired = ({
               },
             })}
             data-testid="kerberos-name"
+            validated={form.errors.name ? "error" : "default"}
           />
-          {form.errors.name && (
-            <div className="error">{form.errors.name.message}</div>
-          )}
         </FormGroup>
 
         <FormGroup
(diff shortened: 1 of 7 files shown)


jenny-s51 (changes requested):
Form validation working as expected @edewit - nice fix!

Can we use the `helperTextInvalid` prop on the `connectionUrl` field? The helper text renders in black for this field but for the other fields the invalid text shows up red. Using `helperTextInvalid` seems to fix it! WDYT @edewit ?

edewit:
Yes @jenny-s51 seems that parts of this form have some custom helper text rendering, updated this PR to change it to PF

src/user-federation/kerberos/KerberosSettingsRequired.tsx
```diff
@@ -125,6 +125,10 @@ export const KerberosSettingsRequired = ({
           }
           fieldId="kc-kerberos-realm"
           isRequired
+          validated={
+            form.errors.config?.kerberosRealm?.[0] ? "error" : "default"
+          }
+          helperTextInvalid={form.errors.config.kerberosRealm[0].message}
```
jenny-s51, line 131:
```suggestion
          helperTextInvalid={form.errors.config.kerberosRealm?.[0].message}
```

src/user-federation/kerberos/KerberosSettingsRequired.tsx
```diff
@@ -187,6 +191,8 @@ export const KerberosSettingsRequired = ({
           }
           fieldId="kc-key-tab"
           isRequired
+          validated={form.errors.config?.keyTab?.[0] ? "error" : "default"}
+          helperTextInvalid={form.errors.config.keyTab[0].message}
```
jenny-s51, line 195:
```suggestion
          helperTextInvalid={form.errors.config.keyTab?.[0].message}
```

jenny-s51 (changes requested):
Thanks for refactoring those @edewit ! LDAP form is looking good -- there are just a couple areas in the Kerberos form where we should be using dot notation to prevent undefined errors

edewit:
Good catch @jenny-s51 don't know why I didn't see that...

jenny-s51 (approved):
LGTM! 👍

PR #2297, opened 2022-03-18
fixed help text

edewit:
fixes: #1181

### DIFF
--- a/src/realm-settings/help.ts
+++ b/src/realm-settings/help.ts
@@ -56,7 +56,7 @@ export default {
     xRobotsTag:
       "Prevent pages from appearing in search engines <1>Learn more</1>",
     xXSSProtection:
-      "This header configures the Cross-site scripting (XSS) filter in your browser. Using the default behaviour, the browser will prevent rendering of the page when a XSS attack is detected. <1>Learn more</1>",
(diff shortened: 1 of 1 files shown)


jenny-s51 (changes requested):
I think we need to keep the "Learn more" link here! In the designs for the new admin console, the labels under "Security Defenses" are no longer clickable links and the "Learn more" button is used instead:

https://marvelapp.com/prototype/13g2f55e/screen/75843595

src/realm-settings/help.ts
```diff
@@ -56,7 +56,7 @@ export default {
     xRobotsTag:
       "Prevent pages from appearing in search engines <1>Learn more</1>",
     xXSSProtection:
-      "This header configures the Cross-site scripting (XSS) filter in your browser. Using the default behaviour, the browser will prevent rendering of the page when a XSS attack is detected. <1>Learn more</1>",
+      "This header configures the Cross-site scripting (XSS) filter in your browser. Using the default behavior, the browser will prevent rendering of the page when a XSS attack is detected (click label for more information)",
```
jenny-s51, line 59:
```suggestion
      "This header configures the Cross-site scripting (XSS) filter in your browser. Using the default behavior, the browser will prevent rendering of the page when a XSS attack is detected <1>Learn more</1>",
```

edewit:
Right, Jenny what do you think  @xianli123 do we need clickable labels?

Issue #2289, opened 2022-03-17
Realm settings(Keys -> Keys list): Filter dropdown is broken


jenny-s51:
### Describe the bug

When selecting a value from the filter dropdown, the selection only works on initial select and a new value cannot be selected. 

### Version

17

PR #2315, opened 2022-03-21
fixed the filter dropdown

edewit:
fixes: #2289

### DIFF
--- a/src/components/table-toolbar/KeycloakDataTable.tsx
+++ b/src/components/table-toolbar/KeycloakDataTable.tsx
@@ -121,11 +121,17 @@ export type Action<T> = IAction & {
   onRowClick?: (row: T) => Promise<boolean> | void;
 };
 
+type LoaderFunction<T> = (
+  first?: number,
+  max?: number,
+  search?: string
+) => Promise<T[]>;
+
 export type DataListProps<T> = Omit<
   TableProps,
   "rows" | "cells" | "onSelect"
 > & {
-  loader: (first?: number, max?: number, search?: string) => Promise<T[]>;
+  loader: T[] | LoaderFunction<T>;
   onSelect?: (value: T[]) => void;
   canSelectAll?: boolean;
   detailColumns?: DetailField<T>[];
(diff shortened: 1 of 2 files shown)


src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -121,11 +121,17 @@ export type Action<T> = IAction & {
...
+  max?: number,
+  search?: string
+) => Promise<T[]>;
+
 export type DataListProps<T> = Omit<
   TableProps,
   "rows" | "cells" | "onSelect"
 > & {
-  loader: (first?: number, max?: number, search?: string) => Promise<T[]>;
+  loader: T[] | LoaderFunction<T>;
```
jonkoops, line 134: Neat, that is quite useful.

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -121,11 +121,17 @@ export type Action<T> = IAction & {
   onRowClick?: (row: T) => Promise<boolean> | void;
 };
 
+type LoaderFunction<T> = (
```
jonkoops, line 124: It looks like some of the arguments here are always passed, perhaps we can make them required?

src/components/table-toolbar/KeycloakDataTable.tsx
```diff
@@ -302,7 +310,7 @@ export function KeycloakDataTable<T>({
       setRows(result);
       setLoading(false);
     },
-    [key, first, max, search]
+    [key, first, max, search, loader]
```
jonkoops, line 313: If `loader` is an unmemoized function this will trigger the effect every time the parent component is rendered.

jonkoops (changes requested):

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: no you can use the keycloak data table with a loader that has 0 arguments when do not use `isPaginated`

src/components/table-toolbar/KeycloakDataTable.tsx
edewit: right loader should only be a dependency when it's data

jonkoops (approved):

PR #2316, opened 2022-03-21
Revert "Bump react-flow-renderer from 9.7.4 to 10.0.3"

jonkoops:
Reverts the upgrade of react-flow-renderer, which is causing issues for the build.


agagancarczyk (approved):
lgtm

PR #2322, opened 2022-03-22
added service account user link

edewit:
fixes: #1285


edewit:
I didn't have a design for this @xianli123 so did my own interpretation, could you have a look:
image

xianli123:
@edewit This is a new feature. I'll look into it. If necessary, I'll make a design for you. What do you think?

edewit:
Yes I think so, see the old console it's there now as well

xianli123:
Hi @edewit I updated this feature based on your proposal and the relevant issues.  As shown in the screenshot below, there is a description that contains the user name. After clicking the name, there will be a modal pop-up to make users confirm this action. What do you think? 

@ssilvert Could you please take a look at this design?

Screen Shot 2022-03-23 at 4 49 33 PM

Screen Shot 2022-03-23 at 5 00 10 PM

ssilvert:
@xianli123 I think your proposal makes it very clear to the user.  In fact, I think it's good enough that we don't need the confirm dialog.

xianli123:
> @xianli123 I think your proposal makes it very clear to the user. In fact, I think it's good enough that we don't need the confirm dialog.

Ok, thanks. We can remove the confirm dialog.

edewit:
changed to design:
image

xianli123:
Looks good to me. Thanks.

jonkoops (approved):

PR #2325, opened 2022-03-22
Use 'del' to remove passing test videos

jonkoops:
Depending on how the tests are ran removing the video files can cause an exception (especially when running in parallel). This aligns with what the Cypress documentation [recommends](https://docs.cypress.io/guides/guides/screenshots-and-videos#Only-upload-videos-for-specs-with-failing-or-retried-tests).

Also includes changes to properly await the deletion of the files.

### DIFF
--- a/cypress/plugins/index.ts
+++ b/cypress/plugins/index.ts
@@ -1,8 +1,8 @@
 // The Webpack preprocessor does not include any types so it will have to be ignored.
 // @ts-ignore
 import webpackPreprocessor from "@cypress/webpack-batteries-included-preprocessor";
+import del from "del";
 import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
-import fs from "fs";
 import path from "path";
 
 // ***********************************************************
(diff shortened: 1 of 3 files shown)


cypress/plugins/index.ts
```diff
@@ -47,9 +48,9 @@ const configurePlugins: Cypress.PluginConfig = (on) => {
       attempts.some(({ state }) => state === "failed")
     );
 
-    // delete the video if the spec passed and no tests retried
+    // Delete the video if the spec passed and no tests were retried.
     if (!failures) {
-      fs.rmSync(results.video);
+      await del(results.video);
```
edewit, line 53: the documentation returns the promise:
```javascript
module.exports = (on, config) => {
  on('after:spec', (spec, results) => {
    if (results && results.video) {
      // Do we have failures for any retry attempts?
      const failures = _.some(results.tests, (test) => {
        return _.some(test.attempts, { state: 'failed' })
      })
      if (!failures) {
        // delete the video if the spec passed and no tests retried
        return del(results.video)
      }
    }
  })
}
```

edewit (changes requested):

edewit (approved):

Aboullos (approved):

PR #2329, opened 2022-03-23
added more missing fields

edewit:
fixes: #1385


jenny-s51 (approved):
LGTM @edewit ! Fields work as expected and data is written to the server correctly 👍

PR #2332, opened 2022-03-23
improved use of indexes making rendering faster

edewit:
@jonkoops optimised use of the indexes to make the rendering faster and more robust

### DIFF
--- a/src/components/multi-line-input/MultiLineInput.tsx
+++ b/src/components/multi-line-input/MultiLineInput.tsx
@@ -29,7 +29,7 @@ export const MultiLineInput = ({
 
   const value = watch(name, defaultValue);
   const [indexes, setIndexes] = useState(
-    Array.isArray(value) ? value.map((_, index) => index) : []
+    Array.isArray(value) ? value.map((_, index) => index) : [0]
   );
   const [counter, setCounter] = useState(0);
 
--- a/src/components/multi-line-input/MultiLineInput.tsx
+++ b/src/components/multi-line-input/MultiLineInput.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useEffect } from "react";
+import React, { Fragment, useState } from "react";
 import { useFormContext } from "react-hook-form";
 import {
   TextInput,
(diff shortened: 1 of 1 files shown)


src/components/multi-line-input/MultiLineInput.tsx
```diff
@@ -25,35 +25,34 @@ export const MultiLineInput = ({
   ...rest
 }: MultiLineInputProps) => {
   const { t } = useTranslation();
-  const { register, watch, setValue } = useFormContext();
+  const { register, watch } = useFormContext();
 
   const value = watch(name, defaultValue);
-  const fields = Array.isArray(value) && value.length !== 0 ? value : [""];
+  const [indexes, setIndexes] = useState<number[]>(
```
jonkoops, line 31: I think you can omit the type here as it should be inferrable.
```suggestion
  const [indexes, setIndexes] = useState(
```

src/components/multi-line-input/MultiLineInput.tsx
```diff
@@ -25,35 +25,34 @@ export const MultiLineInput = ({
...
+    setCounter((prevCounter) => prevCounter + 1);
   };
 
-  useEffect(() => register(name), [register]);
-
   return (
     <>
-      {fields.map((value: string, index: number) => (
+      {indexes.map((index) => (
         <Fragment key={index}>
```
jonkoops, line 51: Would this really prevent re-renders though? The `index` would always be the same :thinking:

jonkoops (changes requested):
I am not sure this would actually make the performance any different. Both before and now the `index` is still changed for all items that are after the point where an entry is removed so they will still re-render.

src/components/multi-line-input/MultiLineInput.tsx
edewit: The problem is removing a row and adding one again, then it would reuse the same number. Now it counts the total number of rows and this newly added row will not share the same row index as the removed one.

jonkoops: Ah great, then this makes sense.

jonkoops (approved):

PR #2339, opened 2022-03-25
another try to make the export  tab more stable

edewit:


cypress/integration/client_authorization_test.spec.ts
```diff
@@ -149,10 +149,11 @@ describe("Client authentication subtab", () => {
     authenticationTab.cancel();
   });
 
-  it("Should copy auth details", () => {
-    authenticationTab.goToExportSubTab();
+  it.only("Should copy auth details", () => {
+    authenticationTab.waitForExportLoad(() => {
+      authenticationTab.goToExportSubTab();
+    });
```
jonkoops, line 155: Do we need to use a callback here? I think you can just call this method and get the same result.
```suggestion
    authenticationTab.waitForExportLoad();
    authenticationTab.goToExportSubTab();
```

cypress/support/pages/admin_console/manage/clients/AuthorizationTab.ts
```diff
@@ -182,4 +182,12 @@ export default class AuthorizationTab {
...
     return this;
   }
+
+  waitForExportLoad(callback: () => void) {
+    cy.intercept(
+      "/admin/realms/master/clients/*/authz/resource-server/settings"
+    ).as("export-load");
+    callback();
+    cy.wait(["@export-load"]);
+  }
```
jonkoops, line 192: I don't think we need to have a callback here since Cypress should manage the command queue for us.
```suggestion
  waitForExportLoad() {
    cy.intercept(
      "/admin/realms/master/clients/*/authz/resource-server/settings"
    ).as("export-load");
    cy.wait(["@export-load"]);
  }
```

jonkoops (changes requested):

jonkoops:
The tests are passing again, awesome!

cypress/integration/client_authorization_test.spec.ts
edewit: no you have to intercept the call first before you can wait

jonkoops (approved):

PR #2340, opened 2022-03-25
Drop snapshot releases in favour of nightlies

jonkoops:
Since we are now publishing our releases on a nightly basis to OSSRH we no longer need to keep publishing snapshots to Github Packages. Note that the nightly releases are facilitated from [keycloak-rel/keycloak-rel](https://github.com/keycloak-rel/keycloak-rel) and not through this repository.

This PR removes the workflow to publish to Github Packages and changes the default repository for distribution to OSSRH.


ssilvert (approved):
LGTM

stianst (approved):

PR #2361, opened 2022-03-28
Disable Dependabot updates for react-flow-renderer

jonkoops:
The new version of `react-flow-renderer` has some breaking changes which will require a manual upgrade. For that reason, let's disable this dependency update for now.


edewit (approved):
:+1:

PR #2370, opened 2022-03-29
fixes empty key value convert

edewit:
Empty attributes `attributes: [{ key: "", value: "" }]` where not detected as such, this fixes that

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -74,9 +74,8 @@ const isAttributeArray = (value: any) => {
     return false;
   }
 
-  return (
-    value.filter((e) => Object.hasOwn(e, "key") && Object.hasOwn(e, "value"))
-      .length !== 0
+  return value.some(
+    (e) => Object.hasOwn(e, "key") && Object.hasOwn(e, "value")
   );
 };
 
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -73,7 +73,14 @@ const isAttributeArray = (value: any) => {
   if (!Array.isArray(value)) {
     return false;
   }
-  return value.filter((e) => e.key && e.value).length !== 0;
+
+  return (
+    value.filter(
+      (e) =>
+        Object.prototype.hasOwnProperty.call(e, "key") &&
+        Object.prototype.hasOwnProperty.call(e, "value")
```
jonkoops, line 81: Instead of using `hasOwnProperty()` you can use [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn), which is intended as a modern replacement.
```suggestion
        Object.hasOwn(e, "key") &&
        Object.hasOwn(e, "value")
```

jonkoops (changes requested):

src/util.ts
edewit: Okay didn't know about that... let's do that

jonkoops (approved):

PR #2395, opened 2022-04-04
Replace development server with nightly version of Keycloak

jonkoops:
This allows us to develop locally against the new [Nightly releases](https://github.com/keycloak/keycloak/releases/tag/nightly) of Keycloak. Also attempts to improve the development section of the readme.


scripts/start-server.mjs
```diff
@@ -0,0 +1,87 @@
...
+  console.info("Downloading and extracting server…");
+
+  const nightlyAsset = await getNightlyAsset();
+  const assetStream = await getAssetAsStream(nightlyAsset);
+
+  await extractTarball(assetStream, SERVER_DIR, { strip: 1 });
+}
+
+async function getNightlyAsset() {
+  const api = new Octokit();
```
edewit, line 63: no more nice progress bars?

edewit:
Although this might be a cool idea in practice it's a bit more complicated say for instance the keycloak changes the way the `redirect_uri` is passed for logging out. `keycloak.js` is also changed to reflect this change. This will only later be released as we use a released version of `keyloak.js` we are then out of sync with the server and our build brakes

scripts/start-server.mjs
jonkoops: No, unfortunately the streams we're using do not give their length so a progress cannot be calculated.

jonkoops:
@edewit I agree with that, but I'd rather have these issues visible during development so we can deal with them early on rather than retroactively. For now this problem has been mitigated by skipping the tests that would have otherwise failed.

The real solution here would be to use a nightly version of Keycloak JS as well.

edewit (approved):

PR #2410, opened 2022-04-08
Realm-settings -> User Profile -> Create/Edit Attribute tests

agagancarczyk:
## Motivation
Cypress tests for creating and editing attributes.

Screenshot 2022-04-08 at 10 38 25

## Verification Steps
1. Running realm_settings_user_profile_tab.spec.ts tests locally. 

## Checklist:


src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
```diff
@@ -41,6 +41,8 @@ export const AttributeGeneralSettings = () => {
     useState(false);
   const { attributeName } = useParams<AttributeParams>();
   const editMode = attributeName ? true : false;
+  const nameInput = document.getElementById("kc-attribute-name");
+  nameInput?.focus();
```
jonkoops, line 45: This should be wrapped in a `useEffect()`

jonkoops (changes requested):

src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
```diff
@@ -41,6 +41,8 @@ export const AttributeGeneralSettings = () => {
     useState(false);
   const { attributeName } = useParams<AttributeParams>();
   const editMode = attributeName ? true : false;
+  const nameInput = document.getElementById("kc-attribute-name");
```
jonkoops, line 44: Instead of using `getElementById()` you can use a `Ref` (see [React documentation](https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components)).

jonkoops (changes requested):

src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
edewit: why would you want to focus the name attribute? you could also solve it by adding the `autofocus` attirbute to the input field

agagancarczyk: thanks @edewit , just added a fix into AttributeInput so that the Annotations key input field doesn't focus on the page load.

src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
agagancarczyk: This isn't required anymore.

jonkoops:
Changes look good like this, let's see if the tests will pass.

Issue #812, opened 2021-07-08
Update react-hook-form v7


jonkoops:
More information can be found in the [migration guide](https://react-hook-form.com/migrate-v6-to-v7/). This is quite a large update as the API surface has changed significantly.

PR #2413, opened 2022-04-08
Replace text inputs with KeycloakTextInput

jonkoops:
The default `TextInput` from PatternFly modifies the `onChange` handler so the value can be more easily accessed. This change in the function signature is however incompatible with what React Hook Form expects. This PR therefore introduces a `KeycloakTextInput` that restores the function signature, fixing the integration.

This change is needed as part of moving towards updating React Hook Form to version 7 (#812).


edewit (approved):
:+1:

agagancarczyk (approved):
Changes look good, but almost all tests are now failing. @jonkoops

jonkoops:
@agagancarczyk I see, I'll take a look at those.

PR #2435, opened 2022-04-11
now also removes the already added validators after closing the dialog

edewit:
Used to only remove the validators as long as the dialog stayed open
now also removes the already installed validators when the dialog
is opened again.

fixes: #2397

### DIFF
--- a/src/realm-settings/user-profile/attribute/AddValidatorDialog.tsx
+++ b/src/realm-settings/user-profile/attribute/AddValidatorDialog.tsx
@@ -9,22 +9,30 @@ import {
   Thead,
   Tr,
 } from "@patternfly/react-table";
+
+import type { KeyValueType } from "../../../components/attribute-form/attribute-convert";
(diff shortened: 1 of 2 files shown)


agagancarczyk (approved):

src/realm-settings/user-profile/attribute/AddValidatorDialog.tsx
```diff
@@ -9,22 +9,30 @@ import {
...
 
 export const AddValidatorDialog = ({
+  selectedValidators,
   toggleDialog,
   onConfirm,
 }: AddValidatorDialogProps) => {
   const { t } = useTranslation("realm-settings");
   const [selectedValidator, setSelectedValidator] = useState<Validator>();
-  const [validators, setValidators] = useState(allValidator);
+  const [validators, setValidators] = useState(
```
jonkoops, line 31: If you use a callback here the initial state will only be computed once instead of every render.

jonkoops (changes requested):

jonkoops (approved):

PR #2449, opened 2022-04-13
fixed error when data not found

edewit:
fixes: #2447

### DIFF
--- a/src/clients/authorization/KeyBasedAttributeInput.tsx
+++ b/src/clients/authorization/KeyBasedAttributeInput.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { Controller, useFieldArray, useFormContext } from "react-hook-form";
 import {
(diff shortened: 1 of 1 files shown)


src/clients/authorization/KeyBasedAttributeInput.tsx
```diff
@@ -58,26 +58,33 @@ const ValueInput = ({
   const { control, register, getValues } = useFormContext();
 
   const [isValueOpenArray, setIsValueOpenArray] = useState([false]);
+  const [attributeValues, setAttributeValues] = useState<
+    { key: string; name: string }[] | undefined
```
jonkoops, line 62: Can we split this into a separate type?

src/clients/authorization/KeyBasedAttributeInput.tsx
```diff
@@ -58,26 +58,33 @@ const ValueInput = ({
   const { control, register, getValues } = useFormContext();
 
   const [isValueOpenArray, setIsValueOpenArray] = useState([false]);
+  const [attributeValues, setAttributeValues] = useState<
```
jonkoops, line 61: Since this value is derived it can be a `useMemo()` instead of a `useState()`.

jonkoops (changes requested):

jonkoops (approved):

ssilvert:
Partial import test didn't pass, but I know this is not a regression.  So I merged.

Issue #2403, opened 2022-04-06
Restrict screens based on Fine-grained Permissions


ssilvert:
### Description

This issue is only for the permission restrictions portion of [Fine Grained Permissions](https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions) functionality.

We need the UI to access based on fine grained permissions.  This means making certain parts of the UI hidden or read-only based on those permissions.

For entities like Users and Clients, you will see that the server returns an access object that tells the UI what is allowed.  

Other parts of fine-grained permissions are handled on the server side.  For instance, if a user is only allowed to manage one client then a query for all clients will only return a single-element array containing that one client.

Updates to the screens that allow a user to set up fine-grained permissions is a separate task handled by a [separate issue](https://github.com/keycloak/keycloak-admin-ui/issues/2402).  We don't need those screens to start on this task because configuration can be done in the old UI.

ssilvert:
Also see #1943

Issue #2402, opened 2022-04-06
Fine-grained Permissions Screens


ssilvert:
### Description

This issue is only for the screens portion of [Fine Grained Permissions](https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions) functionality.  

We need screens that are similar to what we have in the old Admin UI so that the user can configure fine grained permissions.

Updates to the UI that restrict user access based on these permissions is a separate task handled by [a separate issue](https://github.com/keycloak/keycloak-admin-ui/issues/2403).

xianli123:
Hi @ssilvert I am researching this feature based on the documentation and recorded some info in a doc. Here is the link: https://docs.google.com/document/d/1nGWshTYK5CQajgVesyUOcG2_WOeO65PfjLnxKp4lQwA/edit?usp=sharing  It includes some questions that need to be answered by developers.

PR #2451, opened 2022-04-13
Fine-grained permissions for user section.

ssilvert:
## Motivation
This implements fine-grained permissions for the user section only.

Implements part of https://github.com/keycloak/keycloak-admin-ui/issues/2403

## Brief Description
This is a fairly simple approach.  In some cases we could do better with how we present a restricted screen to the user.  But overall it's on par with the implementation in the old UI.  So I think this approach is good for a first pass and we can improve it later.  Higher priority is to get all of FGP implemented and then create an FGP test suite.

## Verification Steps
Since the [fine-grained permissions configuration screens](https://github.com/keycloak/keycloak-admin-ui/issues/2402) are not implemented in the new UI yet, verification requires going to the old UI.  You can follow the [examples in the documentation](https://www.keycloak.org/docs/latest/server_admin/#_fine_grain_permissions).

You can also see the access object that is returned in the user record when a user is queried.  It looks something like this:
`"access":{"manageGroupMembership":false,"view":true,"mapRoles":true,"impersonate":true,"manage":false}`

This will give hints as to what the user should be allowed to do in the UI.  Note that sometimes you also need the proper roles to be set on the realm-management client.

Lastly, make sure your permissions are set on a test realm.  The user restricted by the fine-grained access should log in to the test realm.  Using snowpack, you will need to go to that realm like this localhost:8080/?realm=test

### DIFF
--- a/src/components/key-value-form/AttributeForm.tsx
+++ b/src/components/key-value-form/AttributeForm.tsx
@@ -16,9 +16,15 @@ export type AttributesFormProps = {
   form: UseFormMethods<AttributeForm>;
   save?: (model: AttributeForm) => void;
   reset?: () => void;
+  fineGrainedAccess?: boolean;
 };
 
-export const AttributesForm = ({ form, reset, save }: AttributesFormProps) => {
+export const AttributesForm = ({
+  form,
+  reset,
+  save,
+  fineGrainedAccess,
+}: AttributesFormProps) => {
   const { t } = useTranslation("roles");
   const noSaveCancelButtons = !save && !reset;
   const {
@@ -30,6 +36,7 @@ export const AttributesForm = ({ form, reset, save }: AttributesFormProps) => {
     <FormAccess
       role="manage-realm"
       onSubmit={save ? handleSubmit(save) : undefined}
+      fineGrainedAccess={fineGrainedAccess}
     >
       <FormProvider {...form}>
         <KeyValueInput name="attributes" />
--- a/src/user/UserAttributes.tsx
+++ b/src/user/UserAttributes.tsx
@@ -56,6 +56,7 @@ export const UserAttributes = ({ user: defaultUser }: UserAttributesProps) => {
       <AttributesForm
         form={form}
         save={save}
+        fineGrainedAccess={user.access?.manage}
         reset={() =>
           form.reset({
             attributes: convertAttributes(),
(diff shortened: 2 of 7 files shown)


jonkoops (approved):

edewit:
If we need the user for these fine grained permissions, I think it would be better to put it on the context instead of passing it in. WDYT?

jonkoops:
I agree with @edewit, can we get that change in @ssilvert?

ssilvert:
> If we need the user for these fine grained permissions, I think it would be better to put it on the context instead of passing it in. WDYT?

@edewit I'm not sure what you are referring to.  I never needed to pass the user around as it was already available.

Also, it's been 7 days so it is time to merge.  Perhaps we merge and then you submit a PR with desired refactoring?

src/components/attribute-form/AttributeForm.tsx
```diff
@@ -16,9 +16,15 @@ export type AttributesFormProps = {
...
   reset?: () => void;
+  fineGrainedAccess?: boolean;
 };
 
-export const AttributesForm = ({ form, reset, save }: AttributesFormProps) => {
+export const AttributesForm = ({
+  form,
+  reset,
+  save,
+  fineGrainedAccess,
```
edewit, line 26: Instead of passing this in it would be better to have a context for User or extend the `whoAmI` context with the `fineGrainedAccess`

edewit (changes requested):

src/components/attribute-form/AttributeForm.tsx
ssilvert: It wouldn't make sense to put this in the whoAmI.  Whenever you query for users or clients or whatever, Keycloak puts the access object on that entity (See verification steps above).   The access object says, "The user who made the request has this access to this user/client/group/etc."  The access granted can be different for each entity.  So I might have full manage access for this client but only view access for another client.  The UI needs to look inside the client object to see the access and set the UI accordingly.  You won't know until after you do the query, so it can't go into the whoAmI.

So for this PR, we are looking at managing users.  But AttributeForm shouldn't know anything about users.  It only knows about attributes and whether or not it should display as read-only.  So we pass in fineGrainedAccess to govern that.

edewit (approved):

PR #2490, opened 2022-04-22
with keycloak-js updated logout works again

edewit:
Updated keycloak-js on the admin client to version 18 so that the redirect url problem is fixed and we can enable the logout tests again.


jonkoops (approved):

jonkoops:
@edewit can you rebase this PR on `main`?

jonkoops (approved):

PR #2501, opened 2022-04-25
added user permission tab

edewit:
added permission tab to users and made it more look like [the design](https://marvelapp.com/prototype/6gd34hh/screen/86305765)


src/components/permission-tab/PermissionTab.tsx
```diff
@@ -0,0 +1,218 @@
...
+          clientId: realm,
+        }),
+        (() => {
+          switch (type) {
+            case "clients":
+              return adminClient.clients.listFineGrainPermissions({ id: id! });
+            case "users":
+              return adminClient.realms.getUsersManagementPermissions({
+                realm,
+              }) as unknown as ManagementPermissionReference;
```
jonkoops, line 79: Is this type missing in the admin client? If so, can we fix this type there first?

src/components/permission-tab/PermissionTab.tsx
```diff
@@ -0,0 +1,218 @@
...
+      case "clients":
+        return adminClient.clients.updateFineGrainPermission(
+          { id: id! },
+          { enabled }
+        );
+      case "users":
+        return adminClient.realms.updateUsersManagementPermissions({
+          realm,
+          enabled,
+        }) as unknown as ManagementPermissionReference;
```
jonkoops, line 61: Is this type missing in the admin client? If so, can we fix this type there first?

src/components/permission-tab/PermissionTab.tsx
```diff
@@ -0,0 +1,218 @@
...
+export const PermissionsTab = ({ id, type }: PermissionsTabProps) => {
+  const { t } = useTranslation("common");
+  const history = useHistory();
+  const adminClient = useAdminClient();
+  const { realm } = useRealm();
+  const { whoAmI } = useWhoAmI();
+  const [realmId, setRealmId] = useState("");
+  const [permission, setPermission] = useState<ManagementPermissionReference>();
+
+  const togglePermissionEnabled = (enabled: boolean) => {
```
jonkoops, line 50: Perhaps instead of kind of duplicating this logic here you could add an `enabled` state that could be a dependency of the `useFetch()`. That way it could automatically refresh the data when this dependency changes.

src/components/permission-tab/PermissionTab.tsx
```diff
@@ -0,0 +1,218 @@
...
+                      {t("permissionsScopeName")}
+                    </Th>
+                    <Th id="description">{t("description")}</Th>
+                  </Tr>
+                </Thead>
+                <Tbody>
+                  {Object.entries(permission.scopePermissions || {})
+                    .sort((a, b) =>
+                      a[0]!.localeCompare(b[0]!, whoAmI.getLocale())
+                    )
```
jonkoops, line 171: Starting to think we are using this pattern often enough to justify putting it in it's own hook, but perhaps out of scope for this PR.

jonkoops (changes requested):

src/components/permission-tab/PermissionTab.tsx
edewit: there is already a PR that addresses this: https://github.com/keycloak/keycloak-nodejs-admin-client/pull/463

src/components/permission-tab/PermissionTab.tsx
edewit: what do you mean with duplicating logic in `useFetch` it fetches and here it updates

jonkoops: Ahhh, yes. I missed that one.

jonkoops (approved):

Issue #2381, opened 2022-04-01
"Group" field is blank for existing hardcoded-ldap-group-mapper


flipp5b:
### Describe the bug

We have an LDAP mapper of type hardcoded-ldap-group-mapper with a specified "group" value. In the new admin UI, the "group" field is displayed as empty.

### Version

17.0.1

ssilvert:
Thanks for reporting this @flipp5b. 

I have verified the bug.

PR #2506, opened 2022-04-26
convert single array values to values

edewit:
fixes: #2381

### DIFF
--- a/src/util.test.ts
+++ b/src/util.test.ts
@@ -116,15 +116,15 @@ describe("Tests the form convert util functions", () => {
     const given = {
       config: { group: ["one"], another: { nested: ["value"] } },
     };
-    const values: { [index: string]: any } = {};
-    const spy = (name: string, value: any) => (values[name] = value);
+    const setValue = jest.fn();
 
     //when
-    convertToFormValues(given, spy);
+    convertToFormValues(given, setValue);
 
     //then
-    expect(values).toEqual({
-      config: { group: "one", another: { nested: "value" } },
+    expect(setValue).toHaveBeenCalledWith("config", {
+      group: "one",
+      another: { nested: "value" },
     });
   });
 });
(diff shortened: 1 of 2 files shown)


src/util.test.ts
```diff
@@ -111,4 +111,20 @@ describe("Tests the form convert util functions", () => {
...
+    const values: { [index: string]: any } = {};
+    const spy = (name: string, value: any) => (values[name] = value);
+
+    //when
+    convertToFormValues(given, spy);
+
+    //then
+    expect(values).toEqual({
+      config: { group: "one", another: { nested: "value" } },
+    });
```
jonkoops, line 128: I think this test could be simplified a bit by only testing the contract of the `setValue` callback.
```suggestion
    const given = {
      config: { group: ["one"], another: { nested: ["value"] } },
    };
    const setValue = jest.fn();

    //when
    convertToFormValues(given, setValue);

    //then
    expect(setValue).toHaveBeenCalledWith("config", {
      group: "one",
      another: { nested: "value" },
    });
```

src/util.ts
```diff
@@ -89,7 +89,15 @@ export const convertToFormValues = (
     if (key === "attributes" && isAttributesObject(value)) {
       setValue(key, arrayToKeyValue(value as Record<string, string[]>));
     } else if (key === "config" || key === "attributes") {
-      setValue(key, !isEmpty(value) ? unflatten(value) : undefined);
+      if (!isEmpty(value)) {
```
jonkoops, line 92: This function is getting quite complex and nested, perhaps it could be optimized?

jonkoops (changes requested):

src/util.ts
edewit: We could change the backend to not give back different values, sometimes arrays are use with only one element and sometimes it's a string

jonkoops (approved):

Issue #2504, opened 2022-04-22
User page not loaded in admin console v2 having only roles "view-users, manage-users"


joggeli34:
### Describe the bug

When using a user which has the roles from realm-management "manage-users" and "view-users" the user-page is not loaded in the admin-console v2. Only a spinner is shown.

The group-page is shown and the groups are listed, but they can not be edited. Also no new group can be created. 

![image](https://user-images.githubusercontent.com/34470368/164741558-c051b06b-7987-4659-9a27-c47e0dbbcb3d.png)

With the old admin-console, it works.

### Version

18.0.0-legacy

### Expected behavior

With the two roles, the user should be shown and be editable, same for the groups.

### Actual behavior

Users are not shown, groups are not editable.

### How to Reproduce?

Create a user only having the client-roles "manage-users" and "view-users" from the realm-management

ssilvert:
@edewit I tracked this one down to UsersSection.tsx.  If the user does not have "view-realm" access then the call to /components?type=org.keycloak.storage.UserStorageProvider will fail.

It looks like you are using that call to figure out if there is a federation provider or not.  I suggest that if the user does not have "view-realm" access then just drop back to the behavior where we make the administrator query for users.

While you are in that code, I suggest that you add to the Empty State message.  It currently says, "This realm has a federated provider. Viewing all users may cause the system to slow down. Please search for a user above."  Suggest change to "This realm has a federated provider. Viewing all users may cause the system to slow down, but it can be done by searching for "*". Please search for a user above."

edewit:
@joggeli34 That you can view groups is because view-users is a composite from view-groups

PR #2515, opened 2022-04-27
no spinner when no have rights for realm load

edewit:
fixes: #2504

### DIFF
--- a/public/resources/en/users.json
+++ b/public/resources/en/users.json
@@ -4,7 +4,7 @@
   "userList": "User list",
   "searchForUser": "Search user",
   "startBySearchingAUser": "Start by searching for users",
-  "searchForUserDescription": "This realm has a federated provider. Viewing all users may cause the system to slow down. Please search for a user above.",
+  "searchForUserDescription": "This realm has a federated provider. Viewing all users may cause the system to slow down, but it can be done by searching for \"*\". Please search for a user above.",
   "createUser": "Create user",
   "createNewUser": "Create new user",
   "noUsersFound": "No users found",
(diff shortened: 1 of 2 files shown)


jonkoops (approved):

ssilvert (approved):
Tested manually.  Looks good.

ssilvert (changes requested):
Approved too soon.

With this code, if you have a federation provider, it does not prompt you to search for a user.

edewit:
Yes, how can it know you have user federation setup if you don't have rights to query the user federation?

ssilvert:
> Yes, how can it know you have user federation setup if you don't have rights to query the user federation?

The user federation still gets queried on the back end when you do a user query.  As long as you have "query-users" rights, it will query users in the user federation.  

The problem is that you don't have access to the component that is letting you find out if a federated provider exists or not unless you have "view-realm" access.

That's why I suggested that when you don't have "view-realm" rights we can just fall back to the behavior where we tell the user to do a search.

jonkoops (approved):

public/resources/en/users.json
```diff
@@ -4,7 +4,7 @@
   "userList": "User list",
   "searchForUser": "Search user",
   "startBySearchingAUser": "Start by searching for users",
-  "searchForUserDescription": "This realm has a federated provider. Viewing all users may cause the system to slow down. Please search for a user above.",
+  "searchForUserDescription": "This realm has a federated provider. Viewing all users may cause the system to slow down, but it can be done by searching for \"*\". Please search for a user above.",
```
ssilvert, line 7:
```suggestion
  "searchForUserDescription": "This realm may have a federated provider. Viewing all users may cause the system to slow down, but it can be done by searching for \"*\". Please search for a user above.",
```

PR #2522, opened 2022-04-28
small fix to prevent empty dialog

edewit:
this also makes the re-render better

fixes: #2521

### DIFF
--- a/src/clients/scopes/AddScopeDialog.tsx
+++ b/src/clients/scopes/AddScopeDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
 import {
   Button,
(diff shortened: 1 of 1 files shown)


src/clients/scopes/AddScopeDialog.tsx
```diff
@@ -77,18 +76,18 @@ export const AddScopeDialog = ({
     useToggle(false);
 
   useEffect(() => {
```
jonkoops, line 78: Instead if using an effect to set the state I believe this could be a memoized value. This also allows you to reduce the nesting of this logic by using early returns.

```ts
const clientScopes = useMemo(() => {
  if (filterType === FilterType.Name) {
    return scopes;
  }

  if (protocolType === ProtocolType.OpenIDConnect) {
    return scopes.filter((item) => item.protocol === "openid-connect");
  }

  if (protocolType === ProtocolType.SAML) {
    return scopes.filter((item) => item.protocol === "saml");
  }

  return scopes;
}, [scopes, filterType, protocolType]);
```

jonkoops (changes requested):

jonkoops (approved):

PR #2527, opened 2022-04-28
added role permission tab

edewit:
fixes: #2402

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
       "name": "keycloak-admin-ui",
       "license": "Apache",
       "dependencies": {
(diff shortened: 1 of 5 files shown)


ssilvert (changes requested):
When I go to the "Permissions" tab for a realm role or client role I get an error: 

`PermissionTab.js:108 Uncaught TypeError: adminClient.roles.listPermissions is not a function`

`The above error occurred in the <PermissionsTab> component:`

ssilvert:
> @ssilvert you need to do an `npm install` as package.json has a new version of the adminClient

Yea.  Just realized that after I posted.

ssilvert (approved):
LGTM

PR #2559, opened 2022-05-02
Remove unused 'cli-progress' dependency

jonkoops:


edewit (approved):
I'm a bit sad about it, but :+1:

jonkoops:
Yeah, I'd have liked to see it stay as well. Perhaps we can contribute to the packages we use so the streams can report their size, that way we can actually have it report the download, extraction and writing progress.

Issue #2528, opened 2022-04-28
Clients-->Authorization-->Permission details shows permissions in Policies dropdown


ssilvert:
### Describe the bug

The Policies dropdown should only contain policies.

In this example, the dropdown should only contain the "admin only" policy.  Instead, it also contains all the permissions for the client.

<img width="801" alt="Policies" src="https://user-images.githubusercontent.com/507202/165830275-e6cb1ff5-a20d-434c-aed4-a6da69a4fcec.png">

### Version

18.0.0-SNAPSHOT

edewit:
yes, I think this is correct as there is no way to differentiate between "normal" polices and these generated once

ssilvert:
@edewit The dropdown should only contain policies.  But it is showing permissions as well.

Those things that look like "manage.members.permission.group.27b1d4ed-4721-404e-8a49-a2923bcf7505" are permissions.  They shouldn't be there.

Also, to really fix this properly, we will need to do what the old console does.  It allows searching because the list of policies can get really long.  But I'm OK if we don't get that done before milestone 19.

edewit:
but you can search by typing in the field:
![image](https://user-images.githubusercontent.com/51133/166411687-7b96dd66-2f1c-4729-b283-64f37b156b7f.png)

edewit:
the results shown in the dropdown come from a policy query: `/admin/realms/{realm}/clients/{clientId}/authz/resource-server/policy?first=0&max=10`

PR #2564, opened 2022-05-03
add permission: false to policy query

edewit:
Without this the policy endpoint will return policies and permissions we only want policies in this dropdown.

fixes: #2528


jonkoops (approved):

ssilvert (approved):
Tested manually.  Looks good.

PR #2570, opened 2022-05-04
use flow type instead of flow name

edewit:
Should have been part of the initial change, but it seems I didn't test this enough

### DIFF
--- a/src/authentication/AuthenticationSection.tsx
+++ b/src/authentication/AuthenticationSection.tsx
@@ -73,9 +73,9 @@ export default function AuthenticationSection() {
       throw new Error(t("common:notFound"));
     }
 
-    const defaultFlows = Object.entries(realmRep)
-      .filter((entry) => REALM_FLOWS.includes(entry[0]))
-      .map((entry) => entry[1]);
+    const defaultFlows = Object.entries(realmRep).filter((entry) =>
+      REALM_FLOWS.includes(entry[0])
+    );
 
     for (const flow of flows as AuthenticationType[]) {
       flow.usedBy = { values: [] };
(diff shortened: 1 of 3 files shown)


src/authentication/AuthenticationSection.tsx
```diff
@@ -73,9 +73,9 @@ export default function AuthenticationSection() {
       throw new Error(t("common:notFound"));
     }
 
-    const defaultFlows = Object.entries(realmRep)
-      .filter((entry) => REALM_FLOWS.includes(entry[0]))
-      .map((entry) => entry[1]);
+    const defaultFlows = Object.entries(realmRep).filter((entry) =>
+      REALM_FLOWS.includes(entry[0])
+    );
```
jonkoops, line 78: This value could also be destructured.
```suggestion
    const defaultFlows = Object.entries(realmRep).filter(([key]) =>
      REALM_FLOWS.includes(key)
    );
```

src/authentication/AuthenticationSection.tsx
```diff
@@ -101,10 +101,10 @@ export default function AuthenticationSection() {
         flow.usedBy.values = idps.map(({ alias }) => alias!);
       }
 
-      const isDefault = defaultFlows.includes(flow.alias);
+      const isDefault = defaultFlows.find(([, alias]) => flow.alias === alias);
```
jonkoops, line 104: The `isX` pattern implies this variable contains a boolean value, but it contains a different value. Perhaps something like the following would make sense?

```ts 
const defaultAlias = defaultFlows
  .map(([, alias]) => alias)
  .find((alias) => flow.alias === alias);
```

jonkoops (changes requested):

jonkoops (approved):

PR #2573, opened 2022-05-05
added tool tips on flow icons

edewit:


src/authentication/components/EditFlowDropdown.tsx
```diff
@@ -40,57 +45,62 @@ export const EditFlowDropdown = ({
...
-          name={execution.displayName!}
-          type={type}
-          onSelect={(type) => {
-            if (type) {
-              onAddExecution(execution, type);
-            }
-            setType(undefined);
-          }}
+    <Tooltip content={t("common:edit")}>
+      <>
```
jonkoops, line 49: I think you can get rid of this fragment.

jonkoops (changes requested):

src/authentication/components/EditFlowDropdown.tsx
edewit: No `Tooltip` can have only one child

jonkoops: Oh, yeah that's a bug in PatternFly. Consider this resolved.

jonkoops (approved):

PR #2614, opened 2022-05-10
issue 2580

edewit:
- added missing translations
- small wait in between load and delete query

fixes: #2580

### DIFF
--- a/src/components/role-mapping/RoleMapping.tsx
+++ b/src/components/role-mapping/RoleMapping.tsx
@@ -174,7 +174,7 @@ export const RoleMapping = ({
   const [hide, setHide] = useState(false);
   const [showAssign, setShowAssign] = useState(false);
   const [selected, setSelected] = useState<Row[]>([]);
-  const [wait, toggle] = useToggle();
+  const [wait, toggleWait] = useToggle();
   const setTimeout = useSetTimeout();
 
   const assignRoles = async (rows: Row[]) => {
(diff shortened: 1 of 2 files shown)


src/components/role-mapping/RoleMapping.tsx
```diff
@@ -172,12 +174,16 @@ export const RoleMapping = ({
   const [hide, setHide] = useState(false);
   const [showAssign, setShowAssign] = useState(false);
   const [selected, setSelected] = useState<Row[]>([]);
+  const [wait, toggle] = useToggle();
```
jonkoops, line 177: Might be worth it to add what is being toggled to the name.
```suggestion
  const [wait, toggleWait] = useToggle();
```

src/components/role-mapping/RoleMapping.tsx
```diff
@@ -172,12 +174,16 @@ export const RoleMapping = ({
...
+  const setTimeout = useSetTimeout();
 
   const assignRoles = async (rows: Row[]) => {
     await save(rows);
     refresh();
   };
 
+  useEffect(() => setTimeout(refresh, 200), [wait]);
```
jonkoops, line 185: Interesting, why do you need to do this on a timeout?

jonkoops (approved):

src/components/role-mapping/RoleMapping.tsx
edewit: somehow it gives me the old values

jonkoops: Strange, I guess it still takes a while before the transaction is done.

PR #2625, opened 2022-05-11
fixed the group move test by fixing the app

edewit:
## Motivation
There was a comment on the group move test that the test didn't work.
In fact, the app didn't work fixed the test and the app.
Now it tests 2 scenarios moving a group to the root and moving it back.


edewit:
@Aboullos ahh yes sorry there where some errors in the types... fixed them now to

jonkoops (approved):

PR #2632, opened 2022-05-12
Fix role-based authorization in Clients section.

ssilvert:
## Motivation
The PR fixes role-based authorization in the Clients section by hiding components or making them read-only as needed.

## Brief Description
This is in preparation for implementing fine-grained authorization, where flags such as `isManager` will also include a check for fine-grained authorization.

## Verification Steps
To verify the PR, it is best to log into a test realm with a test user.  Then set limited client roles for that user on the realm-management client.

To see the Clients section the user should have "view-clients" and "query-clients" roles.
To see the Permissions tab, the user also needs "manage-authorization" role.
To see Keys and Credentials tabs, the client needs to have `Client authentication` enabled.
To see Authorization tab, the client needs to have `Authorization` enabled.
To see Service account roles tab, the client needs to have `Service account roles` enabled.

### DIFF
--- a/src/clients/ClientsSection.tsx
+++ b/src/clients/ClientsSection.tsx
@@ -149,7 +149,7 @@ export default function ClientsSection() {
               isPaginated
               ariaLabelKey="clients:clientList"
               searchPlaceholderKey="clients:searchForClient"
-              toolbarItem={ToolbarItems()}
+              toolbarItem={<ToolbarItems />}
               actionResolver={(rowData: IRowData) => {
                 const client: ClientRepresentation = rowData.data;
                 const actions: Action<ClientRepresentation>[] = [
--- a/src/clients/ClientsSection.tsx
+++ b/src/clients/ClientsSection.tsx
@@ -32,6 +32,7 @@ import { toClient } from "./routes/Client";
 import { toImportClient } from "./routes/ImportClient";
 import { isRealmClient, getProtocolName } from "./utils";
 import helpUrls from "../help-urls";
+import { useAccess } from "../context/access/Access";
 
 export default function ClientsSection() {
   const { t } = useTranslation("clients");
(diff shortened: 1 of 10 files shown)


src/clients/authorization/Settings.tsx
```diff
@@ -166,6 +166,7 @@ export const AuthorizationSettings = ({ clientId }: { clientId: string }) => {
           name="authenticationSettings"
           save={() => handleSubmit(save)()}
           reset={() => reset(resource)}
+          isActive={true}
```
jonkoops, line 169: For boolean props that are true the value can be omitted.
```suggestion
          isActive
```

src/components/role-mapping/RoleMapping.tsx
```diff
@@ -253,36 +279,23 @@ export const RoleMapping = ({
...
-              return false;
-            },
-          },
-        ]}
+        actions={
+          isManager
+            ? [
+                {
+                  title: t("common:unAssignRole"),
+                  onRowClick: async (role) => {
```
jonkoops, line 290: There is nothing asynchronous happening here, so I believe you can omit the `async` keyword.
```suggestion
                  onRowClick: (role) => {
```

jonkoops (changes requested):

src/components/role-mapping/RoleMapping.tsx
ssilvert: That was not my code, but I think whoever wrote it was correct.  When I remove the `async` I get `Type 'boolean' is not assignable to type 'void | Promise<boolean>`

src/clients/authorization/Settings.tsx
ssilvert: fixed

src/clients/ClientsSection.tsx
```diff
@@ -117,29 +149,7 @@ export default function ClientsSection() {
...
-                        <Link {...props} to={toImportClient({ realm })} />
-                      )}
-                      variant="link"
-                    >
-                      {t("importClient")}
-                    </Button>
-                  </ToolbarItem>
-                </>
-              }
+              toolbarItem={ToolbarItems()}
```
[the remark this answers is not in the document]
ssilvert: fixed

src/components/role-mapping/RoleMapping.tsx
jonkoops: Ah, yes. That's fine then.

jonkoops (approved):

PR #2635, opened 2022-05-13
throw error when server context can't be loaded

edewit:
fixes: #2631

## Checklist:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,9 +41,7 @@ const AppContexts: FunctionComponent<AdminClientProps> = ({
             <AccessContextProvider>
               <Help>
                 <AlertProvider>
-                  <ServerInfoProvider>
-                    <SubGroups>{children}</SubGroups>
-                  </ServerInfoProvider>
+                  <SubGroups>{children}</SubGroups>
                 </AlertProvider>
               </Help>
             </AccessContextProvider>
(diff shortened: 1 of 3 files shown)


src/context/server-info/ServerInfoProvider.tsx
```diff
@@ -1,30 +1,40 @@
...
-  const [serverInfo, setServerInfo] = useState<ServerInfoRepresentation>({});
+  const [serverInfo, setServerInfo] = useState<ServerInfoRepresentation>();
 
   useFetch(
     async () => {
       try {
         return await adminClient.serverInfo.find();
       } catch (error) {
-        return {};
+        return undefined;
```
jonkoops, line 37: This whole statement can be removed since the implicit return value of any function is `undefined` already.
```suggestion
```

jonkoops (approved):

src/context/server-info/ServerInfoProvider.tsx
edewit: Don't you think that making it explicit is better?

src/App.tsx
```diff
@@ -91,16 +89,18 @@ export const App = ({ adminClient }: AdminClientProps) => {
...
-            {routes.map((route, i) => (
-              <Route
-                exact={route.matchOptions?.exact ?? true}
-                key={i}
-                path={route.path}
-                component={() => <SecuredRoute route={route} />}
-              />
-            ))}
-          </Switch>
+          <ServerInfoProvider>
```
jonkoops, line 92: This was already provided in the `AppContexts` component, perhaps this can be removed and moved up in the hierarchy there to accomplish the same?

jonkoops (approved):

src/context/server-info/ServerInfoProvider.tsx
jonkoops: Nope

src/App.tsx
edewit: No it has to be inside the `<ErrorBoundary` otherwise the 403 will just get ignored and the loading indicator will not go away

jonkoops: Ahhhh, yeah that makes sense.

PR #2655, opened 2022-05-17
Add 'Session' tabs to user and client details

jonkoops:
## Motivation
During the conversion of the 'Managing user sessions' section of the documentation it was found that some of these features were missing. Namely the ability to see sessions for a [specific client](https://www.keycloak.org/docs/latest/server_admin/#application-navigation), or the sessions of a [specific user](https://www.keycloak.org/docs/latest/server_admin/#user-navigation).

## Verification Steps
1. Check the 'Sessions' overview of the entire realm from the sidebar.
2. Check the 'Sessions' tab under a specific client.
3. Check the 'Sessions' tab under a specific user.

## Additional Notes
Tests will be implemented separately.

### DIFF
--- a/public/resources/en/sessions.json
+++ b/public/resources/en/sessions.json
@@ -2,10 +2,10 @@
   "title": "Sessions",
   "sessionExplain": "Sessions are sessions of users in this realm and the clients that they access within the session.",
   "searchForSession": "Search session",
-  "subject": "Subject",
-  "lastAccess": "Last access",
-  "startDate": "Start date",
-  "accessedClients": "Accessed clients",
+  "user": "User",
+  "lastAccess": "Last Access",
+  "started": "Started",
+  "clients": "Clients",
   "sessionsType": {
     "allSessions": "All session types",
     "regularSSO": "Regular SSO",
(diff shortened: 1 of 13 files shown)


jonkoops:
@xianli123 I am adding you as a reviewer as this has no reference design, so I'd like your stamp of approval before we merge this.

xianli123 (approved):
It looks good. Thanks!

public/resources/en/sessions.json
```diff
@@ -2,10 +2,10 @@
   "title": "Sessions",
   "sessionExplain": "Sessions are sessions of users in this realm and the clients that they access within the session.",
   "searchForSession": "Search session",
-  "subject": "Subject",
-  "lastAccess": "Last access",
-  "startDate": "Start date",
-  "accessedClients": "Accessed clients",
+  "user": "User",
+  "lastAccess": "Last Access",
```
edewit, line 6: everywhere else we use lower case so this would need to be:
```suggestion
  "lastAccess": "Last access",
```

src/sessions/SessionsSection.tsx
```diff
@@ -51,7 +45,7 @@ export default function SessionsSection() {
 
     setNoSessions(clientSessions.length === 0);
 
-    const allClients = await adminClient.clients.find();
+    const allClients = await adminClient.clients.find({ realm });
```
edewit, line 48: Realm here is not needed as the adminClient is already setup with the active realm
```suggestion
    const allClients = await adminClient.clients.find();
```

edewit (approved):

PR #2672, opened 2022-05-20
Found some small issues while creating the documentation

edewit:
Fixed:
- fixed namespace
- fallback to JS when type is unknown

## Checklist:

### DIFF
--- a/src/clients/authorization/policy/PolicyDetails.tsx
+++ b/src/clients/authorization/policy/PolicyDetails.tsx
@@ -57,7 +57,8 @@ const COMPONENTS: {
   js: JavaScript,
 } as const;
 
-const isValidComponentType = (value: string): boolean => value in COMPONENTS;
+export const isValidComponentType = (value: string): boolean =>
+  value in COMPONENTS;
 
 export default function PolicyDetails() {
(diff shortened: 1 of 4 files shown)


src/clients/authorization/policy/PolicyDetails.tsx
```diff
@@ -57,7 +57,8 @@ const COMPONENTS: {
   js: JavaScript,
 } as const;
 
-const isValidComponentType = (value: string): boolean => value in COMPONENTS;
+export const isValidComponentType = (value: string): boolean =>
```
jonkoops, line 60: I believe the return type here can be inferred.
```suggestion
export const isValidComponentType = (value: string) =>
```

jonkoops, line 60:
```suggestion
export const isValidComponentType = (value: string): boolean =>I believe the return type here can be inferred.
```suggestion
export const isValidComponentType = (value: string) =>
```
```

jonkoops (changes requested):
Looks like there are some compilation issues.

jonkoops (approved):

PR #2706, opened 2022-05-27
Use correct field for authentication of e-mail server

### LINKED ISSUE #2701: SMTP configuration adds authentication instead of auth
### Describe the bug

Please see https://github.com/keycloak/keycloak/issues/12175 for more details. But the summary is that the email tab is using `authentication` to enable username/password login for sending emails. The real configuration parameter should be `auth` (note that you are doing something special to do the test, it was a mistake, it should be `auth`). I make it work doing this little change in `EmailTab.tsx`:

```diff
diff --git a/src/realm-settings/EmailTab.tsx b/src/realm-settings/EmailTab.tsx
index 9f0f855d..7e8fe9e5 100644
--- a/src/realm-settings/EmailTab.tsx
+++ b/src/realm-settings/EmailTab.tsx
@@ -56,7 +56,7 @@ export const RealmSettingsEmailTab = ({
 
   const authenticationEnabled = useWatch({
     control,
-    name: "smtpServer.authentication",
+    name: "smtpServer.auth",
     defaultValue: "",
   });
 
@@ -93,10 +93,8 @@ export const RealmSettingsEmailTab = ({
       serverSettings.starttls = serverSettings.starttls === true.toString();
     }
 
-    // For some reason the API wants a duplicate field for the authentication status.
-    // Somebody thought this was a good idea, so here we are.
-    if (serverSettings.authentication === true.toString()) {
-      serverSettings.auth = true;
+    if (typeof serverSettings.auth === "string") {
+      serverSettings.auth = serverSettings.auth === true.toString();
     }
 
     try {
@@ -321,7 +319,7 @@ export const RealmSettingsEmailTab = ({
               fieldId="kc-authentication"
             >
               <Controller
-                name="smtpServer.authentication"
+                name="smtpServer.auth"
                 control={control}
                 defaultValue=""
                 render={({ onChange, value }) => (
```

If you want I can send a PR or you can pick the little diff and integrate/inprove it.

Thanks!

### Version

18.0.0

### Expected behavior

Sending mails should work in normal processing out of the test.

### Actual behavior

Sending emails doesn't work as the server internally checks for [auth](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/email/DefaultEmailSenderProvider.java#L78).

### How to Reproduce?

See https://github.com/keycloak/keycloak/issues/12175.

### Anything else?

_No response_

jonkoops:
Closes #2701

### DIFF
--- a/src/realm-settings/EmailTab.tsx
+++ b/src/realm-settings/EmailTab.tsx
@@ -78,23 +78,19 @@ export const RealmSettingsEmailTab = ({
   };
 
   const testConnection = async () => {
+    const toNumber = (value: string) => Number(value);
+    const toBoolean = (value: string) => value === true.toString();
+    const valueMapper = new Map<string, (value: string) => unknown>([
+      ["port", toNumber],
+      ["ssl", toBoolean],
+      ["starttls", toBoolean],
+      ["auth", toBoolean],
+    ]);
+
     const serverSettings = { ...getValues()["smtpServer"] };
 
-    // Code below uses defensive coding as the server configuration uses an ambiguous record type.
-    if (typeof serverSettings.port === "string") {
-      serverSettings.port = Number(serverSettings.port);
-    }
-
-    if (typeof serverSettings.ssl === "string") {
-      serverSettings.ssl = serverSettings.ssl === true.toString();
-    }
-
-    if (typeof serverSettings.starttls === "string") {
-      serverSettings.starttls = serverSettings.starttls === true.toString();
-    }
-
-    if (serverSettings.auth === "string") {
-      serverSettings.auth = serverSettings.auth === true.toString();
+    for (const [key, mapperFn] of valueMapper.entries()) {
+      serverSettings[key] = mapperFn(serverSettings[key]);
     }
 
     try {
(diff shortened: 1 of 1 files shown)


src/realm-settings/EmailTab.tsx
```diff
@@ -93,10 +93,8 @@ export const RealmSettingsEmailTab = ({
       serverSettings.starttls = serverSettings.starttls === true.toString();
     }
 
-    // For some reason the API wants a duplicate field for the authentication status.
-    // Somebody thought this was a good idea, so here we are.
-    if (serverSettings.authentication === true.toString()) {
-      serverSettings.auth = true;
+    if (serverSettings.auth === "string") {
+      serverSettings.auth = serverSettings.auth === true.toString();
```
edewit, line 97: This can be removed as well right? It's always a string but it can also be "false" so don't think we want to then put to change it to "true"

edewit (changes requested):

src/realm-settings/EmailTab.tsx
jonkoops: This is to work around a quirk in the Admin API. Basically when testing the connection the API expects these fields to be a `boolean`, but when saving the resource it expects that it's a `string` since the e-mail connection settings are saved as a `Map<string, string>`.

It's the way the Admin API works, unfortunately.

edewit (approved):

PR #2780, opened 2022-06-09
added groups to dropdown menu

edewit:
fixes: #2747

## Checklist:

### DIFF
--- a/src/realm-settings/NewAttributeSettings.tsx
+++ b/src/realm-settings/NewAttributeSettings.tsx
@@ -114,7 +114,7 @@ export default function NewAttributeSettings() {
     }));
 
   useFetch(
-    () => adminClient.users.getProfile({ realm }),
+    adminClient.users.getProfile,
     (config) => {
       setConfig(config);
       const {
@@ -174,7 +174,8 @@ export default function NewAttributeSettings() {
           },
           profileConfig.isRequired
             ? { required: profileConfig.required }
-            : undefined
+            : undefined,
+          profileConfig.group ? { group: profileConfig.group } : undefined
         );
       });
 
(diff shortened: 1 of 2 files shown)


src/realm-settings/NewAttributeSettings.tsx
```diff
@@ -114,7 +114,7 @@ export default function NewAttributeSettings() {
     }));
 
   useFetch(
-    () => adminClient.users.getProfile({ realm }),
+    adminClient.users.getProfile,
```
jonkoops, line 117: Not a fan of this as it makes the contract between `useFetch` and it's callback implicit. This means that if the callback is changed to accept some parameters then they will automatically be passed, leading to possible undefined behavior.

```suggestion
    () => adminClient.users.getProfile(),
```

src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
```diff
@@ -60,13 +62,8 @@ export const AttributeGeneralSettings = () => {
...
 
-  useFetch(
-    () => adminClient.clientScopes.find(),
-    (clientScopes) => {
-      setClientScopes(clientScopes);
-    },
-    []
-  );
+  useFetch(adminClient.clientScopes.find, setClientScopes, []);
+  useFetch(adminClient.users.getProfile, setConfig, []);
```
jonkoops, line 66: Again, not a fan of making this implicit.
```suggestion
  useFetch(() => adminClient.clientScopes.find(), setClientScopes, []);
  useFetch(() => adminClient.users.getProfile(), setConfig, []);
```

jonkoops (changes requested):

jonkoops (approved):

src/realm-settings/user-profile/attribute/AttributeGeneralSettings.tsx
edewit: we do this everywhere though, if you feel strongly about we should agree on this and make a code style rule for it

PR #2784, opened 2022-06-09
fixed required for + localization

edewit:
fixes: #2746

## Checklist:


src/realm-settings/NewAttributeSettings.tsx
```diff
@@ -162,6 +162,7 @@ export default function NewAttributeSettings() {
           return attribute;
         }
 
+        delete attribute.required;
```
jonkoops, line 165: This mutates the original reference, are you sure you don't want to do this on the resulting object instead?

jonkoops (approved):

Issue #2777, opened 2022-06-08
Navigation links cannot be opened in new tab


danielFesenmeyer:
### Describe the bug

The links in the left-side navigation of the Admin UI are not correct. E.g. when trying to open "Clients" in a new tab, I get an error "404 Not Found".

### Version

18.0.0

### Expected behavior

When opening the "Clients" link in a new tab, the clients of the currently selected realm should be displayed in this tab.

### Actual behavior

When opening the "Clients" link in a new tab, I get an error "404 Not Found".

### How to Reproduce?

1. Start Keycloak.X with the following command:

```
./kc.sh start-dev --features=preview
```

2. Open the Admin UI, for example by opening the URL http://localhost:8080/admin/master/console/#/

3. Open the "Clients" link (from the left-side navigation) in a new tab.
4. The new tab shows an error "404 Not Found".

### Anything else?

This issue seems to affect all links of the left-side navigation, not only the "Clients" link.

Screenshot showing the incorrect "Clients" link (see "localhost:8080/master/clients" in the left bottom corner):
![adminUIV2NavigationLinks](https://user-images.githubusercontent.com/11198630/172647996-8da1156d-87e9-483c-9077-cc58ade7d1c0.PNG)

ssilvert:
Thanks for reporting this @danielFesenmeyer.

PR #2788, opened 2022-06-10
Changed to use NavLink to fix urls based on router

edewit:
fixes: #2777

## Checklist:

### DIFF
--- a/src/PageNav.tsx
+++ b/src/PageNav.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { useHistory, useRouteMatch } from "react-router-dom";
+import { NavLink, useHistory, useRouteMatch } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import {
   Nav,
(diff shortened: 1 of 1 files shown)


src/PageNav.tsx
```diff
@@ -52,17 +52,19 @@ export const PageNav: React.FunctionComponent = () => {
     }
 
     //remove "/realm-name" from the start of the path
-    const activeItem = history.location.pathname.substr(realm.length + 1);
+    const activeItem = history.location.pathname.substring(realm.length + 1);
     return (
-      <NavItem
+      <NavLink
```
jonkoops, line 57: You might be able to pass this as the `component` prop to `NavItem` instead.

jonkoops (approved):

src/PageNav.tsx
edewit: That didn't work as they both have an `isActive` attribute that has a different signature.

PR #2808, opened 2022-06-14
Replace Snowpack with Vite

jonkoops:
I set up an experiment to see how hard it would be to replace Snowpack with Vite and this PR is the result of that. Let's see if we can get this PR in a mergable state over time. The following tasks will have to be completed:

### Checklist
- [x] Some images such as the logo in the nav are not loaded properly.
- [ ] Cypress tests are no longer starting because of changes in the TypeScript configuration.
- [x] Certain assets from Patternfly such as fonts are not loaded properly as [they were in Snowpack](https://github.com/keycloak/keycloak-admin-ui/blob/f775fd63296c54556fcb6d5f1de78fb9fca2a340/snowpack.config.mjs#L9-L25).
- [ ] Jest tests are failing due to TypeScript configuration.
- [x] TypeScript is not checked during development (solution is [`vite-plugin-checker`](https://www.npmjs.com/package/vite-plugin-checker))
- [x] Imports that use the React namespace (`import React from 'react'`) will have to be removed as the [React plugin](https://www.npmjs.com/package/@vitejs/plugin-react) uses the [automatic JSX runtime](https://github.com/alloc/vite-react-jsx#faq).
- [ ] Live reloading code changes throws errors


cypress.config.ts
```diff
@@ -1,8 +1,5 @@
-import webpackPreprocessor from "@cypress/webpack-batteries-included-preprocessor";
 import { defineConfig } from "cypress";
 import del from "del";
-import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
```
jonkoops, line 4: I don't think we should remove this long term as we also want to type-check our Cypress tests. Perhaps there is a better alternative than this though.

edewit: Don't we check already all the files why do we need to run this only when cypress is building?

jonkoops: We don't check these files currently. Putting this in the Cypress build step ensures it is checked at that point during development and production builds. That said, I would love to somehow get rid of Webpack here and use Vite as a build system for the Cypress tests also.

edewit: wouldn't it be as simple as doing:
```javascript
{
    devServer: {
      framework: 'react',
      bundler: 'vite',
 //...
}
```

jonkoops: This has been resolved by moving the type checks into a seperate task.

Issue #2778, opened 2022-06-08
Error message in User Federation provider is not user friendly


Aboullos:
### Describe the bug

When trying to save an `User Federation provider` that is not '`Writable only'` and enabling the `Validate Password Policy` the error message is not user friendly `User federation provider could not be saved: ldapErrorValidatePasswordPolicyAvailableForWritableOnly`

### Version

18.0.0

### Expected behavior

The message should be more understandable for the user

### Actual behavior

The message is not easy to understand for the user

### How to Reproduce?

- Go to an `User Federation provider`
- Set the edit mode to one that is not `Writable only`
- Enable the `Validate Password Policy`
- Save

Aboullos:
@ssilvert Great!, yes, the test is already written in a draft PR, for this one and another message that is also failing

PR #2819, opened 2022-06-17
Add theme.properties to the build.

ssilvert:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/2778

## Brief Description
From my testing, this should fix the problem with User Federation error messages that are not localized.

The problem was that we have not been including a theme.properties.  So unless you run the old admin console first, the message bundles for localization of error messages will never get loaded.

## Verification Steps
See steps in https://github.com/keycloak/keycloak-admin-ui/issues/2778

But note that you wouldn't see the issue if you had already run the old console against the keycloak database.  So for a valid test you need to start with a clean database (or just delete it).

### DIFF
--- a/keycloak-theme/src/main/resources/theme/keycloak.v2/admin/theme.properties
+++ b/keycloak-theme/src/main/resources/theme/keycloak.v2/admin/theme.properties
@@ -1,0 +1,1 @@
+parent=base


ssilvert:
The six tests that failed in CI all passed locally.  Only one test failed locally, but that one passed in CI.

jonkoops (approved):

PR #2869, opened 2022-06-22
Always Display in Console missing from OIDC clients

ssilvert:
## Motivation
In Clients --> Settings, the "Always display in console" field was missing from OIDC clients.

Also, for SAML, Front Channel Logout was in General Settings instead of Logout settings.

## Verification Steps
Verify that "Always display in console" is always visible for both OIDC and SAML clients.
Verify "Front Channel Logout" is in Logout settings for both OIDC and SAML clients.

## Checklist:

- [X] Help has been implemented
- [N/A] axe report has been run and resulting a11y issues have been resolved
- [N/A] Unit tests have been created/updated

### DIFF
--- a/cypress/e2e/realm_settings_tabs_test.spec.ts
+++ b/cypress/e2e/realm_settings_tabs_test.spec.ts
@@ -39,39 +39,35 @@ describe("Realm settings tabs tests", () => {
     cy.findByTestId(realmSettingsPage.userProfileTab).should("exist");
   });
 
+  // Toggles will update the database immediately.
+  // So we need to force realm settings to reload - otherwise you aren't testing anything but patternfly
+  // This also keeps Cypress from getting hung.
+  function reloadRealm() {
+    sidebarPage.goToClientScopes();
+    sidebarPage.goToRealmSettings();
+    cy.findByTestId("rs-login-tab").click();
+  }
+
+  function testToggle(realmSwitch: string, expectedValue: string) {
+    realmSettingsPage.toggleSwitch(realmSwitch);
+    reloadRealm();
+    cy.findByTestId(realmSwitch).should("have.value", expectedValue);
+  }
+
   it("Go to login tab", () => {
     sidebarPage.goToRealmSettings();
     cy.findByTestId("rs-login-tab").click();
-    realmSettingsPage.toggleSwitch(realmSettingsPage.userRegSwitch);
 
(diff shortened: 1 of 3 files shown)


jonkoops (approved):

ssilvert:
@jonkoops I think I also fixed our flakiest test with this PR.  Please re-review.

src/clients/add/LogoutPanel.tsx
```diff
@@ -37,75 +37,71 @@ export const LogoutPanel = ({
...
+              label={t("common:on")}
+              labelOff={t("common:off")}
+              isChecked={value.toString() === "true"}
+              onChange={(value) => onChange(value.toString())}
+            />
+          )}
+        />
+      </FormGroup>
+      {protocol === "openid-connect" &&
+        frontchannelLogout?.toString() === "true" && (
```
edewit, line 67: Is the type in ClientRepresentation wrong or is our Switch above wrong?

edewit (approved):

cypress/e2e/realm_settings_tabs_test.spec.ts
```diff
@@ -39,40 +39,36 @@ describe("Realm settings tabs tests", () => {
     cy.findByTestId(realmSettingsPage.userProfileTab).should("exist");
   });
 
-  it("Go to login tab", () => {
+  // Toggles will update the database immediately.
+  // So we need to force realm settings to reload - otherwise you aren't testing anything but patternfly
+  // This also keeps Cypress from getting hung.
```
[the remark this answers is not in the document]
ssilvert: I was under the impression that the full realm was not reloaded in this case.  Looking at the network calls, I see I'm probably wrong.

The big issue here is that clicking on multiple toggles in succession causes quick re-renderings of the screen.  There is a noticeable flicker during the test. So sometimes, this would screw up the test and cause Cypress to hang.  Clicking to another section each time fixed the problem.

So it looks like just the comment is wrong.  I'll fix the comment.

src/clients/add/LogoutPanel.tsx
ssilvert: I don't know.  I left that conditional as-is.

jonkoops (approved):

jonkoops:
Looks good to me!

agagancarczyk (approved):

PR #2896, opened 2022-06-30
changed to new drag and drop

edewit:
## Checklist:


src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -26,6 +29,14 @@ type ManageOrderDialogProps = {
   onClose: () => void;
 };
 
+// Copied from patternfly this seems not exported
```
jonkoops, line 32: Can we log an issue for this to PatternFly and link it here with a TODO? That way we can keep track of the resolution.

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -34,28 +45,51 @@ export const ManageOrderDialog = ({
...
   };
 
-  const onDragCancel = () => {
-    setLiveText(t("common:onDragCancel"));
-  };
+  const onDrop = (
+    source: DraggableItemPosition,
+    dest?: DraggableItemPosition
+  ): boolean => {
+    if (dest) {
```
jonkoops, line 79: This could potentially benefit from an early return.

```ts
if (!dest) {
  setLiveText(t("common:onDragInvalid"));
  return false;
}

// The rest of the code.
```

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
-      </DataList>
-      <div className="pf-screen-reader" aria-live="assertive">
-        {liveText}
-      </div>
+            {order.map((alias) => (
+              <Draggable key={alias} hasNoWrapper>
+                <DataListItem
+                  aria-labelledby={alias}
+                  id={`${alias}-item`}
+                  ref={React.createRef()}
```
jonkoops, line 152: This ref is created every time this renders and seems to be referenced nowhere. Perhaps it can be removed?

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
+                >
+                  <DataListItemRow>
+                    <DataListControl>
+                      <DataListDragButton
+                        aria-label="Reorder"
```
jonkoops, line 157: This label will need to be translated.

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
+                        aria-labelledby={alias}
```
jonkoops, line 158: An element cannot have both `aria-label` and `aria-labelledby`, it can only have one.

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
+                        aria-describedby={t("manageOrderItemAria")}
```
jonkoops, line 159: `aria-describedby` expects the id of another element, not a label. See the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby).

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
+                        aria-pressed="false"
```
jonkoops, line 160: What is the purpose of passing this? It never seems to change.

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -101,48 +137,46 @@ export const ManageOrderDialog = ({
...
+                      />
+                    </DataListControl>
+                    <DataListItemCells
+                      dataListCells={[
+                        <DataListCell key={`${alias}-cell`} data-testid={alias}>
```
jonkoops, line 165: Since keys do not need to be globally unique I think you can get away with passing this value directly.
```suggestion
                        <DataListCell key={alias} data-testid={alias}>
```

src/identity-providers/ManageOrderDialog.tsx
```diff
@@ -34,28 +45,51 @@ export const ManageOrderDialog = ({
...
+      const result = [...order];
+      const [removed] = result.splice(source.index, 1);
+      result.splice(dest.index, 0, removed);
+      setOrder(result);
 
-  const onDragFinish = (providerOrder: string[]) => {
-    setLiveText(t("common:onDragFinish", { list: providerOrder }));
-    setOrder(providerOrder);
+      setLiveText(
+        t("common:onDragFinish", { list: providers.map((p) => p.alias) })
```
jonkoops, line 86: You could get rid of this one char variable with some destructuring.
```suggestion
        t("common:onDragFinish", { list: providers.map(({ alias }) => alias) })
```

jonkoops (changes requested):

src/identity-providers/ManageOrderDialog.tsx
edewit: even better created a PR for it: https://github.com/patternfly/patternfly-react/pull/7641

jonkoops (approved):

edewit:
tests don't work with new drag and drop

PR #2898, opened 2022-07-01
Serious accessibility fixes for Client Scopes page #2850

agagancarczyk:
## Motivation
Fixes for all serious ay11 issues. Closes https://github.com/keycloak/keycloak-admin-ui/issues/2850.

## Verification Steps
1. Run axe for `Add predefined mapper` and `Configure a new mapper`.

## Checklist:


src/client-scopes/add/MapperDialog.tsx
```diff
@@ -142,7 +149,7 @@ export const AddMapperDialog = (props: AddMapperDialogProps) => {
           aria-label={t("addPredefinedMappers")}
           isCompact
         >
-          <DataListItem aria-labelledby="headerName" id="header">
+          <DataListItem aria-labelledby={t("headerName")} id="header">
```
jonkoops, line 152: The `aria-labeledby` expects a reference to the `id` of another element, not a translated text. This either needs to point to the right element, or `aria-label` should be used.

jonkoops (changes requested):

src/client-scopes/add/MapperDialog.tsx
agagancarczyk: Thanks @jonkoops using `aria-label` instead sounds good to me. This is now fixed.

jonkoops (approved):

PR #2923, opened 2022-07-05
Add utilty function to check for defined values

jonkoops:
Adds an easy to use utility function to check for values that are defined (anything not `undefined` or `null`). This allows for filtering collections in a manner that will provide type safety guarantees.


agagancarczyk (approved):
Very nice util func.

jonkoops:
Merging this PR as I need this function in a couple of other PRs.

PR #2936, opened 2022-07-08
Replace Snowpack with Vite

jonkoops:
Replaces Snowpack with Vite as it is no longer [actively maintained](https://github.com/FredKSchott/snowpack/issues/3491).


edewit (approved):

edewit (commented):
No more rhsso theme?

edewit (approved):

jonkoops:
> No more rhsso theme?

Yeah, we never agreed on this with the productization team so we need to sit down with them and work this out.

agagancarczyk (approved):

agagancarczyk:
@jonkoops all looks good except for the issue with "the context hierarchy" when you make a change on some files i.e Dashboard.tsx, which we have already discussed. I can review this again.

jonkoops:
> all looks good except for the issue with "the context hierarchy" when you make a change on some files

Looking into that now, but should not be a blocker for merging this PR I think. Worst case scenario we have to do some manual refreshing during development every once and a while.

Issue #1673, opened 2021-12-07
Notification message is not disappearing


ikhomyn:
### Describe the bug

Changing _Assigned type_ of the item in Client Scopes is showing multiple notifications which some of those never disappears

### Version

Keycloak 15.0.2 (WildFly Core 15.0.1.Final) 

### Expected behavior

Notification should disappear after some short time. 1 Notification per action.

### Actual behavior

Notification never disappear. Multiple notifications are shown instead of single one per action.

### How to Reproduce?

1. Go to **Client scopes** 
2. Change _Assigned type_ to Default/Optional of the item from the item combobox in the list. One notification should appear.
3. Wait until notification disappears.
4. Again Change _Assigned type_ to Default/Optional of the item from the item combobox in the list.
5. Two notification appear instead of one. One will disappear and another remains.

### Anything else?

More times Assigned type is changed, more notification are shown on next time. Looks like some array need to be cleaned before showing notification after actions.

ikhomyn:
Issue is not longer reproducible

PR #2946, opened 2022-07-08
Always close notifications during testing

jonkoops:
Many test failures are caused by to many notifications on the screen, closing them every time they are checked reduce the chance they cause issues.

### DIFF
--- a/cypress/e2e/client_authorization_test.spec.ts
+++ b/cypress/e2e/client_authorization_test.spec.ts
@@ -47,7 +47,7 @@ describe("Client authentication subtab", () => {
 
   it("Should update the resource server settings", () => {
     policiesSubTab.setPolicy("DISABLED").formUtils().save();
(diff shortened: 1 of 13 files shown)


edewit (approved):

edewit:
Yes we could maybe do it like this, but the problem might be that closing the alert will close multiple and then it fails on closing the next

Aboullos:
I am not sure if this is the best way, but at least is not the best way now. It seemed that the notifications weren't closing because of a bug. If you check the failure video there are some notifications that are still showing from the beginning of the test.
Ivan was trying to reproduce it because he found this bug before, but last time couldn't reproduce it so he closed the issue.
If we close every notification we won't know if the problem keeps happening or not.

ikhomyn:
This was the issue https://github.com/keycloak/keycloak-admin-ui/issues/1673, but when I've tried it, couldn't reproduce it. At least on that time.

PR #2955, opened 2022-07-12
made automatic removal of alerts more robust

edewit:
## Checklist:

### DIFF
--- a/src/components/alert/AlertPanel.tsx
+++ b/src/components/alert/AlertPanel.tsx
@@ -5,14 +5,7 @@ import {
   AlertActionCloseButton,
   AlertVariant,
 } from "@patternfly/react-core";
-
-export type AlertType = {
-  id: number;
-  key: number;
-  message: string;
-  variant: AlertVariant;
-  description?: string;
-};
+import type { AlertType } from "./Alerts";
 
 type AlertPanelProps = {
   alerts: AlertType[];
--- a/src/components/alert/AlertPanel.tsx
+++ b/src/components/alert/AlertPanel.tsx
@@ -7,8 +7,6 @@ import {
 } from "@patternfly/react-core";
 import type { AlertType } from "./Alerts";
 
-const ALERT_TIMEOUT = 8000;
-
 type AlertPanelProps = {
   alerts: AlertType[];
   onCloseAlert: (id: number) => void;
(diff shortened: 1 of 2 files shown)


src/components/alert/AlertPanel.tsx
```diff
@@ -7,31 +7,32 @@ import {
 } from "@patternfly/react-core";
 
 export type AlertType = {
-  key: string;
+  id: number;
+  key: number;
```
jonkoops, line 11: It doesn't look like this prop is used anymore, perhaps it can be removed?
```suggestion
```

src/components/alert/Alerts.tsx
```diff
@@ -28,22 +32,37 @@ export const useAlerts = () => useRequiredContext(AlertContext);
...
     variant: AlertVariant = AlertVariant.success,
     description?: string
   ) => {
-    const key = createId();
-    setTimeout(() => hideAlert(key), 8000);
-    setAlerts([{ key, message, variant, description }, ...alerts]);
+    setAlerts([
+      {
+        id: Math.random(),
+        key: new Date().getTime(),
```
jonkoops, line 59: Can we call this something different than `key`? Since this is a timestamp of when the alert was created.

src/components/alert/Alerts.tsx
```diff
@@ -28,22 +32,37 @@ export const useAlerts = () => useRequiredContext(AlertContext);
...
+  };
 
-  const hideAlert = (key: string) => {
-    setAlerts((alerts) => [...alerts.filter((el) => el.key !== key)]);
+  const hideAlert = (id: number) => {
+    setAlerts((alerts) => [...alerts.filter((el) => el.id !== id)]);
   };
 
+  useEffect(() => {
+    const interval = setInterval(() => clearAlerts(), 3000);
```
jonkoops, line 47: Why run a constant timer to check this if we can set a timeout in the `addAlert()` function? If you are looking to tie this into the component lifecycle we also have a [`useSetTimout()`](https://github.com/keycloak/keycloak-admin-ui/blob/main/src/utils/useSetTimeout.ts) utility.

jonkoops (changes requested):

Aboullos (approved):
Great solution to the notification problem!

src/components/alert/Alerts.tsx
edewit: I would if it would work, but it sometimes doesn't fire this way even if the trigger was missed it will remove the alert

src/components/alert/Alerts.tsx
edewit: renamed it to `created`

jonkoops (approved):

PR #2963, opened 2022-07-13
fine-tuning of layout in Authorization Evaluate page (Issue 2159)

### LINKED ISSUE #2159: [UX Review] Some fine-tuning of the layout in the authorization Evaluate page
### Describe the bug

There is some fine-tuning of the Evaluate page as shown below.

1. The border can be removed. We don't need to use the card on this page.
Screen Shot 2022-02-28 at 3 08 05 PM

2. Every two buttons near each other should be 16px apart.
Screen Shot 2022-02-28 at 3 09 15 PM

3. The label is a bit confusing. So The label named "Key" and "Value" can be removed for the "Authorization resources and scopes".
Screen Shot 2022-02-28 at 3 25 32 PM

### Version

New admin UI

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

dominikkawka:
## Motivation
Close #2159

image

## Brief Description
See #2159

## Verification Steps
Clients -> Client details -> Permissions -> Enable permissions -> Scope Name -> Cancel button -> Authorization -> Evaluate

## Checklist:

## Additional Notes
The "Resources and Authentication Scopes" and "Contextual Attributes" stick out on the right side of the page. This was already present. I've changed the width of the input/selectable fields from 500px to 400px to make it less noticeable.

### DIFF
--- a/src/clients/authorization/auth-evaluate.css
+++ b/src/clients/authorization/auth-evaluate.css
@@ -20,16 +20,12 @@ button#reevaluate-btn {
   margin-right: var(--pf-global--spacer--md);
 }
 
-button#authorization-eval, button#authorization-revert {
-  margin-right: 16px;
-}
-
 .kc-identity-information, .kc-permissions {
   border: none !important;
 }
 
 .kc-identity-information > div.pf-c-card__header.kc-form-panel__header {
-  padding-top: 0px;
+  padding-top: 0;
 }
 
 #resourcesAndAuthScopes > .pf-c-form__group-control > .kc-attributes__table > thead #key, 
--- a/src/clients/authorization/auth-evaluate.css
+++ b/src/clients/authorization/auth-evaluate.css
@@ -20,4 +20,19 @@ button#reevaluate-btn {
   margin-right: var(--pf-global--spacer--md);
 }
 
+button#authorization-eval, button#authorization-revert {
+  margin-right: 16px;
+}
 
+.kc-identity-information, .kc-permissions {
+  border: none !important;
+}
+
+.kc-identity-information > div.pf-c-card__header.kc-form-panel__header {
+  padding-top: 0px;
+}
+
+#resourcesAndAuthScopes > .pf-c-form__group-control > .kc-attributes__table > thead #key, 
+#resourcesAndAuthScopes > .pf-c-form__group-control > .kc-attributes__table > thead #value {
+  display:none
+}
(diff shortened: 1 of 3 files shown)


src/clients/authorization/auth-evaluate.css
```diff
@@ -20,4 +20,19 @@ button#reevaluate-btn {
   margin-right: var(--pf-global--spacer--md);
 }
 
+button#authorization-eval, button#authorization-revert {
+  margin-right: 16px;
```
jonkoops, line 24: Instead of hard coding this value it might be better to use a variable from PatternFly or apply spacing using one of the [utility classes](https://www.patternfly.org/v4/utilities/spacing/) from PatternFly.

Also an ID should be unique enough, so there is no need to make this selector more specific.

src/clients/authorization/auth-evaluate.css
```diff
@@ -20,4 +20,19 @@ button#reevaluate-btn {
...
+}
+
+.kc-identity-information, .kc-permissions {
+  border: none !important;
+}
 
+.kc-identity-information > div.pf-c-card__header.kc-form-panel__header {
+  padding-top: 0px;
```
jonkoops, line 32: You can omit the unit here. Perhaps it would be better to use a utility class though.

```suggestion
  padding-top: 0;
```

jonkoops (approved):

src/clients/authorization/auth-evaluate.css
edewit: don't use ids for styling see our css guide https://github.com/keycloak/keycloak-admin-ui/blob/main/CODING_GUIDELINES.md#css

src/clients/authorization/auth-evaluate.css
```diff
@@ -20,4 +20,19 @@ button#reevaluate-btn {
...
+}
+
+#resourcesAndAuthScopes > .pf-c-form__group-control > .kc-attributes__table > thead #key, 
+#resourcesAndAuthScopes > .pf-c-form__group-control > .kc-attributes__table > thead #value {
+  display:none
```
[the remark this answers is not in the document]
dominikkawka: For now I'll commit the changes regarding the first two recommendations, lmk if I should change this/keep it as it is/approach it differently.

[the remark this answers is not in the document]
jonkoops: Let's keep it like this then, just not a fan of such complex selectors as they are prone to break when things change in PatternFly.

jonkoops (approved):

PR #2969, opened 2022-07-14
fine-tuning in policy details page (issue 2114)

dominikkawka:
## Motivation
Close #2114
## Brief Description
See #2114

image

## Verification Steps
Clients -> Choose Client -> Enable Permissions -> Choose scope-name -> Cancel -> Authorization Policies -> Create Policy -> Choose Client Scope/ Role/ Group -> Add Client Scope/ Role/ Group 

## Checklist:

- [x] Code has been tested locally by PR requester
- [ ] User-visible strings are using the react-i18next framework (useTranslation)
- [ ] Help has been implemented
- [ ] axe report has been run and resulting a11y issues have been resolved
- [ ] Unit tests have been created/updated

## Additional Notes

PR #2971, opened 2022-07-15
Replace CI trigger for Cypress

jonkoops:
This will allow us to run the Cypress tests on PRs from 'external' contributors such as @dominikkawka. See #2969 for an example of a PR where these tests are not running as expected.

Also includes a change to replace the hard-coded record key for the Cypress Dashboard with a secret so that it cannot be abused by 3rd parties. A new key was generated for this and the old key should be revoked as soon as this PR has been merged.

Please hold on merging this PR until @ssilvert has added this new key to the secrets to avoid potential isssues running the tests.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,5 +1,9 @@
 name: Cypress
-on: push
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
 jobs:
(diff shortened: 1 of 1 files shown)


ssilvert:
I have done the update to the Cypress Dashboard secret.

ssilvert (approved):

PR #2982, opened 2022-07-18
Changed button text from Delete to Revoke (Issue 1703)

### LINKED ISSUE #1703: [UX Review] Optimize the button wording of the revoke modal in User consents
### Describe the bug

In the User-Consents page, there will be a modal popup after clicking the **Revoke** option inside the kebab menu. In this modal, the primary button is "Delete". We can change the button from "Delete" to "Revoke" as shown in the screenshot below. Thanks.
Screen Shot 2021-12-14 at 2 01 11 PM

### Version

New admin UI

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

dominikkawka:
## Motivation
Close #1703

## Brief Description
See #1703

## Verification Steps
1. You can enable the "Consent required" and "Display client on screen" in the settings page of the "security-admin-console-v2" client.
2. Then you can logout and login again.
3. You will find some consents in the User>Consent page.
4. After clicking the kebab menu, there will be an option list that includes "Revoke" option.

image

## Checklist:

### DIFF
--- a/src/user/UserConsents.tsx
+++ b/src/user/UserConsents.tsx
@@ -65,7 +65,7 @@ export const UserConsents = () => {
     messageKey: t("users:revokeClientScopes", {
       clientId: selectedClient?.clientId,
     }),
-    continueButtonLabel: "common:delete",
+    continueButtonLabel: "common:revoke",
     continueButtonVariant: ButtonVariant.danger,
     onConfirm: async () => {
       try {
--- a/public/resources/en/common.json
+++ b/public/resources/en/common.json
@@ -13,7 +13,7 @@
   "close": "Close",
   "delete": "Delete",
   "remove": "Remove",
-  "revoke": "Revoke",
+  "revoke": "Revoke", 
   "search": "Search",
   "key": "Key",
   "value": "Value",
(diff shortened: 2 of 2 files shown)


src/user/UserConsents.tsx
```diff
@@ -65,7 +65,7 @@ export const UserConsents = () => {
     messageKey: t("users:revokeClientScopes", {
       clientId: selectedClient?.clientId,
     }),
-    continueButtonLabel: "common:delete",
+    continueButtonLabel: "common:Revoke",
```
jonkoops, line 68: I believe this should start with a lower case.
```suggestion
    continueButtonLabel: "common:revoke",
```

jonkoops (changes requested):

src/user/UserConsents.tsx
dominikkawka: When I change it to lower case it appears as lower case in the button. revoke doesn't seem to be effected by "common"
I looked around for a few min to see if I could but wasn't able to do so. 

image

jonkoops: I think this is because this key does not exist in any of the language bundles, so instead it defaults to the name of the key. You'll have to add a `revoke` entry to [`public/resources/en/common.json`](https://github.com/keycloak/keycloak-admin-ui/blob/main/public/resources/en/common.json).

jonkoops (approved):

PR #2997, opened 2022-07-23
Build Cypress tests with WebPack

jonkoops:
Uses the [`@cypress/webpack-batteries-included-preprocessor`](https://www.npmjs.com/package/@cypress/webpack-batteries-included-preprocessor) to pre-proccess the Cypress tests. This is required to upgrade to Vite 3, which does not play nice with Cypress (see #2991).

Sometime in the future we should switch test compilation back to Vite, but for now this will work and unblock the work to switch to Vite entirely.


jonkoops:
This PR should be merged as soon as possible to fix the broken tests in `main`.

edewit (approved):

agagancarczyk (commented):
Looks good.

PR #3033, opened 2022-08-02
Use new JSX transformation for React

jonkoops:
Introduces the [new JSX transformation](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) which allows us to get rid of the default export of React. It introduces a couple of other benefits:

- Better performance now and in the future. The [Motivations](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#motivation) section in the RFC explains the specifics of performance issues with `React.createElement`.
- In the future, we won't need `React.forwardRef` anymore.
- Faster parsing of JavaScript by web browser (`React.createElement` cannot be minified).
- Aligning with the defaults of Vite, removing complexity from the configuration.

This also introduces an ESLint rule that will prevent the default import of React from being used, instead named exports should be used.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:react/recommended",
+    "plugin:react/jsx-runtime",
     "plugin:@typescript-eslint/base",
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:prettier/recommended",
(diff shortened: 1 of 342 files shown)


ssilvert (approved):
LGTM

agagancarczyk (approved):

edewit (approved):

Issue #3022, opened 2022-07-29
Identity Provider Mapper can not save name of Hardcoded Attribute


lexcao:
### Describe the bug

As title descried and created from https://github.com/keycloak/keycloak/issues/13374

### Version

19.0.1

### Expected behavior

The name of the Hardcoded Attribute should be saved in Identity Provider Mapper.

### Actual behavior

Not saved.

### How to Reproduce?

1. Use the 19.0.0 Keycloak or Keycloak admin UI
2. Go to `Identity Providers`
3. Create a demo provider
4. Go to `Mappers` tab
5. Create a mapper
6. Choose `Hardcoded Attribute`
7. Input some test data
8. Click `save`

### Anything else?

I have some Investigations on my local keycloak.
It may be an expected behavior with `react-hook-form`, we need to adapt for new admin UI.

First, the hardcoded attribute mapper has the below properties
```
config.attribute  // for the name of Hardcoded User Attribute
config.attribute.value // for the value of Hardcoded User Attribute
```

Then, when passing these fields to `react-hook-form`, like this
```
register('attribute');
register('attribute.value');
```

While submitting, it builds a form object for the event, like this
```
{
 "attribute": { "value": "some-value" }
}
```
The `config.attribute` is removed from the form request, so the data is not saved correctly.

lexcao:
I found this issue also impacts other JSON config post requests. 

https://github.com/keycloak/keycloak/issues/13416

Which registers the `react-hook-form` field of this
```
register('x509-cert-auth.mapping-source-selection');
register('x509-cert-auth.xxxx');
```
Then builds the form object like this
```
"config": {
  "x509-cert-auth": {
    "mapping-source-selection": "xx",
    ...
}
```

lexcao:
I have found these 4 util functions, which can help fix the issue.
(There is a draft PR I am going to demo the fixing, #3023)

for react-hook-form
- `convertFormValuesToObject`
- `convertToFormValues`

for MultivaluedString
- `stringToMultiline`
- `toStringValue`

I am willing to fix the issue, but I need come confirmation about the way to fix it.

---
Firstly, for react-hook-form, I have googled the method about converting key of form object but found nothing.
There might be required to call the two util functions in somewhere of `save` and `setupForm` methods.
If so, there might be missing handling of the config fields like this issue in the future, then we need to add the conversion manually.

[...]

PR #3035, opened 2022-08-03
back to changing dots instead of unflatten

edewit:
fixes: #3022
and also fixes: https://github.com/keycloak/keycloak/issues/13416

### DIFF
--- a/src/authentication/components/ExecutionConfigModal.tsx
+++ b/src/authentication/components/ExecutionConfigModal.tsx
@@ -17,13 +17,13 @@ import { CogIcon, TrashIcon } from "@patternfly/react-icons";
 
 import type AuthenticatorConfigRepresentation from "@keycloak/keycloak-admin-client/lib/defs/authenticatorConfigRepresentation";
 import type AuthenticatorConfigInfoRepresentation from "@keycloak/keycloak-admin-client/lib/defs/authenticatorConfigInfoRepresentation";
-import type { ConfigPropertyRepresentation } from "@keycloak/keycloak-admin-client/lib/defs/authenticatorConfigInfoRepresentation";
 import type { ExpandableExecution } from "../execution-model";
 import { useAdminClient, useFetch } from "../../context/auth/AdminClient";
 import { useAlerts } from "../../components/alert/Alerts";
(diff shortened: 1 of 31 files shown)


edewit:
This fix might be better then https://github.com/keycloak/keycloak-admin-ui/pull/3023 @lexcao want to give it a try?

src/util.test.ts
```diff
@@ -3,6 +3,8 @@ import { convertFormValuesToObject, convertToFormValues } from "./util";
 
 vi.mock("react");
 
+const TOKEN = "🍺";
```
jonkoops, line 6: Proost! :beers:

lexcao:
Hi, @edewit 
This fix makes sense to me.
It replaces all fields of the form.
Thank you!

src/identity-providers/add/AddMapper.tsx
```diff
@@ -166,8 +166,6 @@ export default function AddMapper() {
 
   const setupForm = (mapper: IdentityProviderMapperRepresentation) => {
     convertToFormValues(mapper, form.setValue);
-    form.setValue("config.attributes", JSON.parse(mapper.config.attributes));
```
jonkoops, line 169: This should still be parsed, no?

jonkoops (changes requested):

src/authentication/components/ExecutionConfigModal.tsx
```diff
@@ -94,7 +80,7 @@ export const ExecutionConfigModal = ({
   );
 
   useEffect(() => {
-    if (configDescription) setupForm(configDescription, config);
+    if (configDescription) setupForm(config);
```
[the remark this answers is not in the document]
edewit: I think this was a left over of a old implementation now the `configDescription` is used to in conjunction with the dynamic fields

jonkoops (approved):

mortenjoenby:
Hi.
This is a critical issue to us as we are supporting several hundred realms across several Keycloak cluster, which we have just upgraded to 19.0.1 ...!
When can we expect 19.0.2 to be released - assuming this will contain the fix?
Is it an option to run with the old Admin Console until then?

Thanks,
Morten

PR #3054, opened 2022-08-04
Enabled and Disable features incorrect

### LINKED ISSUE #1718: Enabled and Disable features incorrect

mabartos:
Fixes #1718

## Checklist:

## Additional Notes
Before (without the red lines and some minor things):
image

After:
Screenshot from 2022-08-04 12-00-12

-------------------
@edewit @jonkoops Could you please take a look at it? It's only a template with the working solution for the bug, so feel free to add some commit with a cleaner approach or use some parts of this PR. Perhaps, it'd be good to create some separate utility for the feature states. WDYT? 

AFAIK, there's no feature, which will be either preview or experimental and enabled by default ATM. I don't think there's some case when we're enabling a feature, which is not fully supported. 

Thanks!


mabartos:
I've just noticed, that the collection of enabled features is already present in the `useIsFeatureEnabled` utility. It'd be probably suitable for this issue.

edewit (approved):

src/dashboard/Dashboard.tsx
```diff
@@ -212,7 +225,25 @@ const Dashboard = () => {
                           <DescriptionListDescription>
                             <List variant={ListVariant.inline}>
                               {disabledFeatures.map((feature) => (
-                                <ListItem key={feature}>{feature}</ListItem>
+                                <ListItem
+                                  key={feature}
+                                  className="keycloak__features_list_item"
```
jonkoops, line 230: Instead of using custom styling to set the margin you can use the [PatternFly utility classes](https://www.patternfly.org/v4/utilities/spacing). These use a standardized set of spacing sizes.

jonkoops (changes requested):

src/dashboard/Dashboard.tsx
mabartos: Fixed

mabartos:
@jonkoops I've fixed the CSS. Could you please review it again, please? Thank you :)

jonkoops (approved):

jonkoops:
Thanks @mabartos!

mabartos:
@edewit @jonkoops Thanks!

PR #3058, opened 2022-08-04
Fix homepage URL rendering in client list

### LINKED ISSUE #3057: Homepage URL is wrong in Client list in ClientSection
### Describe the bug

If a client configures a `rootUrl` and relative `baseUrl`, then the homepage URL in the client list is wrong.

With Keycloak running on `http://localhost:8080` and the following client configuration:
```
clientId: test-client
rootUrl: http://localhost:4200
baseUrl: /home
```

image

Yields the following url:

image

-> `/home` resolves to `http://localhost:8080/home` instead of `http://localhost:4200/home`.

### Version

19.0.1

### Expected behavior

Client list should render the correct URL

### Actual behavior

The homepage link should consider the rootUrl and baseUrl correctly.

### How to Reproduce?

1) Create a client with a rootUrl and relative baseUrl.
2) Check the homepage link on the client list.

### Anything else?

_No response_

thomasdarimont:
Closes #3057

### DIFF
--- a/src/clients/ClientsSection.tsx
+++ b/src/clients/ClientsSection.tsx
@@ -103,6 +103,40 @@ export default function ClientsSection() {
     </TableText>
   );
 
+  const ClientHomePage = (client: ClientRepresentation) => {
+    const rootUrl = client.rootUrl;
+    const baseUrl = client.baseUrl;
+
+    // absolute base url configured, use base url is
+    if (baseUrl?.startsWith("http")) {
+      return baseUrl;
+    }
+
+    if (rootUrl === "${authBaseUrl}" && baseUrl) {
+      // eslint-disable-next-line prettier/prettier
+      return rootUrl.replace("${authBaseUrl}", addTrailingSlash(adminClient.baseUrl)) + baseUrl.substring(1);
+    }
+
+    if (rootUrl === "${authAdminUrl}" && baseUrl) {
+      // eslint-disable-next-line prettier/prettier
+      return rootUrl.replace("${authAdminUrl}", addTrailingSlash(adminClient.baseUrl)) + baseUrl.substring(1);
+    }
+
+    if (!rootUrl) {
+      return baseUrl;
+    }
+
+    if (rootUrl.startsWith("http")) {
+      let targetUrl = rootUrl;
+      if (baseUrl) {
+        targetUrl = addTrailingSlash(targetUrl) + baseUrl.substring(1);
+      }
+      return targetUrl;
+    }
+
+    return baseUrl;
+  };
+
   const ToolbarItems = () => {
     if (!isManager) return <span />;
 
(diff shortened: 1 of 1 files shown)


jonkoops:
@thomasdarimont could you point towards the issue this PR resolves or add a description to the PR?

src/clients/ClientsSection.tsx
```diff
@@ -103,6 +103,40 @@ export default function ClientsSection() {
...
+      return baseUrl;
+    }
+
+    if (rootUrl.startsWith("http")) {
+      let targetUrl = rootUrl;
+      if (baseUrl) {
+        targetUrl = addTrailingSlash(targetUrl) + baseUrl.substring(1);
+      }
+      return targetUrl;
+    }
```
edewit, line 135:
```suggestion
    if (rootUrl.startsWith("http") && baseUrl) {
        return addTrailingSlash(rootUrl) + baseUrl.substring(1);
   }
   
   return rootUrl;
```

edewit: but shouldn't we test that baseUrl starts with a slash instead of always doing a `baseUrl.substring(1)`?

jonkoops: That's why I created the new [`joinPath`](https://github.com/keycloak/keycloak-admin-ui/blob/main/src/utils/joinPath.ts) utility :wink:

src/clients/ClientsSection.tsx
```diff
@@ -103,6 +103,40 @@ export default function ClientsSection() {
     </TableText>
   );
 
+  const ClientHomePage = (client: ClientRepresentation) => {
```
jonkoops, line 106: Can we extract this logic somewhere and add some unit tests? With the amount of branches in this function it feels hard to refactor without accidentally making breaking changes.

jonkoops (changes requested):

src/clients/ClientsSection.tsx
edewit: :+1: seems that @thomasdarimont is already unsure if all cases are met...

jonkoops (approved):
Let's merge this now as it fixes an actual bug and create a task to add some tests for this later.

src/clients/ClientsSection.tsx
```diff
@@ -103,6 +103,40 @@ export default function ClientsSection() {
...
+
+    if (rootUrl === "${authBaseUrl}" && baseUrl) {
+      // eslint-disable-next-line prettier/prettier
+      return rootUrl.replace("${authBaseUrl}", addTrailingSlash(adminClient.baseUrl)) + baseUrl.substring(1);
+    }
+
+    if (rootUrl === "${authAdminUrl}" && baseUrl) {
+      // eslint-disable-next-line prettier/prettier
+      return rootUrl.replace("${authAdminUrl}", addTrailingSlash(adminClient.baseUrl)) + baseUrl.substring(1);
+    }
```
edewit, line 123: If I read this correctly it means when `baseUrl` is not set and rootUrl contains either `"${authBaseUrl}"` or `"${authAdminUrl}"` then we return `undefined`? @thomasdarimont

PR #3060, opened 2022-08-05
Add utility function to join paths

jonkoops:
Adds a utility function to join paths, which is need for another future PR.


agagancarczyk (approved):

edewit (approved):
now you need to use it ;)

PR #3063, opened 2022-08-07
Add PKCE Enabled and PKCE Method to OIDC Identity Providers

ssilvert:
## Motivation
Fixes https://github.com/keycloak/keycloak-admin-ui/issues/3050

## Brief Description
Adds missing fields

## Verification Steps
1. Create OIDC or Keycloak OIDC identity provider
2. Verify `Use PKCE` and `PKCE Method` fields are saving properly

### DIFF
--- a/src/identity-providers/add/DiscoverySettings.tsx
+++ b/src/identity-providers/add/DiscoverySettings.tsx
@@ -153,7 +153,7 @@ const Fields = ({ readOnly }: DiscoverySettingsProps) => {
               <Select
                 toggleId="pkceMethod"
                 required
-                direction="up"
+                direction="down"
                 onToggle={() => setPkceMethodOpen(!pkceMethodOpen)}
                 onSelect={(_, value) => {
                   onChange(value as string);
--- a/src/identity-providers/add/DiscoverySettings.tsx
+++ b/src/identity-providers/add/DiscoverySettings.tsx
@@ -1,15 +1,19 @@
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
-import { useFormContext, useWatch } from "react-hook-form";
+import { Controller, useFormContext, useWatch } from "react-hook-form";
 import {
   ExpandableSection,
   FormGroup,
   ValidatedOptions,
+  Select,
+  SelectOption,
+  SelectVariant,
 } from "@patternfly/react-core";
 
 import { SwitchField } from "../component/SwitchField";
 import { TextField } from "../component/TextField";
 import { KeycloakTextInput } from "../../components/keycloak-text-input/KeycloakTextInput";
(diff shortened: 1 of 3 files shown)


src/identity-providers/add/DiscoverySettings.tsx
```diff
@@ -19,6 +23,8 @@ type DiscoverySettingsProps = {
 
 const Fields = ({ readOnly }: DiscoverySettingsProps) => {
   const { t } = useTranslation("identity-providers");
+  const pkceMethods = ["plain", "S256"];
```
jonkoops, line 26: Since this is a constant that is never modified you can lift it outside of the scope of the function. This will ensure that it will not get re-declared on every render. You can also add `as const` to allow TypeScript to evaluate it as a constant type.

For example:
```tsx
const PKCE_METHODS = ["plain", "S256"] as const;

type DiscoverySettingsProps = {
  readOnly: boolean;
};

const Fields = ({ readOnly }: DiscoverySettingsProps) => {}
```

src/identity-providers/add/DiscoverySettings.tsx
```diff
@@ -33,6 +39,10 @@ const Fields = ({ readOnly }: DiscoverySettingsProps) => {
     control: control,
     name: "config.useJwksUrl",
   });
+  const isPkceEnabled = useWatch({
+    control: control,
```
jonkoops, line 43: If the key and value are the same you can omit the value.
```suggestion
    control,
```

src/identity-providers/add/DiscoverySettings.tsx
```diff
@@ -118,6 +128,56 @@ const Fields = ({ readOnly }: DiscoverySettingsProps) => {
...
+            defaultValue={pkceMethods[0]}
+            control={control}
+            render={({ onChange, value }) => (
+              <Select
+                toggleId="pkceMethod"
+                required
+                direction="down"
+                onToggle={() => setPkceMethodOpen(!pkceMethodOpen)}
+                onSelect={(_, value) => {
+                  onChange(value as string);
```
jonkoops, line 159: Instead of type-casting you can call the `toString()` method, which exists both on `string` and `SelectOptionObject`.
```suggestion
                  onChange(value.toString());
```

jonkoops (approved):

src/identity-providers/add/DiscoverySettings.tsx
ssilvert: That fails at run time.  "String.prototype.toString requires that 'this' be a String"

jonkoops: Very peculiar, but yeah let's leave this then.

jonkoops (approved):

Issue #2044, opened 2022-02-10
[UX Review] Optimize the number box in the Time-policy creation page of client Authorization


xianli123:
### Describe the bug

In the **Client > Client details > Authorization > Policy > Time policy creation** page, negative numbers should not be allowed in the number box. Please check and fix it.

<img width="626" alt="Screen Shot 2022-02-10 at 2 35 41 PM" src="https://user-images.githubusercontent.com/53931519/153351335-b1456ed2-1ccf-4b90-86ff-919c5439dfc3.png">

### Version

New admin UI

dominikkawka:
@xianli123 
It seems to have been fixed, it won't let me go down to negative numbers

![image](https://user-images.githubusercontent.com/31955648/182880463-f8aad74b-e401-40a5-aa8b-80fff7cd5315.png)

jonkoops:
@dominikkawka I can still change them to negative values by manually changing the value with the keyboard.

I have to say that the UX here is quite poor though, I would not expect this to be several numerical inputs. Instead I would expect two date-like inputs with fields horizontally.

xianli123:
Hi @jonkoops , It is a good idea, but the PF4 can't support that kind of component. I found the PF4 has a numerical inputs component that can fill in negative values, FYI https://www.patternfly.org/v4/components/number-input#varying-sizes. Could we use this component?

edewit:
Can't do a date component as we are not setting two date and time, but rather part of a date. For instance I could fill out only a month to make this policy apply only when in specified month. One could leave the other fields blank.

jonkoops:
@edewit Yeah, what I mean is that I would expect the following:

- Two rows labelled 'From' and 'To' each containing the following fields in columns:
  - A drop-down filled with the names of the month (localized for user).
  - A drop-down filled with possible days of the month (not taking into account leap years).
  - A [time picker](https://www.patternfly.org/v4/components/time-picker) to select a time for that day.
  
@xianli123 WDYT?

jonkoops:
I think it doesn't matter as you are still selecting a time either way.

xianli123:
The time picker can select a time. But how to select a period of time? What do you think? @jonkoops

jonkoops:
If you have two separate rows and add a label to the side with 'From' and 'To' I think it should be user-friendly enough to be understood.

jonkoops:
@xianli123 Perhaps you can sketch something up so we can refine a design?

[...]

PR #3076, opened 2022-08-08
added check to see if value is between min and max

edewit:
fixes: #2044

## Checklist:

### DIFF
--- a/src/clients/authorization/policy/Time.tsx
+++ b/src/clients/authorization/policy/Time.tsx
@@ -93,7 +93,14 @@ type NumberControlProps = {
 
 const NumberControl = ({ name, min, max }: NumberControlProps) => {
   const { control } = useFormContext();
-  const setValue = (newValue: number) => Math.min(newValue, max);
+  const setValue = (newValue: number) => {
+    if (newValue < min) {
+      return min;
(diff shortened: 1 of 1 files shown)


dominikkawka (approved):

src/clients/authorization/policy/Time.tsx
```diff
@@ -93,7 +93,14 @@ type NumberControlProps = {
 
 const NumberControl = ({ name, min, max }: NumberControlProps) => {
   const { control } = useFormContext();
-  const setValue = (newValue: number) => Math.min(newValue, max);
+  const setValue = (newValue: number) => {
```
jonkoops, line 96: How would this work with intermediate values? For example, if the `min` is 10 and I want to type `15`, what will happen? I kind of expect that this would cause the input value to suddenly jump to `10` unexpectedly.

edewit: Okay I don't understand what is expected then it that case. In the issue you put that you don't want to be able to input invalid numbers like `-1`, but you also don't want the your invalid input to change to a valid one. I mean the interaction is similar for the plus and min buttons. When I click the `+` button but the value is already at max nothing happens why would that be expected?

PR #3079, opened 2022-08-08
initial version of cypress component test

edewit:
## Checklist:

### DIFF
--- a/cypress/component/ConfirmDialogModal.cy.tsx
+++ b/cypress/component/ConfirmDialogModal.cy.tsx
@@ -1,3 +1,4 @@
+import { mount } from "cypress/react";
 import { ConfirmDialogModal } from "../../src/components/confirm-dialog/ConfirmDialog";
 
 describe("ConfirmDialogModal", () => {
@@ -6,7 +7,8 @@ describe("ConfirmDialogModal", () => {
   it("should mount", () => {
     const toggle = cy.spy().as("toggleDialogSpy");
     const confirm = cy.spy().as("onConfirmSpy");
-    cy.mount(
+
+    mount(
       <ConfirmDialogModal
         continueButtonLabel="Yes"
         cancelButtonLabel="No"
@@ -26,5 +28,3 @@ describe("ConfirmDialogModal", () => {
     cy.get("@toggleDialogSpy").should("have.been.called");
   });
 });
-
-export {};
--- a/cypress/component/ConfirmDialogModal.cy.tsx
+++ b/cypress/component/ConfirmDialogModal.cy.tsx
@@ -1,0 +1,30 @@
+import { ConfirmDialogModal } from "../../src/components/confirm-dialog/ConfirmDialog";
+
+describe("ConfirmDialogModal", () => {
+  const bodySelector = "#pf-modal-part-2";
+
+  it("should mount", () => {
+    const toggle = cy.spy().as("toggleDialogSpy");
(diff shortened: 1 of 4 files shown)


cypress/component/ConfirmDialogModal.cy.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    cy.get(bodySelector).should("have.text", "Some text");
+    cy.findByTestId("confirm").click();
+    cy.get("@onConfirmSpy").should("have.been.called");
+    cy.findAllByTestId("cancel").click();
+    cy.get("@toggleDialogSpy").should("have.been.called");
+  });
+});
+
+export {};
```
jonkoops, line 30: Can we remove this empty export statement?
```suggestion
```

cypress/support/component-index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
jonkoops, line 5: This is only needed for Internet Explorer, so I think we can get rid of it.
```suggestion
```

cypress/support/component.ts
```diff
@@ -0,0 +1,41 @@
...
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import "./commands";
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
+
```
jonkoops, line 21: We're never going to do this.
```suggestion
```

cypress/support/component.ts
```diff
@@ -0,0 +1,41 @@
...
+// with a <reference path="./component" /> at the top of your spec.
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      mount: typeof mount;
+    }
+  }
+}
+
+Cypress.Commands.add("mount", mount);
```
jonkoops, line 38: Instead of making this a global command I'd prefer to simply import the `mount` function from `cypress/react` in every test.

jonkoops (changes requested):

cypress/component/ConfirmDialogModal.cy.tsx
edewit: seems it doesn't work without this, maybe we need some config in order to be able to remove it?

cypress/support/component.ts
edewit: it's not global it adds it to the `cypress` object

cypress/component/ConfirmDialogModal.cy.tsx
edewit: I can be removed... don't know why it was complaining in the first place... removed it now

cypress/support/component.ts
jonkoops: Which is global 😉

cypress/component/ConfirmDialogModal.cy.tsx
jonkoops: Probably it was that this module was not importing or exporting anything, which is the bare minimum that is expected.

jonkoops (approved):

jonkoops:
@edewit merged this PR, but be aware that these tests might not be running on CI at the moment.

Issue #2664, opened 2022-05-17
Take a close look at flaky tests


ssilvert:
### Describe the bug

Right now, we have disabled the check for flaky tests.  So if flaky tests eventually pass, the PR will show green.

After Keycloak 19, we need to take a closer look at flaky tests.

### Version

19.0.0-SNAPSHOT

### Anything else?

More info:

Erik Jan de Wit, 8:38 AM
btw @Stan Silvert seems like `should create Test Disabled realm ` is a very frequent flacky test

maybe we should take a good look at the realm settings section and improve it

Stan Silvert, 8:40 AM
@Erik Jan de Wit Do you mean take a look at the tests or the realm settings itself?

Erik Jan de Wit, 10:04 AM
the realm settings itself, there are some inconsistent ways of loading the data

that might be the cause of the flackyness

PR #3091, opened 2022-08-10
Fix failing and flaky tests

ikhomyn:
## Motivation
Fixes some failing and flaky tests. Also some fixed some skipped tests. Related to #2664

## Checklist:

### DIFF
--- a/cypress/e2e/group_test.spec.ts
+++ b/cypress/e2e/group_test.spec.ts
@@ -46,9 +46,7 @@ describe("Group test", () => {
     );
   });
 
-  after(async () => {
-    await adminClient.deleteGroups();
-  });
+  after(() => adminClient.deleteGroups());
 
   beforeEach(() => {
     keycloakBefore();
--- a/cypress/e2e/group_test.spec.ts
+++ b/cypress/e2e/group_test.spec.ts
@@ -46,6 +46,10 @@ describe("Group test", () => {
     );
   });
 
+  after(async () => {
+    await adminClient.deleteGroups();
+  });
+
   beforeEach(() => {
     keycloakBefore();
     loginPage.logIn();
--- a/cypress/e2e/users_test.spec.ts
+++ b/cypress/e2e/users_test.spec.ts
@@ -112,8 +112,6 @@ describe("User creation", () => {
       .fillPasswordForm()
       .clickConfirmationBtn()
       .clickSetPasswordBtn();
-
-    masthead.closeAllAlertMessages();
   });
 
   it("Search existing user test", () => {
@@ -173,6 +171,7 @@ describe("User creation", () => {
   });
 
   it("Add user to groups test", () => {
+    masthead.closeAllAlertMessages();
     // Go to user groups
     listingPage.searchItem(itemId).itemExist(itemId);
     listingPage.goToItemDetails(itemId);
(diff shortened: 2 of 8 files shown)


cypress/e2e/group_test.spec.ts
```diff
@@ -46,6 +46,10 @@ describe("Group test", () => {
     );
   });
 
+  after(async () => {
+    await adminClient.deleteGroups();
+  });
```
jonkoops, line 51: Since a Promise is returned here this can be shortened a bit.
```suggestion
  after(() => adminClient.deleteGroups());
```

cypress/e2e/users_test.spec.ts
```diff
@@ -250,10 +251,13 @@ describe("User creation", () => {
       .clickCloseDataDialogBtn();
   });
 
-  it.skip("Delete credential", () => {
+  it("Delete credential", () => {
     listingPage.goToItemDetails(itemIdWithCred);
     credentialsPage.goToCredentialsTab();
+    masthead.closeAllAlertMessages();
+    cy.wait(2000);
```
jonkoops, line 258: Any chance we can remove this wait? Perhaps by verifying that indeed all notification have been dismissed from the DOM?

cypress/support/pages/admin_console/Masthead.ts
```diff
@@ -64,7 +64,9 @@ export default class Masthead extends CommonElements {
   }
 
   closeAllAlertMessages() {
-    cy.get(this.closeAlertMessageBtn).click({ force: true, multiple: true });
+    cy.get(this.closeAlertMessageBtn).each(() => {
+      cy.get(this.closeAlertMessageBtn).click({ force: true, multiple: true });
+    });
```
jonkoops, line 69: I believe the following code (same as before) should yield the same result. At least according to the [Cypress documentation](https://docs.cypress.io/api/commands/click#Click-all-elements-with-id-starting-with-btn).
```suggestion
    cy.get(this.closeAlertMessageBtn).click({ force: true, multiple: true });
```

jonkoops (changes requested):
Awesome work fixing the flaky tests, it does seems like some tests are still failing though.

cypress/support/pages/admin_console/Masthead.ts
ikhomyn: I've tried this way. But the problem is. When Cypress perform cy.get(this.closeAlertMessageBtn).click({ force: true, multiple: true });  it finds multiple elements present, after first notification is closed. Dom is updated, and it loses other elements. So in this case it will only close last notification, but not all of them. So after each click, I have to find again the elements.

cypress/e2e/users_test.spec.ts
ikhomyn: I've tried without timeout but seems it still need some extra time in order to update dom or some refresh request is made on the table. And without timeout it causes the problem that Cypress loses focus on opened dropdown items.
I mean:
- All notifications are closed
- Click on item dropdown element. Dropdown options are shown
- Meanwhile list or Dom is updated and focus is lost, dropdown options are hidden
- Attempt to click on dropdown option which is not longer visible.

I've tried several ways, but only adding timeout fixed the problem.

ikhomyn:
@jonkoops Regarding the other 2 failing tests, I was not sure about the flow of the test in order to fix them.

ikhomyn:
@jonkoops I have updated 2 remaining failing tests for clients_saml_test.spec. But 1 will still fail as of this issue: https://github.com/keycloak/keycloak-admin-ui/issues/3092

cypress/support/pages/admin_console/Masthead.ts
jonkoops: Ahhh, yeah that makes sense. Let's keep it like this then.

cypress/e2e/users_test.spec.ts
jonkoops: Hmmm, well if it's too complex let's just keep it like this for now. Either way it's already a massive improvement.

jonkoops (approved):

PR #3095, opened 2022-08-12
initial version of an endpoint

### LINKED ISSUE #2986: Cannot see roles of user when lots of clients
### Describe the bug

When you have "lots" of clients (I have 75+ clients), listing the roles of a user takes "forever"

### Version

18

### Expected behavior

Listing roles of user should take a maximum of some seconds (like 2s) independently of the number of Clients

### Actual behavior

UI does 2 XHR for each client 
* /admin/realms/xxxx/users/12bd2526-baf3-49a9-8836-36fcccbe9a9c/role-mappings/clients/d1b505a2-f2a8-40e3-81e2-2d489fee1e64
* /admin/realms/xxx/users/12bd2526-baf3-49a9-8836-36fcccbe9a9c/role-mappings/clients/d1b505a2-f2a8-40e3-81e2-2d489fee1e64/composite

So when there are 75 clients, UI performs 150+ XHR requests before displaying the results, so obviously it takes some time

### How to Reproduce?

Create 75 clients, go to Users > Role mapping, wait "forever"

Please note that among my clients, only 3 of them are using role-mapping in my setup so I really wait for "nothing"

### Anything else?

Generally speaking, there are LOT of places in the new UI where each item in a list triggers one (or more) XHR which leads to very poor UX ...

For ex, I already declared #2961 where this is the exact same bug in the Users list

Here, if role mapping is by client, then the UI should display the list of clients and user should select / search which client he is interested in

Otherwise, you should add a new method in the ADMIN REST API in order to list all roles of a user in any client in order to perform a single DB request instead of 150+

edewit:
Closes #2986.

### DIFF
--- a/keycloak-theme/pom.xml
+++ b/keycloak-theme/pom.xml
@@ -38,6 +38,7 @@
 
         <keycloak.version>19.0.1</keycloak.version>
         <kotlin.version>1.7.10</kotlin.version>
+        <mapstruct.version>1.5.2.Final</mapstruct.version>
     </properties>
 
     <licenses>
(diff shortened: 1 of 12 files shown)


keycloak-theme/pom.xml
```diff
@@ -34,6 +35,9 @@
         <nexus.staging.plugin.version>1.6.13</nexus.staging.plugin.version>
         <frontend.maven.plugin.version>1.12.1</frontend.maven.plugin.version>
         <frontend.maven.plugin.nodeVersion>v18.7.0</frontend.maven.plugin.nodeVersion>
+
+        <keycloak.version>19.0.1</keycloak.version>
```
jonkoops, line 39: I believe we are also publishing the nightlies as snapshot versions of the Keycloak server. MIght be something to look into at a later point.

keycloak-theme/src/main/java/org/keycloak/admin/ui/rest/AdminUIExtendedResource.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.util.*
+import java.util.function.Function
+import java.util.stream.Collectors
+import java.util.stream.Stream
+import javax.ws.rs.*
+import javax.ws.rs.core.Context
+import javax.ws.rs.core.MediaType
+
+
+open class AdminUIExtendedResource(
```
jonkoops, line 21: Perhaps we should also add OpenAPI to this immediately?

jonkoops (approved):

keycloak-theme/pom.xml
edewit: I rather stick with something that is a bit more stable... we don't really need the latest and greatest here

jonkoops (approved):

edewit:
fixes: #2986

PR #3127, opened 2022-08-16
changed header translation

ikhomyn:
## Motivation
See #3108. Not sure if is the most right solution, but it works.

## Checklist:

### DIFF
--- a/src/components/view-header/ViewHeader.tsx
+++ b/src/components/view-header/ViewHeader.tsx
@@ -71,7 +71,7 @@ export const ViewHeader = ({
   helpTextKey,
   isReadOnly = false,
 }: ViewHeaderProps) => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
   const { enabled } = useHelp();
   const [isDropdownOpen, setDropdownOpen] = useState(false);
   const [isLowerDropdownOpen, setIsLowerDropdownOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


jonkoops (approved):
This seems like a good solution to me :+1:

src/components/view-header/ViewHeader.tsx
```diff
@@ -95,7 +95,7 @@ export const ViewHeader = ({
               <LevelItem>
                 <TextContent className="pf-u-mr-sm">
                   <Text className={className} component="h1">
-                    {t(titleKey)}
+                    {i18n.exists(titleKey) ? t(titleKey) : titleKey}
```
edewit, line 98: This should be the default behaviour or the `t` function when the key does not exist it should return the key

edewit (approved):

src/components/view-header/ViewHeader.tsx
ikhomyn: I've tried but it doesn't work, it returns value this way
Screenshot from 2022-08-16 12-53-47

jonkoops: I believe the keys are transformed somehow by `react-i18n`.

edewit: yeah it's because of the `:` in the name it takes it as a namespace and strips that off

PR #3129, opened 2022-08-16
Realm email doesn't default to port 25

### LINKED ISSUE #3097: Realm Email doesn't default to port 25
### Describe the bug

Email configuration of the realm doesen't use the default port:

image

### Version

19.0.1

### Expected behavior

Emails should be able to be send with the default port 25 but it needs to be explicitly set 

### Actual behavior

Keycloak tries to send mails without a port:

```bash
2022-08-10 14:33:19,499 ERROR [org.keycloak.services.resources.admin.RealmAdminResource] (executor-thread-1095) Failed to send email
 com.sun.mail.util.MailConnectException: Couldn't connect to host, port: smtp.*****.com, 0; timeout 10000;
  nested exception is:
	java.net.SocketTimeoutException: connect timed out
```

### How to Reproduce?

1. configure SMTP settings and leave port to default (no entry)
2. try to send mail and watch the errors in the logs
3. set the port 25 and redo no. 2

### Anything else?

_No response_

ssilvert:
## Motivation
Fixes #3097 

## Brief Description
In `Realm Settings --> Email`, the port was not being saved properly if left blank.  The back end is expecting a `null` value to signify default port.

## Verification Steps
See #3097

### DIFF
--- a/src/realm-settings/EmailTab.tsx
+++ b/src/realm-settings/EmailTab.tsx
@@ -69,6 +69,10 @@ export const RealmSettingsEmailTab = ({
       }
 
       const savedRealm = { ...realm, ...form };
+
+      // For default value, back end is expecting null instead of empty string
+      if (savedRealm.smtpServer?.port === "") savedRealm.smtpServer.port = null;
+
       await adminClient.realms.update({ realm: realmName }, savedRealm);
       setRealm(savedRealm);
       addAlert(t("saveSuccess"), AlertVariant.success);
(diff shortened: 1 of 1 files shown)


src/realm-settings/EmailTab.tsx
```diff
@@ -93,6 +97,9 @@ export const RealmSettingsEmailTab = ({
       serverSettings[key] = mapperFn(serverSettings[key]);
     }
 
+    // For default value, back end is expecting null instead of 0
+    if (serverSettings["port"] === 0) serverSettings["port"] = null;
```
jonkoops, line 101: Should be able to do a regular property access here.
```suggestion
    if (serverSettings.port. === 0) serverSettings.port = null;
```

jonkoops (changes requested):

src/realm-settings/EmailTab.tsx
ssilvert: I tried that.  VS Code didn't recognize the property.

serverSettings is of type `[x: string]: any;`

jonkoops (approved):

src/realm-settings/EmailTab.tsx
jonkoops: Seemed to work here, I added a commit for it.

PR #3187, opened 2022-08-24
Force refresh of user after save to avoid modal (#3180)

### LINKED ISSUE #3180: Cannot update user on new Keycloak Administration Console theme
### Describe the bug

Hi guys, 
I am using Keycloak 19 for my project to manage users. I found the bug that the user's firstname, and lastname unable to update.

Let's see the below video:

[Keycloak Administration Console.webm](https://user-images.githubusercontent.com/25485764/185541335-bb9ecaf3-8ea5-40de-acaa-9e488e06f9a8.webm)
Keycloak Administration Console

### Version

19.0.1

### Expected behavior

The user can update.

### Actual behavior

The user cannot update when changing to the tab attribute.

### How to Reproduce?

1. In Details tab update First name, Last name.
2. Click Save.
3. You can see the message successfully.
4. Change to Attributes tab.
5. Change the attribute key.
6. Click Save.
7. Back to the user details the first name and last name no change.

### Anything else?

_No response_

kris2kris:
## Motivation
Closes #3180 

## Brief Description
The problem is that field stay dirty after save, so I force refresh of user after an update. Maybe there is a better way to reset dirty field, for exemple display a spinner instead of form during save.

## Verification Steps
1. Update user firstname and save.
2. Switch to Attributes tab.
3. Do it once again if problem doesn't occurred.

## Checklist:

### DIFF
--- a/src/user/UsersTabs.tsx
+++ b/src/user/UsersTabs.tsx
@@ -46,6 +46,7 @@ const UsersTabs = () => {
   const [user, setUser] = useState<UserRepresentation>();
   const [bruteForced, setBruteForced] = useState<BruteForced>();
   const [addedGroups, setAddedGroups] = useState<GroupRepresentation[]>([]);
+  const [forceFetch, setForceFetch] = useState(false);
 
   useFetch(
     async () => {
@@ -72,7 +73,7 @@ const UsersTabs = () => {
       setBruteForced(bruteForced);
       user && setupForm(user);
     },
-    [user?.username]
+    [user?.username, forceFetch]
   );
 
   const setupForm = (user: UserRepresentation) => {
@@ -90,6 +91,7 @@ const UsersTabs = () => {
       if (id) {
         await adminClient.users.update({ id }, user);
         addAlert(t("userSaved"), AlertVariant.success);
+        setForceFetch(!forceFetch);
       } else {
         const createdUser = await adminClient.users.create(user);
 
--- a/src/user/UsersTabs.tsx
+++ b/src/user/UsersTabs.tsx
@@ -74,7 +74,7 @@ const UsersTabs = () => {
       setBruteForced(bruteForced);
       user && setupForm(user);
     },
-    [user?.username, forceFetch]
+    [user?.username, refreshCount]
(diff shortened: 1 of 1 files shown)


src/user/UsersTabs.tsx
```diff
@@ -46,6 +46,7 @@ const UsersTabs = () => {
   const [user, setUser] = useState<UserRepresentation>();
   const [bruteForced, setBruteForced] = useState<BruteForced>();
   const [addedGroups, setAddedGroups] = useState<GroupRepresentation[]>([]);
+  const [forceFetch, setForceFetch] = useState(false);
```
jonkoops, line 49: We're already using a pattern for this that uses a count to force a refresh. Let's keep that aligned with what we are already doing.

```suggestion
  const [refreshCount, setRefreshCount] = useState(0);
  const refresh = () => setRefreshCount((count) => count + 1);
```

src/user/UsersTabs.tsx
```diff
@@ -72,7 +73,7 @@ const UsersTabs = () => {
       setBruteForced(bruteForced);
       user && setupForm(user);
     },
-    [user?.username]
+    [user?.username, forceFetch]
```
jonkoops, line 76:
```suggestion
    [user?.username, refreshCount]
```

src/user/UsersTabs.tsx
```diff
@@ -90,6 +91,7 @@ const UsersTabs = () => {
       if (id) {
         await adminClient.users.update({ id }, user);
         addAlert(t("userSaved"), AlertVariant.success);
+        setForceFetch(!forceFetch);
```
jonkoops, line 94:
```suggestion
        refresh();
```

jonkoops (changes requested):

src/user/UsersTabs.tsx
kris2kris: It's ok for me, I test with the change and it's still working

jonkoops (approved):

edewit (approved):

PR #3189, opened 2022-08-25
Add Keycloak JS to the project

jonkoops:
Adds Keycloak JS to the project and includes it in the [workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces). Eventually a Maven build wil be added so it can be published and included in the Keycloak server build, at which point the code can be removed from that project.


keycloak-theme/pom.xml
```diff
@@ -237,7 +237,7 @@
                             <goal>npm</goal>
                         </goals>
                         <configuration>
-                            <arguments>run build --workspace=admin-ui</arguments>
+                            <arguments>run build --workspace=keycloak-js --workspace=admin-ui</arguments>
```
edewit, line 240: Why not build all workspaces here? like this:
```suggestion
                            <arguments>run build --workspaces</arguments>
```

edewit (approved):

keycloak-theme/pom.xml
jonkoops: Since we are possibly adding more in the future I decided to be explicit here for now. Likely this will be replaced with some toolt that can track dependencies for us (such as [wireit](https://github.com/google/wireit)), so it's temporary either way.

PR #3195, opened 2022-08-26
use avatar picture when it exists

edewit:
The default mapper uses `picture` as field in the token so if it's there let's use that instead
image
I've used github as Idp:
image


jonkoops (approved):

ssilvert:
@edewit Forgot to tell you that I think this is really cool.  Nice work.

PR #3240, opened 2022-08-30
Added an inline Alert for Signed Jwt authenticator in Credentials #1850

agagancarczyk:
## Motivation
Closes https://github.com/keycloak/keycloak-ui/issues/1850
Fixes as per designs https://marvelapp.com/prototype/695h94c/screen/72266137

Note: I had to introduce a custom CSS class because patternfly was only offering a max `p-u-ml-4xl` and I needed more margin spacing.

## Verification Steps
1. Go to `Clients >> Select Client >> Credentials tab >> Select Signed Jwt`.
2. Verify that the Alert is present.
3. Click on `Save` btn and verify that the Alert is removed.

## Checklist:


apps/admin-ui/src/clients/credentials/credentials.css
```diff
@@ -0,0 +1,9 @@
+.kc-signedJWTAlert {
+  margin: 0.1rem 0 0 10.4rem;
+}
+
+@media (max-width: 768px) {
+  .kc-signedJWTAlert {
+    margin: 0.5rem 0 0 0;
+  }
+}
```
jonkoops, line 9: Instead of writing custom CSS we could perhaps use PatternFly's [utility classes](https://www.patternfly.org/v4/utilities/spacing).

jonkoops (changes requested):

apps/admin-ui/src/clients/credentials/credentials.css
agagancarczyk: Thanks @jonkoops yes, I agree and it was my preferred approach as well. I tried to do something like this `p-u-ml-4xl` and 'pf-u-float-right p-u-mr-4xl', but it seems that 4xl is not enough to push the Alert to the right place.

jonkoops: If it's not possible then this seems allright to me.

jonkoops (approved):

PR #3258, opened 2022-09-01
Optimize dependencies with Rollup

jonkoops:
Optimize dependencies using Rollup instead of esbuild. This prevents issues where a dependency can be duplicated and create errors at runtime.


edewit (approved):

edewit:
I'm not sure going this route is the best way forward, I think it's better to not use `esbuild` in production

PR #3260, opened 2022-09-02
Add nightly run using nightly builds

stianst:
Added a nightly run of Cypress to test the nightly built artefacts rather than building it within the GA workflow. This serves two purposes:

a) A nightly check of KC + admin-ui even if there are no pushes to this repo
b) Check KC as it will be built by the release process

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -4,6 +4,8 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  schedule:
+    - cron: '0 4 * * *'
   workflow_dispatch:
     inputs:
       keycloakRepo:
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress.yml
```diff
@@ -60,6 +63,29 @@ jobs:
...
+            echo "::set-output name=path::server/keycloak-999-SNAPSHOT"
+
+        - name: Cache Keycloak server
+          uses: actions/cache@v3
+          with:
+            path: ${{ steps.extract-server.outputs.path }}
+            key: ${{ github.sha }}
+
   test:
     needs: setup
```
jonkoops, line 90: This dependency now no longer exists.

jonkoops (changes requested):

.github/workflows/cypress.yml
stianst: Not following, could you elaborate?

ssilvert: @stianst FYI Jon is out this week.

jonkoops: This job [depends on](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) the `setup` job, which was removed [here](https://github.com/keycloak/keycloak-ui/pull/3260/files/713a363e8fd83af660212479f4fa74d3beb2d5da#diff-5da9a41590fb261de0e464c95862528100c42c77dffad53b7c6de228b6afd2bdL8).

stianst: LOL, of course ;)

Fixed this now by having a need on both jobs, then running it if either passes. A bit weird, but have tested it works (https://github.com/stianst/playground/actions/workflows/main.yml)

jonkoops (approved):

.github/workflows/cypress.yml
jonkoops: Oh yeah, that makes sense. Even if it's a bit counterintuitive.

jonkoops (approved):

.github/workflows/cypress.yml
```diff
@@ -69,8 +72,32 @@ jobs:
...
+    outputs:
+      server-path: ${{ steps.extract-server.outputs.path }}
+    env:
+      KEYCLOAK_SERVER_PATH: keycloak-server
+    steps:
+        - name: Download Keycloak
+          id: extract-server
+          working-directory: ${{ env.KEYCLOAK_SERVER_PATH }}
+          run: |
+            wget -q https://github.com/keycloak/keycloak/releases/download/nightly/keycloak-999-SNAPSHOT.tar.gz
```
stianst, line 87:
```suggestion
            gh release -R keycloak/keycloak download nightly -p keycloak-999-SNAPSHOT.tar.gz
```

jonkoops (approved):

PR #3373, opened 2022-09-21
Adjusted initialization of admin-client-ui with different urls in hostname and hostname-admin

srose:
## Motivation
Fix https://github.com/keycloak/keycloak-ui/issues/3374 

## Brief Description
Use the "correct" URLs for initialization.

## Verification Steps
Start keycloak with --hostname and a different --hostname-admin.
Try to access the admin-konsole.
Either CORS Issues from calling --hostname based URLs from within the --hostname-admin or the calls go to --hostname without error.
The --hostname-admin must be used for all ajax calls from within the admin-console.

## Checklist:

## Additional Notes
Work in progress. Missing tests, but want to verify the direction this is heading.

### DIFF
--- a/apps/admin-ui/src/context/auth/AdminClient.tsx
+++ b/apps/admin-ui/src/context/auth/AdminClient.tsx
@@ -85,7 +85,7 @@ export async function initAdminClient() {
   const adminClient = new KeycloakAdminClient();
 
   adminClient.setConfig({ realmName: environment.loginRealm });
-  adminClient.baseUrl = keycloak.authServerUrl ?? environment.authServerUrl;
+  adminClient.baseUrl = environment.authUrl;
   adminClient.registerTokenProvider({
     async getAccessToken() {
       try {


jonkoops (approved):

jonkoops:
I'm approving this as it resolves an issue that users are having in production. However it would be nice if we can regression test this issue in CI somehow.

edewit (approved):

XcrigX:
I see this commit under the 19.0.3 tag. I upgraded to the the 19.0.3 docker image to get this fix, but when accessing the admin console from _hostname-admin_ the logon is redirecting to _hostname_ for the login screen, then back to hostname-admin after authentication. 
I assume that's not correct?

stianst:
> I see this commit under the 19.0.3 tag. I upgraded to the the 19.0.3 docker image to get this fix, but when accessing the admin console from _hostname-admin_ the logon is redirecting to _hostname_ for the login screen, then back to hostname-admin after authentication. 
> I assume that's not correct? 
> @jonkoops @stianst 
> 

Yes

srose:
Imho thats correct. The admin-ui is another application protected by keycloak. When using the admin-ui by accessing hostname-admin a redirect to the hostname is exactly what should happen.

XcrigX:
My thought was that this commit would make all calls from admin console completely remain on the _hostname-admin_- in most cases this means they remain within the intranet and don't go through the public hostname/proxy. 
I'm not informed enough to understand if that is reasonable/desirable. I may just be misunderstanding what this PR does. 

For context - I am debugging a different issue where I'm getting an _invalid_token_ error when logging into the admin console leading to a re-direct loop. I suspect my proxy/firewall is truncating the payload on the post to _/token_ coming through the _hostname_. I don't have direct access to the firewall to verify this or to see why however, so was hoping this commit was an easy workaround without the need to open a ticket with the infra team who manages the firewall.

XcrigX:
Your response makes sense @srose. 
I don't want to clutter up this thread too much with an unrelated issue - but I just got off a troubleshooting call with our network security team. Fortiweb is flagging the _/token_ request as malicious due to the cookie - "cookie signed verification failed". We don't know exactly what about the cookie it doesn't like, but if we disable cookie security at the firewall it works fine. 
I suspect this may be because the admin hostname is not resolvable for Fortiweb, or perhaps the public/private hostnames are being mixed somewhere perhaps (I'm using 18.0.2 in this setup) ? 

The traffic pattern is like this:   public hostname  (Foriweb firewall/gateway) -> internal "proxy" (Azure API Manager) -> private hostname (Azure load balancer with a private DNS name pointing to a KC running in kubernetes). 

So not necessarily a Keycloak problem - but possibly a problem for some users in production setups. 
If I determine exactly what the cookie problem is I'll report back.

PR #3385, opened 2022-09-23
account ui

edewit:
- initial version
- added initial version of masthead lib
- changed to let account-ui use masthead
- fixed build
- added default dropdown items
- added wireit
- * added way to load data using fetch * initial AccountPage
- finished account page
- added signing in page
- added delete credential
- added alerts
- added device activity page

Initial version of the account ui introduces a lot of new concepts, to review it might be easier to take a look at the individual commits


ssilvert (approved):
I think we should go ahead and merge this as a baseline.  But I'll give others a chance to chime in.

.gitignore
```diff
@@ -34,3 +34,6 @@ server
 
 # Wireit
 .wireit
+
+# Build
+dist/
```
jonkoops, line 39: Let's add this to a dedicated `.gitignore` in `apps/account-ui` like we're already doing for the Admin UI. Reason for this is that Keycloak JS cannot fully ignore the `dist` directory, as it contains files that should be checked in.

apps/account-ui/cypress/cypress.config.mjs
```diff
@@ -0,0 +1,59 @@
+import webpackPreprocessor from "@cypress/webpack-batteries-included-preprocessor";
```
jonkoops, line 1: I was thinking to put the E2E tests in it's own app (e.g. `account-ui-e2e`).

jonkoops, line 1: We might reduce duplication by creating a base config file for Cypress in the root of the project and extend it here and in the Admin UI (see [Cypress docs](https://www.cypress.io/blog/2020/06/18/extending-the-cypress-config-file/)).

apps/account-ui/package.json
```diff
@@ -0,0 +1,80 @@
+{
```
jonkoops, line 1: There should only be a single package lock file in the root of the project, this one should be removed. (commented here because diff is too large on the other file).

apps/account-ui/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "dev": "wireit",
+    "build": "wireit",
+    "preview": "wireit",
+    "lint": "wireit",
+    "test": "wireit",
+    "cy:open": "cypress open --e2e --browser chrome",
+    "cy:run": "cypress run --browser chrome",
+    "cy:check-types": "tsc --project cypress/tsconfig.json"
+  },
+  "dependencies": {
```
jonkoops, line 13: Small nit: the `wireit` section is directly under the `scripts` section in the other projects.

apps/account-ui/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "@types/react-dom": "^17.0.16",
+    "@types/react-router-dom": "^5.3.3",
+    "@vitejs/plugin-react": "^2.1.0",
+    "cypress": "^10.6.0",
+    "vite": "^3.1.0",
+    "vite-plugin-checker": "^0.5.1",
+    "vitest": "^0.23.1",
+    "webpack": "^5.74.0"
+  },
+  "wireit": {
```
jonkoops, line 48: All of these tasks should also rely on `keycloak-js`'s build.

apps/account-ui/src/app.module.css
```diff
@@ -0,0 +1,4 @@
+
+img.logo {
```
jonkoops, line 2: Let's not make this selector overly specific.
```suggestion
.logo {
```

apps/account-ui/public/resources/content.json
```diff
@@ -0,0 +1,72 @@
+[
```
jonkoops, line 1: Several of the fields in this JSON file are never actually used. Perhaps we can remove them for now instead of giving the illusion that these do anything.

apps/account-ui/package.json
jonkoops, line 13: Are we sure all of these dependencies are being used?

apps/account-ui/public/resources/content.json
jonkoops, line 1: Why are we doing `content.json`? I thought we were going to skip this customization for now?

libs/masthead/vite.config.ts
```diff
@@ -0,0 +1,51 @@
...
+export default defineConfig({
+  resolve: {
+    mainFields: ["module"],
+    dedupe: ["react", "react-dom"],
+  },
+  build: {
+    lib: {
+      entry: path.resolve(__dirname, "src/index.ts"),
+      name: "KeycloakMasthead",
+      formats: ["es", "umd"],
```
jonkoops, line 18: Let's not build this as a UMD library.
```suggestion
      formats: ["es"],
```

libs/masthead/vite.config.ts
```diff
@@ -0,0 +1,51 @@
...
+      fileName: (format) => `keycloak-masthead.${format}.js`,
+    },
+    rollupOptions: {
+      external: ["react", "react-dom"],
+      output: {
+        globals: {
+          react: "React",
+          "react-dom": "ReactDOM",
+        },
+      },
```
jonkoops, line 28: We should not be exposing these globals.
```suggestion
```

jonkoops (changes requested):

jonkoops:
I've left a partial review, but I feel like this PR is way too big. I'd like to see this split up into some smaller PRs.

apps/account-ui/package.json
edewit: yes

apps/account-ui/package.json
edewit: well masthead relies on `keycloak-js`

apps/account-ui/public/resources/content.json
edewit: we have to have the structure of the menu somewhere right?

apps/account-ui/src/app.module.css
edewit: it's needed otherwise it doesn't override patternfly

ssilvert:
> I've left a partial review, but I feel like this PR is way too big. I'd like to see this split up into some smaller PRs.

Normally I would agree wholeheartedly.  And if @edewit has an easy way to split it up that would be nice.

But there are also advantages of this port-and-refactor approach where we have quite a bit of working code to start with.  Then any of us can start sending PRs to improve it.

edewit:
I'll close this and create a PR per commit that is 13 PRs

apps/account-ui/public/resources/content.json
edewit: it's only `modulePath` that was use do to dynamically load scripts

ssilvert:
> I'll close this and create a PR per commit that is 13 PRs

@edewit I hope you are being sarcastic.  I was fine with it the way it was.

edewit:
Yes, I am, created a new PR that just only introduces the new modules. After that I'll create another PR that adds all the code

PR #3396, opened 2022-09-26
only load realms when dropdown is opened

edewit:
this will not totally fix the issue but make the console at least usable

fixes: #3381

## Checklist:

### DIFF
--- a/apps/admin-ui/src/components/realm-selector/RealmSelector.tsx
+++ b/apps/admin-ui/src/components/realm-selector/RealmSelector.tsx
@@ -7,6 +7,7 @@ import {
   DropdownItem,
   DropdownToggle,
   Label,
+  Spinner,
   Split,
   SplitItem,
 } from "@patternfly/react-core";
@@ -27,7 +28,7 @@ import "./realm-selector.css";
 
 export const RealmSelector = () => {
   const { realm } = useRealm();
-  const { realms } = useRealms();
+  const { realms, refresh } = useRealms();
   const { whoAmI } = useWhoAmI();
   const [open, setOpen] = useState(false);
   const [search, setSearch] = useState("");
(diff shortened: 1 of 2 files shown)


ssilvert:
> How was this handled in the old console? Surely we had to load the realms there as well?

The old console does http://localhost:8180/admin/realms?briefRepresentation=true

But the briefRepresentation only helps the amount of data returned and not the number of records.  The old console also has problems handling a large number of realms.

jonkoops:
In that case, could we not fix this issue better by adding a new API that ensures these responses are paginated? Or perhaps we can return a list with sparse fields (e.g. the `id` and `name`)?

ssilvert:
> In that case, could we not fix this issue better by adding a new API that ensures these responses are paginated? Or perhaps we can return a list with sparse fields (e.g. the `id` and `name`)?

Sparse fields is what `briefRepresentation` does.  @edewit Are we using that already?

But to really fix the problem, we need pagination.

edewit:
Yes we have `briefRepresentation` but it's not enough until we have a better API let's do this workaround so people can still use the console

jonkoops:
I have no problem merging this, but do we have an issue to track a better improvement for this?

edewit:
Don't have a specific issue for it, but it's part of a bigger effort to make keycloak work with multiple tenants and realms

jonkoops:
Then we should have an epic for this perhaps. @ssilvert?

ssilvert:
@edewit Does this fully fix the issue cited?  If not, maybe we should keep the issue open or open a new one.

@jonkoops As for starting an Epic, it's not a bad idea, but I'm not sure how to define the scope of the work.

ssilvert:
@edewit Does this fully fix the issue cited?  If not, maybe we should keep the issue open or open a new one.

@jonkoops As for starting an Epic, it's not a bad idea, but I'm not sure how to define the scope of the work.

jonkoops (approved):
I am approving this since having this is better than having nothing at all. However we should definitely come up with something better in the future.

jonkoops:
@edewit could you make sure the test failures we are seeing here are not related to your changes? AFAIK we can merge this after.

edewit:
yes, tests are fine!

Issue #3316, opened 2022-09-13
When creating a new realm, can't upload file via "Browse..."


ahus1:
### Describe the bug

When creating a new realm via the UI and using a file picked via "Browse...".

It seems to me that it worked last week / some days ago. 

Workaround: <strike>Copy the file contents manually to the input field.</strike> (content pasted doesn't make it to the server AKA is missing in the request)

Tested on Fedora 36 with Chrome 105.0.5195.102

![image](https://user-images.githubusercontent.com/3957921/189875664-047fa2cf-d73d-4577-8c2d-2ea9fc80198d.png)

### Version

keycloak-admin-ui-999-20220913.041401-171.jar

### Expected behavior

Should succeed in uploading the file and showing the contents of the JSON in the UI.

### Actual behavior

No file contents visible in the UI, exception in the console: 

> FileUpload.0bf28554.js:5 TypeError: Cannot read properties of null (reading 'type')
    at I (FileUpload.0bf28554.js:5:11614)
    at FileUpload.0bf28554.js:5:6469
    at FileUpload.0bf28554.js:5:6898

### How to Reproduce?

* Unzip the file for a JSON file: [test-import-partial.zip](https://github.com/keycloak/keycloak-ui/files/9555935/test-import-partial.zip)
* Go to "create a new realm" (http://localhost:8081/auth/admin/master/console/#/master/add-realm)
* Choose "Browse..." in the UI
* Pick the JSON file and confirm with "Select"

ahus1:
@jonkoops - I see that you've recently worked on FileUpload.tsx, and this might be related. Please let me know if you need more information.

jonkoops:
Yeah, seems like the changes might be related to that. We'll do some investigation.

edewit:
@ahus1 how did you test this? where is `keycloak-admin-ui-999-20220913.041401-171.jar` coming from?

ahus1:
@edewit - I ran this from the main branch of Keycloak, and that one uses the a snapshot of the UI via a Maven repository. 

The file name above is the name of the snapshot build.

edewit:
I can't reproduce this

ikhomyn:
I just tried to reproduce it using the provided steps and got same typeError result
![Screenshot from 2022-09-26 11-23-31](https://user-images.githubusercontent.com/89014675/192241481-07ee9e0f-0bae-4908-8573-12ccc8c9dc52.png)

ikhomyn:
[...]

PR #3398, opened 2022-09-26
added null check for chrome

edewit:
fixes: #3316

## Checklist:

### DIFF
--- a/apps/admin-ui/src/components/json-file-upload/patternfly/FileUpload.tsx
+++ b/apps/admin-ui/src/components/json-file-upload/patternfly/FileUpload.tsx
@@ -117,7 +117,8 @@ export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
   const onDropAccepted = (acceptedFiles: File[], event: DropEvent) => {
     if (acceptedFiles.length > 0) {
       const fileHandle = acceptedFiles[0];
-      if (event.type === "drop") {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      if (event && event.type === "drop") {
         onFileInputChange?.(event, fileHandle);
       }
       if (type === fileReaderType.text || type === fileReaderType.dataURL) {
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/components/json-file-upload/patternfly/FileUpload.tsx
```diff
@@ -117,7 +117,8 @@ export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
   const onDropAccepted = (acceptedFiles: File[], event: DropEvent) => {
     if (acceptedFiles.length > 0) {
       const fileHandle = acceptedFiles[0];
-      if (event.type === "drop") {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      if (event && event.type === "drop") {
```
jonkoops, line 121: Perhaps we can use optional-chaining here?
```suggestion
      if (event?.type === "drop") {
```

apps/admin-ui/src/components/json-file-upload/patternfly/FileUpload.tsx
```diff
@@ -117,7 +117,8 @@ export const FileUpload: React.FunctionComponent<FileUploadProps> = ({
   const onDropAccepted = (acceptedFiles: File[], event: DropEvent) => {
     if (acceptedFiles.length > 0) {
       const fileHandle = acceptedFiles[0];
-      if (event.type === "drop") {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
jonkoops, line 120: We might want to add a comment telling future readers of this code that this only breaks in Chrome. Even a link to the issue would be nice.

jonkoops (approved):

apps/admin-ui/src/components/json-file-upload/patternfly/FileUpload.tsx
edewit: it's not our code it's copied from patternfly we don't need to fix it it's temporary here until the issues have been fixed there.

jonkoops: Ahhh, yeah. Missed that.

Issue #3416, opened 2022-09-28
Include sub-groups users not always visible


ahweis:
### Describe the bug

The feature of showing sub groups members under the "Members" tab is really nice, but it requires minimum one user to be member of the group to display sub-group members.

### Version

19.0.2

### Expected behavior

The option to see all sub-group members should always be possible.

### Actual behavior

It is only possible to see sub group members when the parent group has minimum one user assigned.

### How to Reproduce?

1. Create a group with a sub group
2. Assign user 1 to the sub group
3. Navigate to the parent groups "Members"
4. See that it only shows "No users found"
5. Add user 2 to the parent group
6. Now the option to show sub-group members are visible

edewit:
I'm not quite sure what the problem is, do you mean that you want to see the user in the parent group members of tab because he is a member of the subgroup?

ahweis:
On any group i would like to be able to see members of sub groups, at any time.

Right now i can only see members of subgroups, when the group i am on have a member, as the check box is first visible when a user is assigned the group.

edewit:
I'm not sure if we can do it like that, making it hierarchical querying all child groups members

ahweis:
The feature is already present in the UI it is just not usable as it is now. Let me try to illustrate.

I have created two groups, first group "parent-group" with a sub group "child-group".

I then assign "user1@example.org" to my "child-group".

Then i navigate to the "parent-group" and look at the "Members" tab, and is shown a page saying "no users found".

If i then add "user2@example.org" as a member of "parent-group" i then get the checkbox "Include sub-groups user" in my "Members" view, and if i check it i can see members of sub-groups.

This checkbox would be very helpful to have regardless if my "parent-group" has any members.

edewit:
Okay, sorry now I get it, sorry for being so slow, yeah for sure we can fix that

PR #3445, opened 2022-09-30
add option to include subgroup users to empty

edewit:
fixes: #3416

image


jonkoops (approved):

jonkoops:
We should probably add some tests for this, but that can be a separate issue. Merging.

PR #3471, opened 2022-10-03
Use Keycloak config endpoint in production

jonkoops:
Changes the config for Keycloak JS in when running in production so it uses [this endpoint](https://github.com/keycloak/keycloak/blob/fb24c86a3bf4dcfd661e11288686afe8a88f10ef/services/src/main/java/org/keycloak/services/resources/admin/AdminConsole.java#L180-L189) instead. This is the [same behavior](https://github.com/keycloak/keycloak/blob/fb24c86a3bf4dcfd661e11288686afe8a88f10ef/themes/src/main/resources/theme/base/admin/resources/js/app.js#L14) as the old Admin Console.

This might address issues where the Admin UI uses an incorrect URL to make requests when Keycloak is running behind a reverse proxy that splits the OpenID Connect endpoints from the rest of Keycloak.

### DIFF
--- a/apps/admin-ui/src/context/auth/AdminClient.tsx
+++ b/apps/admin-ui/src/context/auth/AdminClient.tsx
@@ -71,14 +71,21 @@ export function useFetch<T>(
   }, deps);
 }
 
-export async function initAdminClient() {
-  const keycloak = new Keycloak({
+function getKeycloakConfig() {
+  if (environment.isRunningAsTheme) {
+    return environment.consoleBaseUrl + "config";
+  }
+
+  return {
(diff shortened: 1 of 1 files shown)


ssilvert (approved):
This makes it work just like the old UI.  Hopefully, it will fix it for good. LGTM.

jonkoops:
The test failure is just a flaky test, I'm going to go ahead and merge this.

stianst:
I'd probably revert this as it doesn't make any difference (or at least shouldn't) and introduces one more request to the server. Basically what is templated into the JSON config file is already templated into the index.ftl file.

PR #3548, opened 2022-10-11
Remove optional param from new permission route

jonkoops:
Removes an optional parameter from the new permission route. Optional parameters are not supported by React Router v6, so this is causing errors at runtime.


edewit (approved):

agagancarczyk (commented):
Looks good!

PR #3619, opened 2022-10-21
Improve flow for testing e-mail server settings

### LINKED ISSUE #3593: Username and email address changed after testing SMTP settings for another realm with "Email as username" enabled
### Describe the bug

When logged in with a user of the "Master" realm and testing SMTP settings for an other realm, both username and email of the logged in user are changed when "Email as username" is enabled for the other realm.

### Version

19.0.3

### Expected behavior

Username and email address of the logged in user are not changed.
The test email is sent.

### Actual behavior

Username and email address of the logged in user are both set to the email address used to test the SMTP settings.
No email are sent even though I clicked the "Test connection" button in the popup.

### How to Reproduce?

1. Start Keycloak 
  ```
  docker run -p 8080:8080 -e KEYCLOAK_ADMIN=admin -e KEYCLOAK_ADMIN_PASSWORD=admin quay.io/keycloak/keycloak:19.0.3 start-dev
  ```
2. log in the Keycloak admin console with the `admin` user
3. create a new realm
4. switch to this new realm
5. go to "Configure > Realm settings > Login" and enable "Email as username"
6. go to "Configure > Realm settings > Email" and configure the `From`, `Host` and `Port` fields
7. click "Save"
8. a popup appears and ask to provide an email address to test the SMTP settings
9. enter any email address and click "Test connection"
10. switch back to the "Master" realm
11. go to "Manage > Users"
12. both username and email address of the `admin` user are changed, both new values are the email address provided in the popup

### Anything else?

_No response_

jonkoops:
This PR aims to improve the flow for testing the connection of the e-mail server configured in the realm settings. When testing these settings it is required that the currently signed-in user has an e-mail address configured, so that a test e-mail can be sent.

Previously when either saving these settings or testing the connection the user would be prompted with a modal dialog (see below) to confirm their e-mail address. The issue with this method is that it implicitly configures the entered e-mail address as the address of the currently logged-in user, leading to confusion (see #3593).

A screenshot showing the dialog asking the user to enter their e-mail address.

Besides poorly conveying that this implicitly configures the e-mail address of the user (instead of just sending an e-mail to the entered address), it also uses the same message when saving the settings. This might lead to further confusion as the user now gets presented with texts related to testing a connection.

To resolve this issue I have opted to remove the aforementioned dialog entirely, from both the 'Save' and 'Test connection' flows. In fact, it is not even required to configure an e-mail address when simply saving these settings (the old console does this as well).

When the user attempts to test the connection without having an e-mail address configured they will find the 'Test connection' button disabled. To avoid confusion an alert has been added below the button, notifying the user their e-mail address is yet to be configured. A link to their account settings is also presented to allow configuration of said e-mail address.

A screenshot of the alert presented to the user, asking them to configure their e-mail address.

Once the user has configured their e-mail address and returns to test the connection they will find that the 'Test connection' button is now enabled. Another alert is presented underneath the button to notify the user that an e-mail will be sent to their configured address.

A screenshot showing the enabled 'test connection' button and accompanying alert.

The UX that is being presented here is something that I came up with, so it will have to be vetted by @xianli123 before merging.

Closes #3593


jonkoops:
@lostunicorn I'd also like to know if this resolves the issue as you have reported it in #3614. Could you also take a look at this solution?

apps/admin-ui/public/resources/en/realm-settings.json
```diff
@@ -114,17 +114,20 @@
...
   "verifyEmail": "Verify email",
   "verifyEmailHelpText": "Require user to verify their email address after initial login or after address changes are submitted.",
   "userInfoSettings": "User info settings",
   "editUsernameAllowed": "Edit username",
   "enableSwitchSuccess": "{{switch}} changed successfully",
   "enableSwitchError": "Could not enable / disable due to {{error}}",
   "testConnection": "Test connection",
+  "testConnectionHint": {
+    "withEmail": "When testing the connection an e-mail will be sent to the currently logged-in user ({{email}}).",
+    "withoutEmail": "To test the connection you must first configure an e-mail address for your user.",
```
ssilvert, line 126: @jonkoops Let's make the `withoutEmail` as clear as `withEmail`.  Suggested change:
```suggestion
    "withoutEmail": "To test the connection you must configure an e-mail address for the current user ({{userName}}).",
```

ssilvert (changes requested):
I like this solution a lot.  It's a big improvement on the old console too.

lme-atolcd:
@jonkoops IMHO this PR totally makes sens to me as an end-user. The message alerting that an e-mail address must be configured for the current user is much more clear than the previous modal dialog. This will also prevent the username from being changed without the current user noticing it.

apps/admin-ui/public/resources/en/realm-settings.json
jonkoops: Done, I've also changed `withEmail` to be more consistent with the flow of `withoutEmail`.

jonkoops:
Thanks a lot for the quick feedback @lme-atolcd :+1:

ssilvert (approved):
Looks good.

lostunicorn:
> @lostunicorn I'd also like to know if this resolves the issue as you have reported it in #3614. Could you also take a look at this solution?

yes, this would solve my issue 🙂
As the email address of the user is no longer stored during the connection test, the username would not be changed, regardless of the registrationEmailAsUsername setting in the realm for witch the smtp configuration is tested.

edewit (approved):

jonkoops:
Incorporated some feedback from @xianli123 to give the alerts a different style.

xianli123 (approved):

jonkoops:
Thanks @lostunicorn, it helps a lot knowing this works for you as well.

jonkoops:
I've merged this work in, it should come along with the release of Keycloak 20 (which is expected this week).

PR #3633, opened 2022-10-24
Add labels for supported applications to 'OTP Policy'

jonkoops:
Adds some labels for the supported applications of the 'OTP Policy' tab. These changes were added upstream under https://github.com/keycloak/keycloak/pull/14800. This should allow the test suite to pass again.

Also changes the text input to a chip group, which can show this readonly listed value a bit better:

image


edewit (approved):
Nice much better then the Text field

PR #3641, opened 2022-10-25
when using timeout we don't need onTimeout

edewit:
this fixes the issue with the notification timeout being cumulative
see: https://github.com/patternfly/patternfly-react/issues/8064

## Checklist:

### DIFF
--- a/apps/admin-ui/src/components/alert/AlertPanel.tsx
+++ b/apps/admin-ui/src/components/alert/AlertPanel.tsx
@@ -28,7 +28,6 @@ export function AlertPanel({ alerts, onCloseAlert }: AlertPanelProps) {
             />
           }
           timeout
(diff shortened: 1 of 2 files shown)


ssilvert (approved):

jonkoops:
Oh damn, nice find Erik!

jonkoops (changes requested):
Since we are not passing this callback anymore, how is the state in the `AlertProvider` updated? Seems like it could be a memory leak since it will continue to hold all previous alerts in the state.

edewit:
Yes @jonkoops you make a good point. How about I change it so that "expired" alerts get removed at some interval?

jonkoops:
Yeah, that seems like a good solution to me.

jonkoops (approved):

PR #3702, opened 2022-11-03
escape internal output

### LINKED ISSUE #3701: [Urgent] weird charachters when saving Authenticators config
### Area

admin/api

### Describe the bug

when I try to save any authenticator config I see glass emojis 🍺 instead of dots in the database record. Here is what I am seeing now when running: select * from authenticator_config_entry;

keycloak_config

This can be seen also from the http request payload when clicking on save:

```
{alias: "review profile config", config: {update🍺profile🍺on🍺first🍺login: "off"}}
alias
: 
"review profile config"
config
: 
{update🍺profile🍺on🍺first🍺login: "off"}
```

This also breaks the logic inside authenticators because it prevents sometimes from doing the right comparison (update🍺profile🍺on🍺first🍺login is obviously not equal to update.profile.on.first.login), for example here:  
https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/broker/IdpReviewProfileAuthenticator.java#L88

### Version

20.0.0

### Expected behavior

when saving for example any authentication flow step config, the values keys should be separated by dots not 🍺

### Actual behavior

the values keys are separated by 🍺

### How to Reproduce?

Steps for reproducing: in the admin -> going to Authentication menu -> selecting some flow (e.g first broker login flow) -> clicking on the settings icon on some step that allows configuration -> changing the configuation and clicking on save -> examining the HTTP request and the authenticator_config_entry table

### Anything else?

_No response_

edewit:
Closes #3701

### DIFF
--- a/apps/admin-ui/src/authentication/components/ExecutionConfigModal.tsx
+++ b/apps/admin-ui/src/authentication/components/ExecutionConfigModal.tsx
@@ -23,7 +23,7 @@ import { useAlerts } from "../../components/alert/Alerts";
 import { HelpItem } from "../../components/help-enabler/HelpItem";
 import { DynamicComponents } from "../../components/dynamic/DynamicComponents";
 import { KeycloakTextInput } from "../../components/keycloak-text-input/KeycloakTextInput";
-import { convertToFormValues } from "../../util";
+import { convertFormValuesToObject, convertToFormValues } from "../../util";
 
 type ExecutionConfigModalForm = {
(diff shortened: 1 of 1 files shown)


jonkoops (approved):

jonkoops:
Cheers :beers:

jonkoops:
Test failure seems to be from an upstream change. Going to merge this.

PR #3771, opened 2022-11-15
Use `react-form-hook` v7 for user IDP modal

jonkoops:
Refactors the user IDP modal to use v7 of `react-form-hook`, and cleans up the code a little.


edewit (commented):
I think you still need an `id` on the input fields, because then the you bind the `FormGroup` to the field

jonkoops:
> I think you still need an `id` on the input fields, because then the you bind the `FormGroup` to the field

Just pushed a fix for that :+1:

edewit (approved):
:tada:

PR #3785, opened 2022-11-17
Use `react-form-hook` v7 for flow form

jonkoops:
Refactors the flow form to use v7 of `react-form-hook`, and cleans up the code a little.

### DIFF
--- a/apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
+++ b/apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
@@ -1,7 +1,4 @@
-import { useEffect } from "react";
-import { useTranslation } from "react-i18next";
-import { FormProvider, useForm } from "react-hook-form";
-import { useNavigate } from "react-router-dom-v5-compat";
+import AuthenticationFlowRepresentation from "@keycloak/keycloak-admin-client/lib/defs/authenticationFlowRepresentation";
 import {
   AlertVariant,
   Button,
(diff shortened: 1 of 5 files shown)


apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
```diff
@@ -31,27 +32,25 @@ export const DuplicateFlowModal = ({
...
 
   useEffect(() => {
-    setValue("description", description);
     setValue("alias", t("copyOf", { name }));
-  }, [name, description, setValue]);
+    setValue("description", description);
+  }, [name, description]);
 
-  const save = async () => {
+  const onSubmit = async () => {
```
edewit, line 47: forgot to point this out in the previous PR,  but why change this to `onSubmit` that is not what is happening here, save is a better name. As `handleSubmit` from react hook form is handling the on submit

apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
```diff
@@ -82,33 +81,34 @@ export const DuplicateFlowModal = ({
...
       onClose={toggleDialog}
       variant={ModalVariant.small}
       actions={[
         <Button
-          id="modal-confirm"
           key="confirm"
-          onClick={save}
           data-testid="confirm"
+          type="submit"
+          form="duplicate-flow-form"
```
edewit, line 91: :+1:

apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
jonkoops: The reason why I chose this approach is that it's inline with the examples from the React Form Hook documentation, which I try to follow as closely as possible.

We're also using different names for these callbacks in an inconsistent manner throughout the code-base, so this is an effort to make those names consistent throughout.

apps/admin-ui/src/authentication/DuplicateFlowModal.tsx
jonkoops: Yeah, this is much better for accessibility. Found this gem in some of Jenny's code, now I am trying to apply it everywhere.

edewit (approved):

agagancarczyk (commented):
LGTM

PR #3786, opened 2022-11-17
Ignore requests failing with a 404 status code for localization

jonkoops:
Adds a custom loader for i18next to ignore any requests for localization files that return with a 404 status code. For some languages not all of the JSON files have been created, so this change ensures that those languages remain at least partially loaded, instead of falling back to the default language.

Works towards resolving #3431.


edewit (changes requested):
I think it would be better to remove the config that isn't there instead of making this more robust

PR #3788, opened 2022-11-17
Move translations from `attributes-groups.json` to `realm-settings.json`

jonkoops:
Moves translations from `attributes-groups.json` to `realm-settings.json`. These translations should belong to the section they are used for.

Also fixes 404 errors when switching between languages, as this JSON file is not present for all languages.


apps/admin-ui/src/realm-settings/user-profile/AttributesGroupTab.tsx
```diff
@@ -33,9 +33,9 @@ export const AttributesGroupTab = () => {
   }
 
   const [toggleDeleteDialog, DeleteConfirm] = useConfirmDialog({
-    titleKey: "attributes-group:deleteDialogTitle",
+    titleKey: "realm-settings:deleteDialogTitle",
```
edewit, line 36: these didn't need prefixes as it's already the default in `useTranslation`

edewit (approved):
:+1:

PR #3789, opened 2022-11-17
Use `react-hook-form` v7 for edit flow modal

jonkoops:
Refactors the flow form to use v7 of `react-hook-form`, and cleans up the code a little.

### DIFF
--- a/apps/admin-ui/src/authentication/components/EditFlow.tsx
+++ b/apps/admin-ui/src/authentication/components/EditFlow.tsx
@@ -1,6 +1,3 @@
-import { useEffect } from "react";
-import { useTranslation } from "react-i18next";
-import { useForm } from "react-hook-form";
 import {
   Button,
   ButtonVariant,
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/authentication/components/EditFlow.tsx
```diff
@@ -1,6 +1,3 @@
-import { useEffect } from "react";
-import { useTranslation } from "react-i18next";
-import { useForm } from "react-hook-form";
 import {
```
edewit, line 1: I always order them like this react, i18n hook, patternfly emtpty line then our own. but this is fine as well

edewit (approved):

agagancarczyk (approved):
LGTM!

apps/admin-ui/src/authentication/components/EditFlow.tsx
jonkoops: Yeah, I just use the `Organize imports` functionality of VSCode and it sorts it like this. We should really have a linting config to standardize this.

PR #3804, opened 2022-11-18
Upload Keycloak server logs when running Cypress

jonkoops:
Upload the Keycloak server logs when running the Cypress tests. This will allow us to diagnose issues when API calls fail unexpectedly.


edewit (approved):
nice, let's give that a try

PR #3816, opened 2022-11-21
Use `string` type for dates for `findEvents` method

DenisBessa:
t looks like the types here are incorrect. The Date object is invalid to query Keycloak api and throws an error: Invalid value for 'Date(From)', expected format is yyyy-MM-dd" when used.


jonkoops (approved):

jonkoops:
Thanks for your contribution @DenisBessa!

PR #3818, opened 2022-11-22
Use cryptoInfo to populate key type dropdowns.

### LINKED ISSUE #3616: Make sure that when uploading or generating keystore, the supported keystore types are dynamically retrieved
After https://github.com/keycloak/keycloak/issues/14964 change was applied in keycloak main branch, the Keycloak server is able to notify which keystore types it supports. This  is available via `ServerInfoEndpoint` and in the retrieved JSON response, it is  inside `cryptoInfo` attribute and it's `supportedKeystoreTypes` sub-attribute, which contains array of strings with supported keystore format (By default it is 3 options: JKS, PKCS12, BCFKS). This can be consumed by those places:
- OIDC client with client authentication enabled -> tab `Keys` -> Button `Generate new keys` . Now the option `Archive format` should use the types dynamically retrieved as described above instead of have 2 hardcoded options `JKS` and `PKCS12`
- The same like above, but button `Import` used in last step instead of `Generate new keys` . Now the option `Archive format` should use the types dynamically retrieved, but the options `Certificate PEM`, `Public key PEM` and `JSON Web Key Set` should be always added on top of those dynamically obtained options
- SAML client -> tab `Keys` -> button `Import keys` . Now the option `Archive format` should be again dynamically retrieved

By default, the supported keystore types include `JKS`, `PKCS12` and `BCFKS`. So effectively, all the cases above should effectively contain `BCFKS` in addition to already existing options.

One additional thing: In case of `Generate new keys` or `Export`, it will be nice if the exported keystore for BCFKS is the file like `keystore.bcfks` by default. It seems that this also needs to be handled in the admin console.

douglaspalmer:
closes #3616

### DIFF
--- a/apps/admin-ui/src/clients/keys/ExportSamlKeyDialog.tsx
+++ b/apps/admin-ui/src/clients/keys/ExportSamlKeyDialog.tsx
@@ -4,7 +4,7 @@ import { Button, Modal, Form } from "@patternfly/react-core";
 import { saveAs } from "file-saver";
 
 import KeyStoreConfig from "@keycloak/keycloak-admin-client/lib/defs/keystoreConfig";
-import { KeyForm } from "./GenerateKeyDialog";
+import { KeyForm, getFileExtension } from "./GenerateKeyDialog";
 import { useRealm } from "../../context/realm-context/RealmContext";
 import { useAdminClient } from "../../context/auth/AdminClient";
 import { useAlerts } from "../../components/alert/Alerts";
@@ -39,7 +39,7 @@ export const ExportSamlKeyDialog = ({
       );
       saveAs(
         new Blob([keyStore], { type: "application/octet-stream" }),
-        `keystore.${config.format == "PKCS12" ? "p12" : "jks"}`
+        `keystore.${getFileExtension(config.format ?? "")}`
       );
       addAlert(t("samlKeysExportSuccess"));
       close();
--- a/apps/admin-ui/src/clients/keys/ImportKeyDialog.tsx
+++ b/apps/admin-ui/src/clients/keys/ImportKeyDialog.tsx
@@ -17,19 +17,12 @@ import {
 import { HelpItem } from "../../components/help-enabler/HelpItem";
 import { StoreSettings } from "./StoreSettings";
 import { FileUpload } from "../../components/json-file-upload/patternfly/FileUpload";
+import { useServerInfo } from "../../context/server-info/ServerInfoProvider";
 
 type ImportKeyDialogProps = {
   toggleDialog: () => void;
   save: (importFile: ImportFile) => void;
 };
-
-const baseFormats = ["JKS", "PKCS12"];
-
-const formats = baseFormats.concat([
-  "Certificate PEM",
-  "Public Key PEM",
-  "JSON Web Key Set",
-]);
 
 export type ImportFile = {
   keystoreFormat: string;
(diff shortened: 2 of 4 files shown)


mposolda:
@douglaspalmer The change looks good to me, Thanks!

@ssilvert Are you please also able to review or delegate to someone from UI team to review?

apps/admin-ui/src/clients/keys/ImportKeyDialog.tsx
```diff
@@ -53,6 +46,15 @@ export const ImportKeyDialog = ({
     name: "keystoreFormat",
     defaultValue: "JKS",
   });
+
+  const baseFormats = useServerInfo().cryptoInfo?.supportedKeystoreTypes;
```
jonkoops, line 50: Providing a default value here will remove the need to use optional chaining later on.
```suggestion
  const baseFormats = useServerInfo().cryptoInfo?.supportedKeystoreTypes ?? [];
```

apps/admin-ui/src/clients/keys/Keys.tsx
```diff
@@ -39,6 +39,17 @@ type KeysProps = {
 
 const attr = "jwt.credential";
 
+export const getFileExt = (keystoreType: string | undefined) => {
```
jonkoops, line 42: This could be re-used, as mentioned before.

apps/admin-ui/src/clients/keys/Keys.tsx
```diff
@@ -80,7 +91,7 @@ export const Keys = ({ clientId, save, hasConfigureAccess }: KeysProps) => {
       );
       saveAs(
         new Blob([keyStore], { type: "application/octet-stream" }),
```
jonkoops, line 93: Just wondering, is this the correct mime type for all of the extensions?

jonkoops (changes requested):

apps/admin-ui/src/clients/keys/ExportSamlKeyDialog.tsx
```diff
@@ -14,6 +14,17 @@ type ExportSamlKeyDialogProps = {
...
+export const getFileExt = (keystoreType: string | undefined) => {
+  if (keystoreType == "PKCS12") {
+    return "p12";
+  } else if (keystoreType == "JKS") {
+    return "jks";
+  } else if (keystoreType == "BCFKS") {
+    return "bcfks";
+  }
+  return "";
+};
```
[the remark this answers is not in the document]
douglaspalmer: @jonkoops any suggestion as to where the shared code should live?

apps/admin-ui/src/clients/keys/Keys.tsx
douglaspalmer: @jonkoops there appears to be a mime type application/x-pkcs12 but I don't know how much it's used.

edewit: this will at least make your browser download it, I think we use this already in other places

douglaspalmer: @edewit this isn't part of the pull request; it's the original code.

apps/admin-ui/src/clients/keys/ExportSamlKeyDialog.tsx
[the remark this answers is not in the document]
jonkoops: Yeah, that seems fine to me 👍

jonkoops (approved):
LGTM. We can handle the mime type as a separate issue.

PR #3823, opened 2022-11-23
Make form attribute conversion type-safe

jonkoops:
Makes the conversion of attribute names (beers) type-safe.


edewit (approved):
it's fun that you can make a type out of this, but in the end it's still a string so don't really get how this is useful

jonkoops:
@edewit It will actually infer that type because the generic enforces it, see:
image

PR #3869, opened 2022-11-25
Make `convertAttributeNameToForm` type-safe

jonkoops:
Make `convertAttributeNameToForm` type-safe, so we can improve type checking on forms in the future.


agagancarczyk (approved):
LGTM!

PR #3887, opened 2022-11-28
Use `react-hook-form` v7 for add host dialog form

jonkoops:
Refactors the add host dialog form to use v7 of `react-hook-form`, and cleans up the code a little.


edewit (approved):
nice

PR #3893, opened 2022-11-29
Only cancel CI jobs on pushes for PRs

jonkoops:
Sets up the concurrency for CI so that jobs on branches like `main` will always run to completion (not getting cancelled by new merges). This change will still cancel previous jobs when pushing new commits on PRs, cutting down our GitHub Actions & Cypress Dashboard runtime.


edewit (approved):

agagancarczyk (commented):
LGTHM!

PR #3897, opened 2022-11-30
Add worflow name to concurrency group

jonkoops:
Adds the name of the workflows (`main`, and `cypress`) to the concurrency group. This should prevent these jobs from cancelling each other out.


edewit (approved):
okay, nice

PR #3926, opened 2022-12-02
Added label in Client policies > global profile details heading

### LINKED ISSUE #1897: [UX Review] Client policies > global profile details: label color is not correct
### Describe the bug

The label should be in the same color as the outside list.

### Version

15.0.

### Expected behavior

image

### Actual behavior

image

### How to Reproduce?

_No response_

### Anything else?

_No response_

dominikkawka:
## Motivation
Close #1897 
## Brief Description
Changed the badge into a label
image

## Verification Steps
Realm settings -> Client policies -> fapi-1-baseline client profile details

### DIFF
--- a/apps/admin-ui/src/realm-settings/ClientProfileForm.tsx
+++ b/apps/admin-ui/src/realm-settings/ClientProfileForm.tsx
@@ -21,6 +21,7 @@ import {
   Text,
   TextVariants,
   ValidatedOptions,
+  Label,
 } from "@patternfly/react-core";
 
 import type ClientProfileRepresentation from "@keycloak/keycloak-admin-client/lib/defs/clientProfileRepresentation";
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/realm-settings/ClientProfileForm.tsx
```diff
@@ -196,7 +197,9 @@ export default function ClientProfileForm() {
         badges={[
           {
             id: "global-client-profile-badge",
-            text: isGlobalProfile ? t("global") : "",
+            text: (
+              <Label color="blue">{isGlobalProfile ? t("global") : ""}</Label>
```
jonkoops, line 201: This will show the `<Label>` element even if there is no content, perhaps the conditional can be around the element itself?

jonkoops (approved):

PR #3946, opened 2022-12-06
Add wait to 'enable user events' test

jonkoops:
Adds an arbitrary wait to the 'enable user events'  test. It looks like a re-render is causing the element reference to be lost, so hopefully this will allow it to settle.

Still need a better fix, as awaits are bad practice. For now this (hopefully) allows us to merge some PRs.


edewit (approved):
crap, but if if fixes the test let's put it in, we'll fix it afterwards

PR #3956, opened 2022-12-06
Upgrade `useParams()`  to React Router v6

jonkoops:
Replaces calls to `useParams()` with the compatibility library version for React Router v6.

### DIFF
--- a/apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
+++ b/apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
@@ -1,8 +1,5 @@
-import { useState } from "react";
-import { useParams, useRouteMatch } from "react-router-dom";
-import { Link, useNavigate } from "react-router-dom-v5-compat";
-import { useTranslation } from "react-i18next";
-import { FormProvider, useForm } from "react-hook-form";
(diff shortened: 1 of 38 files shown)


apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
```diff
@@ -37,7 +37,7 @@ export default function MappingDetails() {
   const { adminClient } = useAdminClient();
   const { addAlert, addError } = useAlerts();
 
-  const { id, mapperId, type } = useParams<MapperParams>();
+  const { id, mapperId, type } = useParams<MapperParams>() as MapperParams;
```
edewit, line 40: This is a lot less nice, can't believe they did it like that

edewit (dismissed):
can't we create a mapper so that we don't need to cast everywhere?

apps/admin-ui/src/client-scopes/details/MappingDetails.tsx
jonkoops: Yeah, it sucks. I agree with your suggestion, I'll put it in a util so we don't have to cast it all the time.

PR #3986, opened 2022-12-09
JKS keystore type available as default option in FIPS mode

### LINKED ISSUE #3977: JKS keystore type available as default option in FIPS mode
In FIPS scenario, the JKS keystore type should not be available in the options and it works fine. However it is still set as the default option.

When I login to admin console and go to clients -> Some OIDC confidential client -> Tab "Keys" -> Click to "Generate key" . I see the JKS as default set option even if it should not be available. When I click on it and I change it to something else (for example `PKCS12`), then I don't see `JKS` anymore in the available options, which is OK.

So `JKS` is not available in the options, however it is set as default option even if it should not be available.

#### Expected behaviour

Instead of hardcoded to `JKS`, the 1st available option should be set as the default option (In this case usually `PKCS12`)

#### How to test

If you don;t have FIPS environment, this can be easily reproduced by temporarily disable `JKS` keystore in the `DefaultCryptoProvider`. The change from this commit should be sufficient to be able to test with `JKS` disabled: https://github.com/mposolda/keycloak/commit/e23a06b9accaf6316b486d36a93c73add189831e

douglaspalmer:
Removed default values.
Closes #3977

### DIFF
--- a/apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
+++ b/apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
@@ -30,8 +30,8 @@ export default class KeysTab extends CommonPage {
     keyPassword: string,
     storePassword: string
   ) {
-    cy.get('[id="archiveFormat"]').click();
-    cy.findAllByRole("option").contains("JKS").click();
+    cy.get("#archiveFormat").click();
+    cy.findAllByRole("option").contains(archiveFormat).click();
     cy.findByTestId(this.keyAlias).type(keyAlias);
     cy.findByTestId(this.keyPassword).type(keyPassword);
     cy.findByTestId(this.storePassword).type(storePassword);
--- a/apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
+++ b/apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
@@ -4,6 +4,7 @@ export default class KeysTab extends CommonPage {
   private generateBtn = "generate";
   private confirmBtn = "confirm";
   private useJwksUrl = "useJwksUrl";
+  private archiveFormat = "archiveFormat";
   private keyAlias = "keyAlias";
   private keyPassword = "keyPassword";
   private storePassword = "storePassword";
@@ -24,10 +25,13 @@ export default class KeysTab extends CommonPage {
   }
 
   fillGenerateModal(
+    archiveFormat: string,
     keyAlias: string,
     keyPassword: string,
     storePassword: string
   ) {
+    cy.get('[id="archiveFormat"]').click();
+    cy.findAllByRole("option").contains("JKS").click();
     cy.findByTestId(this.keyAlias).type(keyAlias);
     cy.findByTestId(this.keyPassword).type(keyPassword);
     cy.findByTestId(this.storePassword).type(storePassword);
--- a/apps/admin-ui/src/clients/keys/GenerateKeyDialog.tsx
+++ b/apps/admin-ui/src/clients/keys/GenerateKeyDialog.tsx
@@ -66,6 +66,10 @@ export const KeyForm = ({
   const format = watch("format");
 
   const { cryptoInfo } = useServerInfo();
+  const supportedKeystoreTypes = [
+    ...(cryptoInfo?.supportedKeystoreTypes ?? []),
+    ...(hasPem ? CERT_PEM : []),
+  ];
 
   return (
     <Form className="pf-u-pt-lg">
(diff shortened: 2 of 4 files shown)


mposolda (changes requested):
@douglaspalmer The change seems to be in right direction, but there is test error and it looks that it is regression of this PR?

I guess the `defaultValue` may be still set, but possibly to the 1st option in the available options. So something like `defaultValue=options[0]` (Just pseudo-code for illustration, I don't have clue what is correct syntax :-) )

At least, if it is not possible to set the default value right away, it should be probably ok to set to the 1st option right before it is saved (Button `Save` is clicked)?

I can be wrong but I guess that the test error is caused by the fact that HTTP request to server fails because `keystoreType` is not set in the REST request to server. So when the `keystoreType` is set to the 1st option, the HTTP request sent to the server should have `keystoreType` correctly set to the 1st option (if it was not changed by the admin by explicitly selecting different option).

jonkoops:
@mposolda If I read the issue correctly then the default should be `PKCS12` in a FIPS environment. However is `JKS` not still the best option in a non-FIPS environment? Or should the default always be `PKCS12` no matter the environment?

mposolda:
@jonkoops The default should be the 1st option available. In the typical (non-FIPS) environment, it is usually `JKS`. In the FIPS environment, it can be usually `PKCS12`. However even in some FIPS environment (for example with BCFIPS library set to "Approved mode"), the PKCS12 may not be available and hence default should be `BCFKS`.

Simply said, really the 1st option available from combobox.

douglaspalmer:
> @jonkoops The default should be the 1st option available. In the typical (non-FIPS) environment, it is usually `JKS`. In the FIPS environment, it can be usually `PKCS12`. However even in some FIPS environment (for example with BCFIPS library set to "Approved mode"), the PKCS12 may not be available and hence default should be `BCFKS`.
> 
> Simply said, really the 1st option available from combobox.

Isn't the first option what we get when we don't set a default? From my cursory testing that appears to be the case.

apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
```diff
@@ -24,10 +25,13 @@ export default class KeysTab extends CommonPage {
   }
 
   fillGenerateModal(
+    archiveFormat: string,
     keyAlias: string,
     keyPassword: string,
     storePassword: string
   ) {
+    cy.get("#archiveFormat").click();
```
jonkoops, line 33: Added an id selector here instead of querying using an attribute selector.

apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
```diff
@@ -24,10 +25,13 @@ export default class KeysTab extends CommonPage {
...
+    cy.findAllByRole("option").contains(archiveFormat).click();
```
jonkoops, line 34: Restored `archiveFormat` here, instead of hard-coding one of the options.

apps/admin-ui/src/clients/keys/GenerateKeyDialog.tsx
```diff
@@ -77,7 +81,7 @@ export const KeyForm = ({
       >
         <Controller
           name="format"
-          defaultValue="JKS"
+          defaultValue={supportedKeystoreTypes[0]}
```
jonkoops, line 84: Added the first option of the `supportedKeystoreTypes` here, as not providing a default will log warnings.

apps/admin-ui/src/clients/keys/ImportKeyDialog.tsx
```diff
@@ -55,6 +49,12 @@ export const ImportKeyDialog = ({
     "JSON Web Key Set",
   ]);
 
+  const format = useWatch({
+    control,
+    name: "keystoreFormat",
+    defaultValue: formats[0],
```
jonkoops, line 55: Same here, I added the first option as a default.

jonkoops (approved):

jonkoops:
@mposolda @douglaspalmer I added a commit with my review feedback and implementation based on what was discussed here. Could you take a look and let me know if it looks and behaves as you would expect?

@edewit would you mind reviewing this PR as I added some of my own code?

edewit (approved):
:+1:

apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
douglaspalmer: Do we still need to modify the test at all since you've set defaults? I think it was only failing because I removed the default.

apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/client_details/tabs/KeysTab.ts
douglaspalmer: Do we still need to modify the test at all since you've set defaults?

mposolda:
@jonkoops Thanks for your work on this!

PR #4012, opened 2022-12-12
Updates to PatternFly 4.222

thatblindgeye:
Updates to PatternFly version 4.222


thatblindgeye:
@jonkoops these are the final release candidates!

jonkoops (changes requested):
Can we update these dependencies in other places of the workspace as well? We're also using them in `apps/account-ui` and some of the projects in the `lib/` directory.

edewit (approved):

PR #4013, opened 2022-12-12
Allow add/remove user with only query-users role.

ssilvert:
## Motivation
Fixes https://github.com/keycloak/keycloak-ui/issues/3965

## Brief Description
Also needs back end change.  See https://github.com/keycloak/keycloak/pull/15967

## Verification Steps
See #3965 

Simpler way than what is in the issue is to grant permissions with a user policy instead of a group policy.

## Additional Notes
Need to wait for back end change to be merged.

### DIFF
--- a/apps/admin-ui/src/user/UserForm.tsx
+++ b/apps/admin-ui/src/user/UserForm.tsx
@@ -272,7 +272,9 @@ export const UserForm = ({
           <FederatedUserLink user={user} />
         </FormGroup>
       )}
-      {isFeatureEnabled(Feature.DeclarativeUserProfile) ? (
+      {isFeatureEnabled(Feature.DeclarativeUserProfile) &&
+      realm?.attributes?.userProfileEnabled !== undefined &&
+      realm.attributes.userProfileEnabled === "true" ? (
         <UserProfileFields />
       ) : (
         <>
--- a/apps/admin-ui/src/user/UserForm.tsx
+++ b/apps/admin-ui/src/user/UserForm.tsx
@@ -272,8 +272,7 @@ export const UserForm = ({
           <FederatedUserLink user={user} />
         </FormGroup>
       )}
-      {isFeatureEnabled(Feature.DeclarativeUserProfile) &&
-      realm?.attributes?.userProfileEnabled === "true" ? (
+      {isFeatureEnabled(Feature.DeclarativeUserProfile) ? (
         <UserProfileFields />
       ) : (
         <>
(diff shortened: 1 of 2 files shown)


apps/admin-ui/src/user/UserForm.tsx
```diff
@@ -256,8 +256,7 @@ export const UserForm = ({
           <FederatedUserLink user={user} />
         </FormGroup>
       )}
-      {isFeatureEnabled(Feature.DeclarativeUserProfile) &&
-      realm?.attributes?.userProfileEnabled === "true" ? (
+      {isFeatureEnabled(Feature.DeclarativeUserProfile) ? (
```
ssilvert, line 259: @edewit Wonder if this change is OK?  The problem I had was that the `userProfileEnabled` attribute would be undefined if the user does not have the `view-realm` role.

edewit: there have been more issues like this, I guess it's fine

edewit (approved):

ssilvert:
API change is now merged into keycloak/keycloak.  Removed hold.

apps/admin-ui/src/user/UserForm.tsx
jonkoops: Are we sure this is ok? I think that for these fields to be presented the feature must also be enabled in the realm settings, otherwise what is the purpose of that switch?

jonkoops (changes requested):
Not quite sure if we can remove the condition from the `UserForm`.

apps/admin-ui/src/user/UserForm.tsx
edewit: the problem is that we can't check the realm attributes when you don't have rights

ssilvert:
> Not quite sure if we can remove the condition from the `UserForm`.

Anyway, I had to add an extra check for `undefined` so that the screen would be rendered the old way and the tests will pass.  Now that part is equivalent to what we had before and it fixes the case where the user does not have the `view-realm` role.

jonkoops:
@ssilvert there are some merge conflicts on the PR, could you resolve them?

apps/admin-ui/src/user/UserForm.tsx
```diff
@@ -273,7 +273,8 @@ export const UserForm = ({
         </FormGroup>
       )}
       {isFeatureEnabled(Feature.DeclarativeUserProfile) &&
-      realm?.attributes?.userProfileEnabled === "true" ? (
+      realm?.attributes?.userProfileEnabled !== undefined &&
+      realm.attributes.userProfileEnabled === "true" ? (
```
jonkoops, line 277: The first statement here is redundant. `realm?.attributes?.userProfileEnabled === "true"` will already implicitly assert that `realm.attributes.userProfileEnabled` is not `undefined`.

So unless I am missing something the following code should yield the same result.
```suggestion
      realm?.attributes?.userProfileEnabled === "true" ? (
```

jonkoops (changes requested):

Issue #2873, opened 2022-06-24
Consider styled-components or css modules


edewit:
### Description

Right now we override extend patternfly styles for curtain components and initially we thought that this would enable users to change the look and feel of the admin console. Right now this is not so realistic anymore due to how the console is installed. So we should rethink this approach and maybe use [styled components](https://styled-components.com/) or css / sass modules.

edewit:
The easiest way to solve the problem thought is to switch to css modules, that way we can have / reuse the css files we have now, but just apply them differently and that will make overriding patternfly styles more easily.

jonkoops:
Note that this would only affect our styling and not the styling we're importing from PatternFly. I think we should proceed with CSS modules rather than Styled Components as we have very little custom styling and it would prevent adding another runtime dependency.

jonkoops:
@edewit @agagancarczyk @ssilvert can we commit to the above as a team?

ssilvert:
> Note that this would only affect our styling and not the styling we're importing from PatternFly. I think we should proceed with CSS modules rather than Styled Components as we have very little custom styling and it would prevent adding another runtime dependency.

I have some serious questions:
How hard would it be to get rid of our custom styling?  
Are we using custom styling in places where we could accomplish the same thing with PatternFly's CSS?
To the extent that PatternFly doesn't supply what we need, should we get the PatternFly team to address it?

I hate the idea that a customer who wants to theme our consoles requires him to deal with both our CSS and PatternFly's CSS.  If he only has to care about PatternFly styling, doesn't that make life simpler?

jonkoops:
I think we should settle on using CSS modules. I'll add this to the agenda for our next team meeting.

PR #4044, opened 2022-12-15
added tooltip and css to group name

edewit:
fixes: #3970

## Checklist:

### DIFF
--- a/apps/admin-ui/src/groups/components/GroupTree.tsx
+++ b/apps/admin-ui/src/groups/components/GroupTree.tsx
@@ -9,6 +9,7 @@ import {
   DropdownSeparator,
   InputGroup,
   KebabToggle,
+  Tooltip,
   TreeView,
   TreeViewDataItem,
 } from "@patternfly/react-core";
@@ -25,6 +26,8 @@ import { useSubGroups } from "../SubGroupsContext";
 import { fetchAdminUI } from "../../context/auth/admin-ui-endpoint";
 import { useRealm } from "../../context/realm-context/RealmContext";
 import { joinPath } from "../../utils/joinPath";
+
+import "./group-tree.css";
 
 type GroupTreeContextMenuProps = {
   group: GroupRepresentation;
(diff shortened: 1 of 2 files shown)


apps/admin-ui/src/groups/components/group-tree.css
```diff
@@ -0,0 +1,6 @@
+.keycloak_groups_treeview .pf-c-tree-view__node {
```
jonkoops, line 1: Perhaps we can use a CSS module for this one? Vite has support for it built-in for it.

apps/admin-ui/src/groups/components/GroupTree.tsx
```diff
@@ -128,13 +131,15 @@ export const GroupTree = ({ refresh: viewRefresh }: GroupTreeProps) => {
...
-          key={group.id}
-          to={`/${realm}/groups/${joinPath(...groups.map((g) => g.id!))}`}
-          onClick={() => setSubGroups(groups)}
-        >
-          {group.name}
-        </Link>
+        <Tooltip content={group.name}>
+          <Link
+            key={group.id}
+            to={`/${realm}/groups/${joinPath(...groups.map((g) => g.id!))}`}
```
jonkoops, line 137: Any chance this could be a link function?

apps/admin-ui/src/groups/components/GroupTree.tsx
```diff
@@ -128,13 +131,15 @@ export const GroupTree = ({ refresh: viewRefresh }: GroupTreeProps) => {
...
-        <Link
-          key={group.id}
-          to={`/${realm}/groups/${joinPath(...groups.map((g) => g.id!))}`}
-          onClick={() => setSubGroups(groups)}
-        >
-          {group.name}
-        </Link>
+        <Tooltip content={group.name}>
+          <Link
+            key={group.id}
```
jonkoops, line 136: Why would we need a `key` here?
```suggestion
```

jonkoops (changes requested):

apps/admin-ui/src/groups/components/group-tree.css
edewit: we don't use that anywhere yet, if we want to use that I feel that we should migrate everything at once

apps/admin-ui/src/groups/components/GroupTree.tsx
edewit: unfortunately that is not an option, as these ids can be any number

apps/admin-ui/src/groups/components/group-tree.css
edewit: one problem we should solve though, before we can switch, is how can we still override these styles

apps/admin-ui/src/groups/components/GroupTree.tsx
jonkoops: Yeah, I figured as much. But thanks for checking.

apps/admin-ui/src/groups/components/group-tree.css
jonkoops: Fair enough. We have #2873 to track this.

PR #4046, opened 2022-12-15
Add attributes tab to realm settings

### LINKED ISSUE #3931: Admin-UI: Provide a way to view and edit realm attributes
### Description

At the moment it is not possible to edit realm attributes (for example custom realm attributes that are used in extensions) in the admin ui realm settings.

In my opinion it would be nice to have a tab for attributes in the realm settings which lists all attributes of a realm and provides a way to edit/add or remove them. Similiar to user attributes e.g. 

### Discussion

_No response_

### Motivation

Provide a non API way of editing realm attributes.

### Details

_No response_

antikalk:
## Motivation
Add a possibility to edit realm attributes in realm settings. Fixes #3931.

## Brief Description
Added an additional tab to the realm settings. Unfortunately I was not able to use the key-value-convert, as realm attributes are represented differently on API level than other attributes (string: string instead of string: string[]). **This also leads to the effect that when adding an attribute with a key that is already present the value is just overwritten.** 

Any suggestions are welcome :)

## Verification Steps
1. Go to `Realm Settings >> Attributes`.
2. Create a new attribute with key 'test' and value 'value'.
3. Click save.
4. Verify that the new attribute is available.

## Checklist:

## Additional Notes
image

### DIFF
--- a/apps/admin-ui/src/components/key-value-form/key-value-convert-unique-keys.test.ts
+++ b/apps/admin-ui/src/components/key-value-form/key-value-convert-unique-keys.test.ts
@@ -1,0 +1,82 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  arrayToKeyValue,
+  keyValueToArray,
+  KeyValueType,
+} from "./key-value-convert-unique-keys";
+
+vi.mock("react");
+
+describe("Tests the convert functions for attribute input", () => {
+  it("converts empty array into form value with unique key", () => {
+    const given: KeyValueType[] = [];
+
+    //when
+    const result = keyValueToArray(given);
+
+    //then
+    expect(result).toEqual({});
+  });
+
+  it("converts array into form value with unique key", () => {
+    const given = [{ key: "theKey", value: "theValue" }];
+
+    //when
+    const result = keyValueToArray(given);
+
+    //then
+    expect(result).toEqual({ theKey: "theValue" });
+  });
+
+  it("convert only values with unique key", () => {
+    const given = [
+      { key: "theKey", value: "theValue" },
+      { key: "", value: "" },
+    ];
+
+    //when
+    const result = keyValueToArray(given);
+
+    //then
+    expect(result).toEqual({ theKey: "theValue" });
+  });
+
+  it("convert empty object to attributes with unique key", () => {
+    const given: {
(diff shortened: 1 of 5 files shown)


edewit:
I can see how this is useful for custom attributes, but implemented like this it also shows attributes that have "proper" input fields

xianli123:
Hi @stianst, could you please take a look at this issue? I am not sure about the use cases of this feature. Do we need to list all realm attributes and allow users to customize attributes? Thanks.

xgp:
@edewit Agree with your point on attributes with "proper" input fields. Do we have a definitive list somewhere of those attributes? If we could not show (or show with some kind of warning) the keys for which there are "official" attributes.

However, even as written, plus some kind of "expert mode" warning at the top, I think this would be an excellent addition, as it really helps with custom attributes, and is helpful to extension writers that use custom attributes for configuration.

xianli123:
Hi @xgp In this pr, you mentioned:
> Add a possibility to edit realm attributes in realm settings.

Could you please explain what realm attributes you want to edit or view? Thanks!

xgp:
@xianli123 I build several Keycloak extensions that use realm attributes for configuration. It would be great to have a UI for configuring those attributes.

stianst:
In it's current shape I don't think we can add this for two reasons:

1. Custom realm attributes are a very specialised thing, as such the attributes tab should maybe be added under some "advanced settings" tab or something. We may also want it to be something that has to be explicitly enabled, perhaps as a feature flag, or something a custom extension that uses custom realm attributes can enable
2. We should not expose "internal" attributes as it is doing right now. We don't really have a reliable way to know what are internal and what are custom attributes currently, as strictly speaking they are all internal, as we don't strictly support custom realm attributes, that's just something folks do. In general we prefer customisation at the provider level (for example authenticators and mappers can have their own config)

All in all I think this is a lot more effort than it is worth to do in a proper way, and I would argue that instead custom providers should be configurable by themselves.

stianst (changes requested):
I'm not in favour of adding this, especially not in its current form

xianli123:
@xgp Thanks 👍

ssilvert:
Sorry.  We can not accept this PR for the reasons stated above.

PR #4047, opened 2022-12-15
Turn protocol to required property

AmusedShadow:
## Motivation
When using the API, it is not exactly clear which values are mandatory when creating the individual entities. Mandatory does not refer to the acceptance or rejection by the REST API endpoint but to the functioning of the Admin UI. 
If a ClientScope is created without a protocol, the entire Admin UI delivered with the keycloak breaks down. Errors also appear internally on the console. 

Accordingly, the developers should be given the possible protocols on the one hand, and these values should also be validated and requested on the client side.

## Brief Description
Including a formal TyeScript definition based on th SSO Protocol types mentioned within the Keycloak documentation itself!

## Verification Steps
Create a ClientScope within Admin CLI, and validate admin console ui on your Keycloak instance. It should be up and running and not broken :) 

## Checklist:

- [x] Unit tests have been created/updated  **( Network issue with Identity provider but same issue is on the main branch!)**

## Additional Notes

Unfortunately, it is a problem to use the same interface for CREATE and UPDATE. In the case of CREATE, protocol is mandatory, but in the case of UPDATE, only really modified values should be sent. As a result, protocol must not be required as a value. Therefore, two different interfaces are needed.

### DIFF
--- a/libs/keycloak-admin-client/src/defs/ssoProtocols.ts
+++ b/libs/keycloak-admin-client/src/defs/ssoProtocols.ts
@@ -1,0 +1,21 @@
+/**
+ * Based upon the keycloak documentation could be found here
+ * https://www.keycloak.org/docs/latest/server_admin/#sso-protocols
+ * or here
+ * https://wjw465150.gitbooks.io/keycloak-documentation/content/server_admin/topics/sso-protocols.html
+ * there are actually three different SSO Protocols named and implemented
+ */
+export const SSOProtocol = {
+  "openid-connect": "openid-connect",
+  saml: "saml",
+  "docker-v2": "docker-v2",
+};
+
+/**
+ * This 'complex' reconstruction of type by keyof typeof is actually equal to
+ * export type SSOProtocol = "openid-connect" | "saml" | "docker-v2";
+ * In case SSO Protocol will be modified, you do not have to touch this type configuration by using this setup.
+ * This is inline with the TypeScript concept by replaceing an enum with an const object:
+ * Object vs. Enums - https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums
+ */
+export type SSOProtocol = typeof SSOProtocol[keyof typeof SSOProtocol];
--- a/libs/keycloak-admin-client/src/utils/constants.ts
+++ b/libs/keycloak-admin-client/src/utils/constants.ts
@@ -1,3 +1,3 @@
-export const defaultBaseUrl = "http://127.0.0.1:8180";
+export const defaultBaseUrl = "http://127.0.0.1:8080";
 
 export const defaultRealm = "master";
--- a/libs/keycloak-admin-client/test/constants.ts
+++ b/libs/keycloak-admin-client/test/constants.ts
@@ -2,7 +2,7 @@ import type { Credentials } from "../src/utils/auth.js";
 
 export const credentials: Credentials = {
   username: "admin",
-  password: "admin",
+  password: "change_me",
   grantType: "password",
   clientId: "admin-cli",
 };
--- a/libs/keycloak-admin-client/src/defs/clientRepresentation.ts
+++ b/libs/keycloak-admin-client/src/defs/clientRepresentation.ts
@@ -3,6 +3,7 @@
  */
 import type ResourceServerRepresentation from "./resourceServerRepresentation.js";
 import type ProtocolMapperRepresentation from "./protocolMapperRepresentation.js";
+import { SSOProtocol } from "./ssoProtocols.js";
 
 export default interface ClientRepresentation {
   access?: Record<string, boolean>;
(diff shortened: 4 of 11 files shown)


libs/keycloak-admin-client/src/defs/ssoProtocols.ts
```diff
@@ -0,0 +1,13 @@
...
+ * or here
+ * https://wjw465150.gitbooks.io/keycloak-documentation/content/server_admin/topics/sso-protocols.html
+ * there are actually three different SSO Protocols named and implemented
+ */
+export const SSOProtocol = {
+  "openid-connect": "openid-connect",
+  saml: "saml",
+  "docker-v2": "docker-v2",
+};
+export type SSOProtocol = typeof SSOProtocol[keyof typeof SSOProtocol];
```
edewit, line 13: this could be simpler right?
```suggestion
export type SSOProtocol = "openid-connect" | "saml" | "docker-v2";
```

edewit (commented):
Nice, it's good to have better types

libs/keycloak-admin-client/src/defs/ssoProtocols.ts
AmusedShadow: This is a possibility but, in case we are adding new values to the SSO const we need to add them manually also to the type definition ore else it would differ from each other and lead to a syntax issue. 

May we can add your suggestion as commit above and link the official typescript docu that explains the behavior of this single line. 

In my first days with this line I also thought it is a little bit over complex but we stepped in to many problems that the 'setter' const does not fit at the end to the type validator ... so we replaced the direkt string types by this and based type on configuration.

edewit: what kind of problems did you run into? we use this everywhere, didn't have any problems. And when the const changes the type needs to change as well, but that is true for your version. WDYT @jonkoops

AmusedShadow:
```
 const SSOProtocol = {
    openIDConnect: "openid-connect",
    saml:"saml",
    docker: "docker-v2" 
}
 type SSOProtocol = typeof SSOProtocol[keyof typeof SSOProtocol];

const x: SSOProtocol = "openid-connect" // Not a good idea at all. It is a REST-API value and should be represented by ENUM
console.log(SSOProtocol.openIDConnect) // Should be used becasue it could be ignored if value of openIDConnect will change


type SSOProtocolAlternative = "openid-connect" | "saml" | "docker-v2";

const y:SSOProtocolAlternative = "openid-connect" // Same so fare 
console.log(SSOProtocolAlternative.openid-connect) // ERROR! Could not found
```

On a fluid API or an API on change, it will always cause trouble in case you hard code property-values that are constructed as ENUM values within the REST API in your Client Code. 
In case you are using `SSOProtocolAlternative` your whole code need a refactor on multiple positions (see test cases, I switched from direct using "openeid-connect" to `SSOProtocol.openIDconnect`)  when the API changes.
Ok to be honest, this part of API should be relative stable, but ... 

In case you are using `SSOProtocol.openIDconnect` you just change the value of a property in 1 line, and everything will transpire correctly and because it is defined within the lib you are consuming all user of nom just need an update of package :)

AmusedShadow:
@edewit @jonkoops 
May you have a look to this Discussion: 
https://github.com/keycloak/keycloak/discussions/16086

May this should be part of PR? It is a missing property ... but I am not sure if it is just for the ui or also for the server itself ...

libs/keycloak-admin-client/src/utils/constants.ts
```diff
@@ -1,3 +1,3 @@
-export const defaultBaseUrl = "http://127.0.0.1:8180";
+export const defaultBaseUrl = "http://127.0.0.1:8080";
```
jonkoops, line 1: This should not have been changed, please revert it.
```suggestion
export const defaultBaseUrl = "http://127.0.0.1:8180";
```

jonkoops (changes requested):

libs/keycloak-admin-client/test/constants.ts
```diff
@@ -2,7 +2,7 @@ import type { Credentials } from "../src/utils/auth.js";
 
 export const credentials: Credentials = {
   username: "admin",
-  password: "admin",
+  password: "change_me",
```
jonkoops, line 5: Please revert this to what it was before.

jonkoops (changes requested):

jonkoops:
Turning this into a draft until we have reached some consensus on how to land this.

libs/keycloak-admin-client/src/utils/constants.ts
AmusedShadow: Sorry for this.

libs/keycloak-admin-client/src/defs/ssoProtocols.ts
AmusedShadow: No sorry, I will not do that. If you prefer to have it based on string literals, that's entirely your own decision within the community, which has already done a really good job! But from my point of view, besides KIS and Clean Code, the rule is everywhere: Never hardcode values. It may be common use to implement things this way, also because string enums have slipped in only in TS >= 2.4 and until today somehow do not quite fit into the overall concept. 
There are things I don't train myself to avoid after years of practice. :) You now know the problem and a solution. Then do it yourself with pleasure in your own way.

PR #4051, opened 2022-12-16
query the realm less by adding it to the context

edewit:
## Checklist:


edewit:
this has to be worked out more as it needs to update the realm when things change

Issue #4055, opened 2022-12-16
The client list can be difficult to read since client IDs can be any string


stustison:
### Description

Client IDs can be any string and in some systems are generated (like a UUID, etc.) and it makes the table difficult to read. The Clients already have a human readable attribute, "name," that appears to be unused. Adding this field to the list would make the clients easier to discern.

PR #4056, opened 2022-12-16
Added the client name to the client list view to make it easier to read the list

stustison:
fixes: #4055 

## Checklist:

### DIFF
--- a/apps/admin-ui/public/resources/ca/common.json
+++ b/apps/admin-ui/public/resources/ca/common.json
@@ -26,6 +26,7 @@
   },
   "credentials": "Credencials",
   "clientId": "ID Client",
+  "clientName": "Nom",
   "id": "ID",
   "mapperType": "Tipus d''assignador",
   "password": "Contrasenya"
--- a/apps/admin-ui/public/resources/de/common.json
+++ b/apps/admin-ui/public/resources/de/common.json
@@ -47,6 +47,7 @@
   "attributes": "Attribute",
   "credentials": "Passwörter",
   "clientId": "Client-ID",
+  "clientName": "Name",
   "leave": "Verlassen",
   "password": "Passwort",
   "passwordConfirmation": "Passwort bestätigen",
--- a/apps/admin-ui/public/resources/en/common.json
+++ b/apps/admin-ui/public/resources/en/common.json
@@ -176,6 +176,7 @@
   "valuePlaceholder": "Type a value",
   "credentials": "Credentials",
   "clientId": "Client ID",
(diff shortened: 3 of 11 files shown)


edewit (changes requested):
wouldn't it be better to use the name instead of the clientId if it's set instead of introducing a new column? WDYT @xianli123

xianli123:
Hi @edewit, We thought about putting the name out there and even did a rough design. Here is the link: https://marvelapp.com/prototype/6722i0a/screen/70973389

However, after discussion, we gave up that design because the **Name** is not required and the **Name** field is not the client's unique identifier. 

With regard to the original issue https://github.com/keycloak/keycloak-ui/issues/4055, the client ID could also be a human-readable string.  In addition, the client ID is the unique identifier of the client. We may not need to change the Client ID to Name. What do you think?

xianli123:
One thing we wanted to change is the terminology of "Client ID". In the beginning, we wanted to change the "Client ID“ to ”Name" and change the "Name" to the "Display name". But we didn't do that due to the backend limitation.

stustison:
This is most commonly an issue when configuring clients in K8s using the CRDs. You have no control over the client IDs that might be generated by another part of the system. So, yes, they "can" be human readable, but it is up to a different part of the system to generate these IDs and drop in the new resource. If the name was visible, then we could at least attach a real name to each client, which would make it easier to see in the list. Attached what our current client list looks like.
clients

xianli123:
It's weird. I used the local server to run the console, and it looks good. I am not sure why it's different with K8s. Could you please help explain it? @stianst Thanks!

Screen Shot 2022-12-20 at 11 28 13

stianst:
@xianli123 this just depends on how the client is created. A client-id is a machine identifiable id, which can be completely random (like a uuid for instance) or could use a more human friendly name like my-client.

Name on the other hand is aimed purely at humans.

I'd suggest adding both name and client-id to the table, with the name in the first column.

xianli123:
Hi @stianst Thanks! The client ID is a unique identifier for the client, whether it's human-readable or not. In addition, the **Name** is not required. Some users may not set **Name**s for clients. So I think we can add the **Name** to the list but put it after the **Client ID** as shown below.

Screen Shot 2022-12-20 at 17 10 55

If it's possible, we could even allow users to manage the columns as shown in the design here. https://marvelapp.com/prototype/6722i0a/screen/70973389. 

What do you think? Thanks.

edewit (approved):

xianli123:
Hi @edewit , did you update the client list?

PR #4078, opened 2022-12-21
add openid as default scope and add option to set custom scopes

Axenu:
## Motivation
Since the release of keycloak 20.0.0 it is required to specify the scope when authenticating. This is easily discovered when calling the /userinfo endpoint in keycloak with an auth token that is created without the scope.
https://www.keycloak.org/docs/latest/upgrading/#userinfo-endpoint-changes

## Brief Description
I have added a default scope as "openid" to solve above problem. 
I have also added an optional parameter for users who wish to set their own scope.

## Verification Steps
1. get an auth token by calling
```
const keycloakAdmin = new KeycloakAdmin({<some settings>})
keycloakAdmin.auth(authInfo)
```
2. Use the accessToken created above and call the `protocol/openid-connect/userinfo` endpoint.3. 
3. Before this PR it results in a 403.

## Checklist:

- [N/A] User-visible strings are using the react-i18next framework (useTranslation)
- [N/A] Help has been implemented
- [N/A] axe report has been run and resulting a11y issues have been resolved

### DIFF
--- a/libs/keycloak-admin-client/src/utils/auth.ts
+++ b/libs/keycloak-admin-client/src/utils/auth.ts
@@ -61,9 +61,7 @@ export const getToken = async (settings: Settings): Promise<TokenResponse> => {
     grant_type: credentials.grantType,
     client_id: credentials.clientId,
     totp: credentials.totp,
-    ...(credentials.offlineToken
-      ? { scope: "offline_access" }
-      : { scope: "openid" }),
+    ...(credentials.offlineToken ? { scope: "offline_access" } : {}),
     ...(credentials.scopes ? { scope: credentials.scopes.join(" ") } : {}),
     ...(credentials.refreshToken
       ? {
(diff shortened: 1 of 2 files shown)


jonkoops (changes requested):
Running this code against the latest nightly version of the Keycloak server gives me the following failures:

A screenshot of the failing tests

Any idea what might be causing this?

jonkoops:
Also thanks for including tests with your changes, it's very much appreciated.

Axenu:
The nightly build failing tests looks very odd. I assume this is due to some change with the new version of keycloak. I will check to see if I can find some way forward.

Axenu:
So it seems like I missed it. I have changed it so that my PR doesn't change the default scope param, but rather just ads it as optional. This works for nightly as well.

jonkoops (approved):
LGTM

jonkoops:
Tests are all passing, I am merging this in. Thanks for your contribution @Axenu!

PR #4113, opened 2023-01-02
Use React Router v6 for the routable tabs

jonkoops:
Converts the routable tabs code to use the latest version of React Router. This is one of the last pieces needed finish the migration to the new React Router.

### DIFF
--- a/apps/admin-ui/src/identity-providers/add/DetailSettings.tsx
+++ b/apps/admin-ui/src/identity-providers/add/DetailSettings.tsx
@@ -299,6 +299,20 @@ export default function DetailSettings() {
     return components;
   };
 
+  const toTab = (tab: IdentityProviderTab) =>
+    toIdentityProvider({
+      realm,
+      alias,
+      providerId,
+      tab,
+    });
+
+  const useTab = (tab: IdentityProviderTab) => useRoutableTab(toTab(tab));
+
+  const settingsTab = useTab("settings");
+  const mappersTab = useTab("mappers");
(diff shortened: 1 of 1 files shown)


jonkoops:
Converting this to a draft as it results in too many errors and should probably be split into several PRs.

apps/admin-ui/src/clients/ClientDetails.tsx
```diff
@@ -227,6 +225,55 @@ export default function ClientDetails() {
...
+  const settingsTab = useTab("settings");
+  const keysTab = useTab("keys");
+  const credentialsTab = useTab("credentials");
+  const rolesTab = useTab("roles");
+  const clientScopesTab = useTab("clientScopes");
+  const authorizationTab = useTab("authorization");
+  const serviceAccountTab = useTab("serviceAccount");
+  const sessionsTab = useTab("sessions");
+  const permissionsTab = useTab("permissions");
+  const advancedTab = useTab("advanced");
```
edewit, line 246: can't we do this like we did before? These variables are only used one time

edewit (changes requested):

apps/admin-ui/src/clients/ClientDetails.tsx
jonkoops: Because we need to call them as hooks, and the tabs render conditionally.

jonkoops:
Closing this as all the work has been merged in other PRs.

PR #4114, opened 2023-01-03
Use React Router v6 for the routable tabs of user details

jonkoops:
Converts the routable tabs of user details page to use the latest version of React Router. Also splits out the edit and create pages into separate components to reduce complexity.

Based off the work in #4113, but split out to reduce review burden.


agagancarczyk (approved):
LGTM!

PR #4118, opened 2023-01-03
Use React Router v6 for the routable tabs of authentication page

jonkoops:
Converts the routable tabs of user authentication page to use the latest version of React Router. Based off the work in #4113, but split out to reduce review burden.


agagancarczyk (approved):
LGTM!

PR #4128, opened 2023-01-05
Add `clear-data` script to Keycloak server

jonkoops:
Adds a convenient  `clear-data` script to the Keycloak server. This removes any stored information from a previously downloaded Keycloak server, allowing you to start with a clean database.


edewit (approved):
though this will not work for windows users

Issue #4116, opened 2022-12-09
New Admin Console only, unable to add client profile in the first client policy


mpugge:
### Before reporting an issue

- [X] I have searched existing issues
- [X] I have reproduced the issue with the latest release

### Area

admin/ui

### Describe the bug

This is a very strange bug, and it's happening only when using the new Admin Console.
Basically it seems to be **impossible to add a Client Profile to the very first Client Policy using the Form View** (it's possible to configure it with the JSON editor), deletion works fine.
the second and subsequent policies are not affected, only the very first policy seems to be affected. 

### Version

20.0.1

### Expected behavior

It should be possible to use the Form View to add a Client Profile to the first Client Policy.

### Actual behavior

Using Form View to add a Client Profile to the first Client Policy has no effect.

### How to Reproduce?

Use the Form View to add a Client Profile to the first Client Policy in the list. There will be no changes.
This happens regardless the number of clients already added to the policy.
There seems to be no way to use the Form view to add a Client Profile to the Client Policy currently in the first position.

ssilvert:
Moved to keycloak-ui repo.

ssilvert:
I have reproduced this.

PR #4140, opened 2023-01-06
fixed logic

edewit:
fixes: #4116

## Checklist:

### DIFF
--- a/apps/admin-ui/src/realm-settings/NewClientPolicyForm.tsx
+++ b/apps/admin-ui/src/realm-settings/NewClientPolicyForm.tsx
@@ -401,7 +401,7 @@ export default function NewClientPolicyForm() {
       (policy) => createdPolicy.name === policy.name
     );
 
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/realm-settings/NewClientPolicyForm.tsx
```diff
@@ -401,7 +401,7 @@ export default function NewClientPolicyForm() {
       (policy) => createdPolicy.name === policy.name
     );
 
-    if (!index || index === -1) {
+    if (index === undefined || index === -1) {
```
jonkoops, line 404: Perhaps this is something we could lint? Ideally we'd never use these kind of conditions on numbers, as they always could be `0`.

Sidenote, would it perhaps be better to assert the type of `number` here instead?
```suggestion
    if (typeof index !=='number' || index === -1) {
```

jonkoops (approved):

PR #4141, opened 2023-01-06
Split out role form for realm and client in seperate components

jonkoops:
Splits the form for role details into a separate shared component. This component is then re-used for the individual realm and client role forms, which are now two separate components. This allows the routes to be in their respective module, without sacrificing re-usability of the role details form itself.

This produces slightly more code, but makes it easier to follow as we now no longer try to handle 4 individual cases in a single component. Also fixes several bugs, such as:

- Incorrect ids for inputs, causing accessibility issues.
- Wrong link for the 'Cancel' button in some situations.
- Incorrect permissions for the form (`manage-realm` vs. `manage-clients`).
- Broken form submission when using the 'Enter' key.
- Remove the 'Revert' button from the form

This PR is a pre-requisite to landing the React Router upgrade (specifically #3963).


ssilvert:
@jonkoops I tried out your code and it gets a 404 when you try to create a realm role.  404 on http://localhost:8080/#/test/roles/new

That's probably why the realm role tests are failing.

jonkoops:
Looks like I forgot to update a link to the realm role page. Should be good now @ssilvert, can you take another look?

ssilvert (approved):

PR #4183, opened 2023-01-12
Refactor OpenID Connect IDP to use React Router v6

jonkoops:
Refactors the OpenID Connect identity provider code to use the new React Router v6 API.

### DIFF
--- a/apps/admin-ui/src/identity-providers/add/AddOpenIdConnect.tsx
+++ b/apps/admin-ui/src/identity-providers/add/AddOpenIdConnect.tsx
@@ -1,3 +1,4 @@
+import type IdentityProviderRepresentation from "@keycloak/keycloak-admin-client/lib/defs/identityProviderRepresentation";
 import {
   ActionGroup,
   AlertVariant,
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/identity-providers/add/AddOpenIdConnect.tsx
```diff
@@ -28,8 +27,8 @@ type DiscoveryIdentity = IdentityProviderRepresentation & {
 export default function AddOpenIdConnect() {
   const { t } = useTranslation("identity-providers");
   const navigate = useNavigate();
-  const { url } = useRouteMatch();
-  const isKeycloak = url.includes("keycloak-oidc");
+  const { pathname } = useLocation();
```
edewit, line 30: why like this wouldn't it be better if this is a parameter?

edewit (approved):

apps/admin-ui/src/identity-providers/add/AddOpenIdConnect.tsx
jonkoops: It would require us to refactor the route definitions as this is currently static, but yeah it would likely be less brittle.

PR #4187, opened 2023-01-12
Refactor roles list to React Router v6

jonkoops:
Refactors the roles list to React Router v6. Replaces dynamic route paths with router links and cleans up route related code.


agagancarczyk (approved):
LGTM!

PR #4188, opened 2023-01-12
Refactor realm keys providers to React Router v6

jonkoops:
Refactor the realm keys providers to the React Router v6 API. Also replaces dynamically built links with route functions.


agagancarczyk (approved):
LGTM!

PR #4194, opened 2023-01-13
Reduce excessive re-renders for realm & client roles tab

jonkoops:
Fixes a bug introduced by #1838 where the realm & client roles tab would create an excessive amount of requests and subsequently re-renders. This was caused by providing the current URL as a dependency to `useFetch()` which would trigger a request every time the tab was changed, even when such a fetch was not needed.

This also removes a bunch of request intercepts from the Cypress tests, which were added to stabilize the tests. Presumably this instability was actually caused by the changes made in #1838 as well.

### DIFF
--- a/apps/admin-ui/src/realm-roles/RealmRoleTabs.tsx
+++ b/apps/admin-ui/src/realm-roles/RealmRoleTabs.tsx
@@ -100,14 +100,11 @@ export default function RealmRoleTabs() {
 
       const convertedRole = convert(role);
 
+      reset(convertedRole);
       setRealm(realm);
       setRole(convertedRole);
-
-      Object.entries(convertedRole).map((entry) => {
-        setValue(entry[0], entry[1]);
-      });
     },
-    [key, url]
+    [key]
   );
 
   const onSubmit: SubmitHandler<AttributeForm> = async (formValues) => {
--- a/apps/admin-ui/cypress/e2e/clients_test.spec.ts
+++ b/apps/admin-ui/cypress/e2e/clients_test.spec.ts
@@ -535,9 +535,7 @@ describe("Clients test", () => {
     });
 
     it("Should add attribute to client role", () => {
-      cy.intercept("/admin/realms/master/roles-by-id/*").as("load");
       commonPage.tableUtils().clickRowItemLink(itemId);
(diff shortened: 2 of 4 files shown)


apps/admin-ui/src/realm-roles/RealmRoleTabs.tsx
```diff
@@ -100,14 +100,11 @@ export default function RealmRoleTabs() {
 
       const convertedRole = convert(role);
 
+      reset(convertedRole);
```
jonkoops, line 103: The `setValue()` calls have been replaced by this `reset()` call, otherwise the form will not be re-rendered with the latest value.

ssilvert:
CI number 1: All tests passed. Retry.

ssilvert:
CI number 2: All tests passed. Retry.

jonkoops:
First run of the test suite is [passing](https://cloud.cypress.io/projects/j4yhox/runs/5075/specs).

jonkoops:
Second run is [passing](https://cloud.cypress.io/projects/j4yhox/runs/5078/specs) as well.

jonkoops:
The third run [failed](https://cloud.cypress.io/projects/j4yhox/runs/5079/test-results), but on a case unrelated to these changes.

jonkoops:
The fourth run [passed](https://cloud.cypress.io/projects/j4yhox/runs/5080/specs) again.

jonkoops:
The fifth run [failed](https://cloud.cypress.io/projects/j4yhox/runs/5081/test-results?utm_source=github&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D), but on another unrelated test.

jonkoops:
Run number six [passed](https://cloud.cypress.io/projects/j4yhox/runs/5082/specs?utm_source=github) again.

jonkoops:
There have been a couple of failed runs, but the failures seem to be unrelated to the changes made here. Since this fixes some broken tests (and improves performance) I think we should go ahead and merge this in. If more tests continue to fail on `main` we should investigate them further.

@ssilvert WDYT?

edewit (approved):

PR #4226, opened 2023-01-18
more robust xml check

### LINKED ISSUE #4122: Importing XML Client Descriptor File Fails When First Line Blank
### Before reporting an issue

- [X] I have searched existing issues
- [X] I have reproduced the issue with the latest release

### Area

admin/ui

### Describe the bug

In the latest release of keycloak, using the "Import Client" feature fails when the first line of an XML file imported is blank.

This produces an error, saying:

Screenshot (1)

### Version

20.0.2

### Expected behavior

Client configurations should import successfully. If the extension of an imported client is XML, this should be enough to deduce the format of the file imported, and the file should be parsed as an XML file instead of JSON.

### Actual behavior

"Import Client" fails, presumably because Keycloak is trying to interpret an XML file as JSON.

### How to Reproduce?

You can download a config file from SamlTest.id from https://samltest.id/saml/sp . This file has a blank first line. Rename the file to <filename>.xml, attempt to import, and the import will fail because of a blank first line. When the file is edited and the first line is no longer blank, the file is parsed correctly and no errors are raised.

### Anything else?

_No response_

edewit:
fixes #4122

### DIFF
--- a/apps/admin-ui/src/clients/import/ImportForm.tsx
+++ b/apps/admin-ui/src/clients/import/ImportForm.tsx
@@ -32,7 +32,7 @@ import { FormFields } from "../ClientDetails";
 import { toClient } from "../routes/Client";
 import { toClients } from "../routes/Clients";
 
-const isXml = (text: string) => text.startsWith("<");
+const isXml = (text: string) => text.match(/(<.[^(><.)]+>)/g);
 
 export default function ImportForm() {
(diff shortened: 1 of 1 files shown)


jonkoops (approved):

edewit:
fixes: https://github.com/keycloak/keycloak-ui/issues/4122

PR #4228, opened 2023-01-18
Allow add/remove user with only query-users role.

ssilvert:
## Motivation
Fixes https://github.com/keycloak/keycloak-ui/issues/3965

## Brief Description
Relies on recent back end change. See https://github.com/keycloak/keycloak/pull/15967

## Verification Steps
See https://github.com/keycloak/keycloak-ui/issues/3965

Simpler way than what is in the issue is to grant permissions with a user policy instead of a group policy.


jonkoops (approved):

drduker:
@ssilvert - can you please please give me an example or point me to a doc to setup delegated group control?  Really struggling here thanks!

PR #4248, opened 2023-01-23
Update PatternFly to 4.224

thatblindgeye:
## Motivation

Testing pre-release numbers for PatternFly.

Additionally made updates as several beta PatternFly components were recently updated to add an `event` arg to some callback props. This PR just passes in an `_event` param where applicable.


apps/admin-ui/src/clients/authorization/policy/Time.tsx
```diff
@@ -71,15 +71,15 @@ const DateTime = ({ name }: { name: string }) => {
             <SplitItem>
               <DatePicker
                 value={dateTime[1]}
-                onChange={(_, date) => {
+                onChange={(_event, _, date) => {
```
jonkoops, line 74: Since we're not using `_event` here we can omit the rest of the name of the variable.
```suggestion
                onChange={(_, _, date) => {
```

apps/admin-ui/src/clients/authorization/policy/Time.tsx
```diff
@@ -71,15 +71,15 @@ const DateTime = ({ name }: { name: string }) => {
...
                   field.onChange(parseDate(field.value, date));
                 }}
               />
             </SplitItem>
             <SplitItem>
               <TimePicker
                 time={`${dateTime[2]}:${dateTime[3]}`}
-                onChange={(_, hour, minute) =>
+                onChange={(_event, _, hour, minute) =>
```
jonkoops, line 82: Same here
```suggestion
                onChange={(_, _, hour, minute) =>
```

jonkoops (commented):
Couple of small nits, otherwise LGTM.

apps/admin-ui/src/clients/authorization/policy/Time.tsx
thatblindgeye: Looks like that causes an error and build failure for "Duplicate identifier"

jonkoops: Ah yes of course, let's write out the param names fully then. The linter should not complain as long as a param after is being used.

jonkoops (approved):

jonkoops (approved):

jonkoops:
We need some of the work in this PR for the React 18 upgrade so I am merging this in. A PR with the stable versions can be made to follow this up.

Thanks again @thatblindgeye!

Issue #3638, opened 2022-10-09
Client registration policies not available in Keyclok 19.0.1 admin UI


souravs17031999:
### Describe the bug

Prior to keycloak 19, there was a separate tab and process for creating client registration policies, we need those policies in Keycloak 19 as well for OIDC dynamic client registrations.
I am unable to find any tab for creating those.  

Even as per Keycloak latest docs, it says there's a plan to remove these "client registration policies" and migrate to "client policies" but I can't see anywhere where it says it has been deprecated/removed already and also following policies which were available earlier for ex. "Max Clients Policy" are not available now in client policies.   

![qa](https://user-images.githubusercontent.com/33771969/194738136-19408453-d02a-45b6-8090-8b7540c8e51b.PNG)

![qa](https://user-images.githubusercontent.com/33771969/194738263-92bf0bf0-ce1c-4e70-ba16-577efc954673.PNG)

### Version

19.0.1

### Expected behavior

There should be a client registration policies tabs for managing client policies related to dynamic client registrations.
If those are not available, then existing client registration policies should be migrated and available in client policies tabs.

### Actual behavior

Client registration policies is not available in admin console.

PR #4250, opened 2023-01-24
client registration

edewit:
fixes: #3638


apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/ClientRegistrationPage.ts
```diff
@@ -0,0 +1,23 @@
...
+    return this;
+  }
+
+  createPolicy() {
+    cy.findAllByTestId("createPolicy").click();
+    return this;
+  }
+
+  selectRow(name: string) {
+    cy.get(`#${name}`).click();
```
jonkoops, line 15: Should we not be using a test id here instead?

apps/admin-ui/src/clients/routes/AddRegistrationProvider.ts
```diff
@@ -0,0 +1,36 @@
+import { lazy } from "react";
```
jonkoops, line 1: Can we split this into two separate route files?

jonkoops (changes requested):

apps/admin-ui/cypress/support/pages/admin-ui/manage/clients/ClientRegistrationPage.ts
edewit: I would if `data-testid` would be on the rendered html, but it isn't

jonkoops (approved):

PR #4268, opened 2023-01-26
Regenerate lockfile in v3 format

jonkoops:
Since the [release](https://github.com/npm/cli/releases/tag/v9.0.0) of NPM v9 the latest version of the lockfile (v3) has become the default. This change regenerates the lockfile from scratch in this new format, and in the process bumps all transitive dependencies to their latest versions.

This change is backwards compatible with version 8 of NPM.

### DIFF
--- a/apps/account-ui/package.json
+++ b/apps/account-ui/package.json
@@ -18,7 +18,7 @@
     "react-dom": "^17.0.2",
     "react-hook-form": "^7.38.0",
     "react-i18next": "^12.1.4",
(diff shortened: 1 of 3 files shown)


jonkoops:
Marking this as a draft as I want to move in the other dependency upgrades first.

agagancarczyk (approved):

jonkoops:
Had to specify an exact version of React Router as the new patch version has a bug that breaks our application. Next version should fix it so we'll deal with it then.

PR #4274, opened 2023-01-27
moved common controls to ui shared

edewit:
## Checklist:


libs/ui-shared/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "scripts": {
+    "build": "wireit",
+    "lint": "wireit"
+  },
+  "wireit": {
+    "build": {
+      "command": "vite build",
+      "dependencies": [
+        "../../libs/keycloak-js:build"
+      ],
```
jonkoops, line 25: I don't believe we have a dependency on `keycloak-js` here, let's remove this one.
```suggestion
```

libs/ui-shared/src/main.ts
```diff
@@ -0,0 +1,5 @@
+export { ContinueCancelModal } from "./continue-cancel/ContinueCancelModel";
+export { SelectControl } from "./controls/SelectControl";
+export { TextControl } from "./controls/TextControl";
+export { KeycloakTextInput } from "./keycloak-text-input/KeycloakTextInput";
+export { AlertProvider, useAlerts } from "./alerts/Alerts";
```
jonkoops, line 5: We should also export the prop types here so they can be re-used.

jonkoops (changes requested):

libs/ui-shared/src/main.ts
edewit: right

jonkoops (changes requested):
@edewit looks like there are some build errors, can you take a look?

jonkoops (approved):

PR #4287, opened 2023-01-30
added component test

edewit:
## Motivation

created component test for the multi line edit see https://github.com/keycloak/keycloak-ui/issues/4273

## Checklist:

### DIFF
--- a/apps/admin-ui/cypress/component/MultiLineInput.cy.tsx
+++ b/apps/admin-ui/cypress/component/MultiLineInput.cy.tsx
@@ -1,0 +1,155 @@
+import {
+  ActionGroup,
+  Button,
+  Form,
+  FormGroup,
+  Page,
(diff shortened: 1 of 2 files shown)


jonkoops:
Note that as far as I know we are currently not running any component tests on CI.

edewit:
@jonkoops that is something we could change in the future, but still a tests like this is useful to be sure a complex component like this works, even when it means running this by hand

jonkoops:
@edewit Yeah, fair enough. I want to avoid these tests regressing over time and then when we add a CI job it fails and will require fixing.

I'll take a look if I can add this quickly tomorrow.

jonkoops:
Implemented this under #4307, I am giving the PR a rebase so we can see this pass on CI.

jonkoops (approved):

Issue #4121, opened 2022-12-23
Groups names in search results in Admin UI (Users) cropping in middle


igor-nikiforov:
### Before reporting an issue

- [X] I have searched existing issues
- [X] I have reproduced the issue with the latest release

### Area

admin/ui

### Describe the bug

In Admin UI in User section if we need to join users to some group using "Join Group" button in "Groups" tab and trying to search some group using part of group name like this:

![image](https://user-images.githubusercontent.com/41969772/209315946-26d1d5d8-1732-41c0-93a0-233553d19dd2.png)

Group names just cropping in middle and it is impossible to read full group name. This bug was introduced with new Admin UI.

### Version

20.0.2

### Expected behavior

Group names in search result should not crop in middle.

### Actual behavior

Group names in search result cropping in middle.

### How to Reproduce?

Go to Admin UI in Users sections, select any user, go to Groups tab ("Join Group" button) and start typing any part of existing group.

edewit:
what should it do instead?

igor-nikiforov:
@edewit I'm expecting the same behavior when you just click "Join Group" button in "Groups" tab but **NOT trying** to type anything in search bar.

edewit:
wouldn't that be annoying as well though, then you'll have to click down into all the groups

edewit:
what about a tree view instead:

![image](https://user-images.githubusercontent.com/51133/214831863-6b6535f9-2bca-4fb9-9912-aa9544455e88.png)

PR #4290, opened 2023-01-30
made dialog more wide when searching and used css to cut off the end when needed

edewit:
fixes: #4121

this creates more room for the path:
image

and when there is still not enough room it will be cut off in the end with have tooltip:
image

## Checklist:


apps/admin-ui/src/components/group/group-picker-dialog.css
```diff
@@ -0,0 +1,5 @@
+.keycloak-groups-group-path {
```
jonkoops, line 1: Can we start using a CSS module for this?

jonkoops (changes requested):

jonkoops (approved):

PR #4292, opened 2023-01-30
Introduce linting rule warn about nested components

jonkoops:
Introduces the [`react/no-unstable-nested-components`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md) linting rule. This will warn when nested components are used which causes accidental component re-mounts.

I have created #4293 to fix these in the future.


edewit (approved):
seems we are doing this, and it's easy to fix

PR #4300, opened 2023-02-01
Add CIBA policy tab to authentication policies

### LINKED ISSUE #4286: "Ciba Policy" configuration doesn't seem to be implemented
### Describe the bug

In the old admin theme, there is the possibility to configure the CIBA policy, but if I switch to the new v2 theme (which is default BTW), I don't have that configuration page anymore (unless it's been redesigned and is housed somewhere else).

Old
afbeelding

New
afbeelding

### Version

20.0.2

### Environment

Firefox 108.0.1 (64-bits)

### Expected behavior

I would have expected "CIBA Policy" tab to be visible under Authentication > [tab]Policies. 

### Actual behavior

afbeelding

I've checked the code as well to see if it is mentioned, but it doesn't seem to be implemented at all actually.

### How to Reproduce?

I would expect starting a server with CIBA enabled would enable this automatically so..

1. ./kc.sh start-dev --spi-ciba-auth-channel-ciba-http-auth-channel-http-authentication-channel-uri=http://example.com/
2. Go to http://localhost:8080
3. Log in and navigate to "Authentication > [tab]Policies"
4. Observe that CIBA Policy tab is not there

### Anything else?

Altough all documentation seems to be pointing in that direction, it was only until I started digging in the code that I actually found that it was available in the old theme.

jonkoops:
Closes #4286.

Adds the "CIBA Policies" tab to authentication policies (Authentication :arrow_right: Policies :arrow_right: CIBA Policy). The form is different than the original in several ways:

- The "Backchannel Token Delivery Mode" is no longer marked as required, as there is always a value selected.
- Validation messages have been added for minimal and maximum values for the "Expires In" and "Interval" fields.
- The "Expires In" and "Interval" fields are now required (the original allowed saving empty values, which would error out when saving).
- ~~The "Authentication Requested User Hint" field has been omitted, as there is only one value that can be selected at all times.~~ This has been restored with a slightly different design.

### DIFF
--- a/apps/admin-ui/public/resources/en/authentication-help.json
+++ b/apps/admin-ui/public/resources/en/authentication-help.json
@@ -45,5 +45,6 @@
   },
   "cibaBackchannelTokenDeliveryMode": "Specifies how the CD (Consumption Device) gets the authentication result and related tokens. This mode will be used by default for the CIBA clients, which do not have other mode explicitly set.",
   "cibaExpiresIn": "The expiration time of the \"auth_req_id\" in seconds since the authentication request was received.",
-  "cibaInterval": "The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint."
+  "cibaInterval": "The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint. If set to 0, the CD must use 5 as the default value according to the CIBA specification.",
+  "cibaAuthRequestedUserHint": "The way of identifying the end-user for whom authentication is being requested. Currently only \"login_hint\" is supported."
 }
--- a/apps/admin-ui/public/resources/en/authentication-help.json
+++ b/apps/admin-ui/public/resources/en/authentication-help.json
@@ -42,5 +42,8 @@
     "digits": "The number of numerical digits required in the password string.",
     "hashAlgorithm": "Applies a hashing algorithm to passwords, so they are not stored in clear text.",
     "maxLength": "The maximum number of characters allowed in the password."
-  }
+  },
+  "cibaBackchannelTokenDeliveryMode": "Specifies how the CD (Consumption Device) gets the authentication result and related tokens. This mode will be used by default for the CIBA clients, which do not have other mode explicitly set.",
+  "cibaExpiresIn": "The expiration time of the \"auth_req_id\" in seconds since the authentication request was received.",
+  "cibaInterval": "The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint."
 }
--- a/apps/admin-ui/public/resources/en/authentication.json
+++ b/apps/admin-ui/public/resources/en/authentication.json
@@ -43,6 +43,7 @@
   },
   "cibaExpiresIn": "Expires In",
   "cibaInterval": "Interval",
+  "cibaAuthRequestedUserHint": "Authentication Requested User Hint",
   "updateCibaSuccess": "CIBA policy successfully updated",
   "updateCibaError": "Could not update CIBA policy: {{error}}",
   "webAuthnPolicySignatureAlgorithms": "Signature algorithms",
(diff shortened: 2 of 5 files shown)


edewit (approved):

tnorimat:
@jonkoops Thanks. I will review the PR. Also, I will built the theme of the PR and run keycloak with the theme. After that, I will give you feedbacks.

xianli123 (approved):

tnorimat:
@jonkoops I have only two comments about "interval" and "Authentication Requested User Hint". I will write them in detail later.

jonkoops:
> I have only two comments about "interval" and "Authentication Requested User Hint". I will write them in detail later.

Sure, just let me know and I will see if I can incorporate it in this PR.

jonkoops:
@xianli123 I took a note from your design and added the unit (seconds) after the input. Small difference is that I used PatternFly's built-in [Input Group](https://www.patternfly.org/v4/components/input-group/) component.

Screen Shot 2023-02-01 at 17 16 26

tnorimat:
@jonkoops

[1]  As for "interval" parameter, Keycloak's CIBA implementation treats 0 as "no value provided" and processes CIBA request according to [the CIBA specification](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.3), which means that Keycloak returns to a CD an authentication request acknowledgement without "interval" parameter.
  Therefore, my proposal is to show this specification explicitly in the tooltip's hint. 
  For example, in authentication-help.json,
  ```
  "cibaInterval": "The minimum amount of time in seconds that the CD (Consumption Device) must wait between polling requests to the token endpoint. If set to 0, the CD must use 5 as the default value according to the CIBA specification."
  ```

 [2] As for "Authentication Requested User Hint", my proposal is to show it in the new admin console the same as in the old admin console. The reason is as follows:
  - According to [the CIBA specification](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_request), there are three applicable values: `login_hint`, `id_token_hint`,  and `login_hint_token`.
  - The current Keycloak's CIBA implementation only supports `login_hint`.
    [The PR for CIBA](ttps://github.com/keycloak/keycloak/pull/7679 ) stated that
>This PR implements the provider that converts the user representation on CD, keycloak and the decoupled authentication server.
>
>The default implementation of this provider only support login_hint parameter. Also, this value of login_hint must be "username" or "email" attributes of the user registered on keycloak.

  - If "Authentication Requested User Hint" is not shown, the user of Keycloak's CIBA may not recognize that Keycloak uses `login_hint`.
  - In the future, there is a chance that `id_token_hint` and `login_hint_token` be supported.

xianli123:
Hi @jonkoops, the screen looks good. One question is why the "Authentication Requested User Hint" is removed. I found [tnorimat](https://github.com/tnorimat) also mentioned that. Do we need to implement it?

jonkoops:
@xianli123 as I mentioned in the PR description:
> The "Authentication Requested User Hint" field has been omitted, as there is only one value that can be selected at all times.

However considering the feedback from @tnorimat, I'll find a way to re-incorporate it.

jonkoops:
I've added the "Authentication Requested User Hint" field and updated the help text for the "Interval" field with @tnorimat's suggestion.

I've disabled the select for the "Authentication Requested User Hint"  field so that it is clear to the user that it's value cannot be changed. The help text for this field includes a note that "login_hint" is currently the only supported value. I wanted to put this information in a tooltip on the select as well, but this was not possible due to limitations in PatternFly.

Screen Shot 2023-02-02 at 12 37 36

@xianli123 @tnorimat can you let me know if this looks good to you now? @edewit could you also take another look at the new code?

xianli123 (approved):

edewit (approved):

tnorimat (approved):
@jonkoops It looks good to me.

jonkoops:
Awesome, I am merging this in. Thanks for your time and feedback all!

PR #4343, opened 2023-02-06
Remove theme.properties

stianst:
keycloak.v2 theme should not import the admin base theme, which is being deleted as it's part of the old admin console

## Checklist:


jonkoops (commented):
Although I agree with this change this was introduced to fix a bug (see https://github.com/keycloak/keycloak-ui/pull/2819). Does this re-introduce this issue?

PR #4346, opened 2023-02-07
Clean up realm selector code

jonkoops:
Cleans up the realm selector code a bit which will hopefully prevent failures in the Cypress test suite. Some components have been un-nested to prevent re-renders that might affect the tests.


edewit (approved):

jonkoops:
Merging this in as it might help the testing situation, also because I'd like to improve this further in another PR.

PR #4351, opened 2023-02-08
Introduce `useStoredState()` hook

jonkoops:
Introduces a `useStoredState()` hook to replace [`react-use-localstorage`](https://github.com/dance2die/react-use-localstorage), which has some bugs and is largely unmaintained. This hook also adds a higher level of serialization so that primitive values can be stored.

Wrote this as I needed something more high level for some other code I want to introduce, and the old library got in the way. Figured I might as well fix this up a little.


jonkoops:
Merging this in as I need it for some other work.

PR #4363, opened 2023-02-09
Fix 'Home URL' link in clients overview

jonkoops:
Ran into this bug when working on the clients overview for an unrelated reason. I noticed that the 'Home URL' links in the table on the clients overview were getting malformed (and also the link target).

Before:
Screenshot 2023-02-09 at 18-33-19 Keycloak Administration UI

After:
Screenshot 2023-02-09 at 18-33-40 Keycloak Administration UI


ssilvert (approved):
LGTM

edewit (approved):

PR #4381, opened 2023-02-13
just to see what it looks like without any css

edewit:
#3380

## Checklist:


edewit:
the stuff that looks bad are customisations what we do based on patternfly like using components inside components, for example:
image
image
image
image
image
image
This is no longer sticky only on the top of the page
image

PR #4389, opened 2023-02-13
Remove WebPack pre-proccessor from Cypress config

jonkoops:
Removes the WebPack pre-proccessor from Cypress config, which is no longer needed now that we do type checking on CI.

### DIFF
--- a/apps/admin-ui/cypress.config.mjs
+++ b/apps/admin-ui/cypress.config.mjs
@@ -1,9 +1,4 @@
-import webpackPreprocessor from "@cypress/webpack-batteries-included-preprocessor";
 import { defineConfig } from "cypress";
-import { deleteAsync } from "del";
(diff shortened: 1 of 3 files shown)


apps/admin-ui/cypress.config.mjs
```diff
@@ -23,30 +18,6 @@ export default defineConfig({
...
   e2e: {
-    setupNodeEvents(on) {
-      on(
-        "file:preprocessor",
-        webpackPreprocessor({
-          typescript: require.resolve("typescript"),
-        })
-      );
-
-      on("after:spec", async (spec, results) => {
```
jonkoops, line 34: Also removed this code that removes video files as we're no longer uploading these to GitHub.

edewit (approved):

PR #4393, opened 2023-02-14
moved loading of resources to row expand

edewit:
moved loading of resources to expand action

fixes: #4383

### DIFF
--- a/apps/admin-ui/src/clients/authorization/Scopes.tsx
+++ b/apps/admin-ui/src/clients/authorization/Scopes.tsx
@@ -24,7 +24,6 @@ import { KeycloakSpinner } from "../../components/keycloak-spinner/KeycloakSpinn
 import { PaginatingTableToolbar } from "../../components/table-toolbar/PaginatingTableToolbar";
 import { useAdminClient, useFetch } from "../../context/auth/AdminClient";
 import { useRealm } from "../../context/realm-context/RealmContext";
-import { MoreLabel } from "./MoreLabel";
 import { toScopeDetails } from "../routes/Scope";
 import { toNewScope } from "../routes/NewScope";
 import { ListEmptyState } from "../../components/list-empty-state/ListEmptyState";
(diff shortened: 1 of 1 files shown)


apps/admin-ui/src/clients/authorization/Scopes.tsx
```diff
@@ -87,27 +75,13 @@ export const AuthorizationScopes = ({ clientId }: ScopesProps) => {
         deep: false,
         name: search,
       };
-      const scopes = await adminClient.clients.listAllScopes({
+      return await adminClient.clients.listAllScopes({
```
jonkoops, line 78: If a Promise is returned here there is no need to `await` it's value. Also no need for the function it is in to be marked as `async`.

Side-note, this is coming up as a pretty common review comment. Perhaps we should enable the [`no-return-await`](https://eslint.org/docs/latest/rules/no-return-await) linting rule to prevent this in the future.

apps/admin-ui/src/clients/authorization/Scopes.tsx
```diff
@@ -131,6 +105,10 @@ export const AuthorizationScopes = ({ clientId }: ScopesProps) => {
             id: clientId,
             scopeId: scope.id!,
           }),
+          permissions: await adminClient.clients.listAllPermissionsByScope({
```
jonkoops, line 108: Awaiting the Promise here makes the `Promise.all` above inefficient, these calls should be made in parallel.

jonkoops (changes requested):

jonkoops (changes requested):
@edewit I think you can replace your fetching logic with this, which will remove the need to do that additional mapping:

```ts
const newlyOpened = collapsed
  .filter((row) => row.isExpanded)
  .map(({ id }) => getScope(id))
  .filter((s) => !s.isLoaded);

return Promise.all(
  newlyOpened.map(async (scope) => {
    const [resources, permissions] = await Promise.all([
      adminClient.clients.listAllResourcesByScope({
        id: clientId,
        scopeId: scope.id!,
      }),
      adminClient.clients.listAllPermissionsByScope({
        id: clientId,
        scopeId: scope.id!,
      }),
    ]);

    return {
      ...scope,
      resources,
      permissions,
      isLoaded: true,
    };
  })
);
```

jonkoops (approved):

PR #4395, opened 2023-02-14
Consistent language fallbacks for admin ui theme

### LINKED ISSUE #4390: Realm localization: Inconsistent message resolving regarding language fallbacks for different themes
### Describe the bug

This ticket is for aligning the Admin UI with the implementation done in this Keycloak issue: https://github.com/keycloak/keycloak/issues/15845

**IMPORTANT**: The keycloak issue should be merged first, otherwise the linked PR will NOT completely work.

It would be nice to have the exactly same fallback implementation as for the other themes, but unluckily the i18next library (https://github.com/i18next/i18next) seems NOT to support variants (like "de-CH-*1996*"), just regions (like "de-CH") and languages without region (like "de"), of course.
If variants should be supported, that feature could be implemented as a PR for i18next - or maybe i18next's language selection hook could be implemented accordingly.

The fallback priorities implemented in the linked PR are as follows (note that variant is NOT supported yet):

RL \<language-region> **>** T \<language-region> **>** RL \<language> **>** T \<language> **>** RL en **>** T en

(Abbreviations: RL = realm localization, T = Theme i18n files)

### Version

20.0.3

### Environment

all environments

### Expected behavior

GIVEN the supported locales for the realm are "de-CH", "de" and "en"
AND realm localization texts with key "common:realmSettings" are defined for the locales "de" and "en"
AND the user has selected the language "de-CH"
WHEN the user selects any page in the UI
THEN the sidebar shows the "realm settings" text for locale "de" 

### Actual behavior

GIVEN the supported locales for the realm are "de-CH", "de" and "en"
AND realm localization texts with key "realmSettings" are defined for the locales "de" and "en"
AND the user has selected the language "de-CH"
WHEN the user selects any page in the UI
THEN the sidebar shows the "realm settings" text for locale "en" 

It's incorrect that the "en" translation is shown, because it's less specific than the "de" translation.

Also note that the localization text keydoes not contain the namespace prefix "common:". Currently whether the prefix has to be part of the localization text key, depends on how it is called from the react code (with or without the default "common:" prefix). I think it's obvious that the namespace must always be part of the localization text key - or not at all.

### How to Reproduce?

see behavior description above.

### Anything else?

Precondition: https://github.com/keycloak/keycloak/issues/15845 has to be merged

danielFesenmeyer:
- i18next does not support variants, hence the prio of messages is as follows (RL = realm localization, T = Theme i18n files): RL \<region> > T \<region> > RL \<language> > T \<language> > RL en > T en
- remove the param useRealmDefaultLocaleFallback=true from the call to /admin/realms/{realm}/localization/{locale}, because otherwise realm localization texts of the realm default language would overwrite more specific texts from translation files (and unnecessary data will be loaded)
- add cypress test i18n_test.spec.ts, which checks the fallback implementation
- log a warning instead of an error, when messages for some languages/namespaces cannot be loaded (the page will probably work with fallbacks in that case)

Closes #4390

## Motivation
This is the admin ui part of https://github.com/keycloak/keycloak/pull/17076

## Brief Description
Changes the prio of localized messages as follows (RL = realm localization, T = Theme i18n files): RL \<region> > T \<region> > RL \<language> > T \<language> > RL en > T en

## Verification Steps
See the tests in i18n_test.spec.ts

## Checklist:

- [not relevant] Help has been implemented
- [?] axe report has been run and resulting a11y issues have been resolved

## Additional Notes
Only works completely when this Keycloak PR has been merged: https://github.com/keycloak/keycloak/pull/17076

### DIFF
--- a/apps/admin-ui/src/i18n.ts
+++ b/apps/admin-ui/src/i18n.ts
@@ -1,4 +1,4 @@
-import { init, use, InitOptions, TOptions } from "i18next";
+import i18n, { init, use, InitOptions, ResourceStore } from "i18next";
 import HttpBackend, { LoadPathOption } from "i18next-http-backend";
 import { initReactI18next } from "react-i18next";
 import type KeycloakAdminClient from "@keycloak/keycloak-admin-client";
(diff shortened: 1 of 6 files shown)


apps/admin-ui/src/i18n.ts
```diff
@@ -21,7 +72,7 @@ const initOptions = async (
     if (namespaces[0] === "overrides") {
       return `${addTrailingSlash(adminClient.baseUrl)}admin/realms/${
         adminClient.realmName
-      }/localization/{{lng}}?useRealmDefaultLocaleFallback=true`;
+      }/localization/{{lng}}`;
```
edewit, line 75: shouldn't we keep this `useRealmDefaultLocaleFallback` option? see https://github.com/keycloak/keycloak/issues/11851

edewit:
@ssilvert can you have a look at this

apps/admin-ui/src/i18n.ts
danielFesenmeyer: This is no longer needed, because the postProcessor is no longer used (for why, see my comments in i18n.ts, function addInterceptorForGetResource). The english (default) translation is now loaded in a separate request and applied, if a more specific translation is not available.

If the param `useRealmDefaultLocaleFallback` was still used, it would have the following disadvantages:
- english realm localization texts would overwrite more specific translations from the json files. For example, an english realm localization text would overwrite a german text from the json files, if locale de was requested. (The english realm localization text should only overwrite the english text from the json files.)
- unnecessary network traffic: due to loading of the same data multiple times - english texts would be both loaded for locale en, and for the user requested locale (e.g. de)

jonkoops:
Converting this to a draft since it requires upstream work to be merged first.

danielFesenmeyer:
Just fixed some formatting errors in the PR description.

apps/admin-ui/src/i18n.ts
```diff
@@ -8,10 +8,61 @@ import { getAuthorizationHeaders } from "./utils/getAuthorizationHeaders";
...
+ * <ul>
+ *     <li>The key might or might not contain the namespace, thus key parsing would need to be re-implemented.</li>
+ *     <li>In case of plurals, the key will not contain the "_one" or "_other" suffix - would need to be re-implemented</li>
+ *     <li>For the "overrides", interpolation (replacing placeholders) would need to be applied.</li>
+ *     <li>Probably more ...</li>
+ * </ul>
+ */
+function addInterceptorForGetResource() {
+  const functionNameGetResource = "getResource";
+  const i18nStore = i18n.store as any;
```
ssilvert, line 35: We try to stay away from the `any` type.  Looks like you may need to declare some new types?

ssilvert (changes requested):
Overall, I like what you are doing here.  Very nice.

apps/admin-ui/src/i18n.ts
danielFesenmeyer: Ok, I got rid of the `any` and declared a type instead.

Furthermore, I've adjusted the tests in order to run with cypress 12 (turned out they didn't).
Some of them will still fail until the upstream PR is merged: https://github.com/keycloak/keycloak/pull/17076

jonkoops:
Hi @danielFesenmeyer, we have moved the Admin UI to the main Keycloak repo as part of our mono-repoing efforts. Could you open up a PR with the same changes there? If you have a PR already open for this feature feel free to add the changes there.

danielFesenmeyer:
@jonkoops Thanks for the hint - I will create a corresponding PR at the keycloak repo.

danielFesenmeyer:
@jonkoops: Does that mean that the account UI is now also developed in the keycloak repo again?

ssilvert:
@danielFesenmeyer Just to be clear, V2 of the account console still lives in themes/src/main/resources/theme/keycloak.v2/account.

V3 is the one in the /js directory.  V3 is not shipping with keycloak yet.

PR #4398, opened 2023-02-14
Add Client Scopes mappers tests

Aboullos:
Add Client Scopes mappers tests


jonkoops (changes requested):
Looks like some of these tests are failing on CI.

Aboullos:
> Looks like some of these tests are failing on CI.

Yes, I removed the skip on a test and it seems like some changes from the rebase weren't applied to it because it was being skipped. And some optimizations that I had to do to another test, let's see how it goes

jonkoops (approved):

PR #4401, opened 2023-02-15
Use Cypress record key from secrets

jonkoops:
We're still using a hard-coded key for recording our Cypress tests. This information should not be stored publicly, so we need to get this from the secrets instead.


edewit (approved):

ssilvert (approved):

ssilvert:
Waiting for Jon to return before we merge this.

agagancarczyk (approved):

PR #4418, opened 2023-02-16
Make role restrictions on group tree match group table.

### LINKED ISSUE #3978: Group navigation has link when only query-groups
### Describe the bug

If the user has only query-groups role, then the link for groups in the lefthand tree navigation is still active.  If you click on it you will get an error.

The group names should not be rendered as a link unless the user has the `view-users` role.  This is correct in the list view but not in the tree view.

### Version

20.x

### How to Reproduce?

Give a user only `query-groups` role from the realm-management client.

Go to `Groups` section.

ssilvert:
## Motivation
Fixes #3978 

## Brief Description
If the user has only query-groups role, then the link for groups in the lefthand tree navigation is still active. If you click on it you will get an error.

The group names should not be rendered as a link unless the user has the view-users role. This is correct in the list view but not in the tree view.

## Verification Steps
Give a user only query-groups role from the realm-management client.

Go to Groups section.  See that the group names in the tree are not rendered as links and the kabab menu is hidden.

### DIFF
--- a/apps/admin-ui/src/groups/GroupTable.tsx
+++ b/apps/admin-ui/src/groups/GroupTable.tsx
@@ -21,9 +21,13 @@ import { MoveDialog } from "./components/MoveDialog";
 
 type GroupTableProps = {
   refresh: () => void;
+  canViewDetails: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


jonkoops (approved):
LGTM

PR #4457, opened 2023-02-24
revert use of randomUUID

edewit:
fixes: #4442

## Checklist:

### DIFF
--- a/apps/admin-ui/src/components/dynamic/MapComponent.tsx
+++ b/apps/admin-ui/src/components/dynamic/MapComponent.tsx
@@ -11,6 +11,7 @@ import { MinusCircleIcon, PlusCircleIcon } from "@patternfly/react-icons";
 import { useEffect, useState } from "react";
 import { useFormContext } from "react-hook-form";
 import { useTranslation } from "react-i18next";
+import { generateId } from "../../util";
 
 import { HelpItem } from "../help-enabler/HelpItem";
 import { KeyValueType } from "../key-value-form/key-value-convert";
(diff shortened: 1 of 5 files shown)


SamTV12345:
Is [this](https://www.npmjs.com/package/uuid) library not a better replacement for the problem? We currently hope that Math.random()*1000 doesn't generate the same number. UUIDs are [specified](https://www.rfc-editor.org/rfc/rfc4122) and should never appear twice.

edewit:
@SamTV12345 the number is used to identify an alert that stays on the page for 8 seconds so at most you have 3 of those at the same time. Even if they have the same number by change the worst that is going to happen is that one disappears a bit quicker. I don't think it's worth adding an entire library to prevent that.

SamTV12345:
@edewit It is just a bit cleaner with UUIDs. But I guess it won't make a difference in practice. I created another pr that solves the problem with the library included.

apps/admin-ui/src/components/dynamic/MapComponent.tsx
```diff
@@ -18,8 +18,9 @@ import type { ComponentProps } from "./components";
 import { convertToName } from "./DynamicComponents";
 
 type IdKeyValueType = KeyValueType & {
-  id: string;
+  id: number;
 };
+const generateId = () => Math.floor(Math.random() * 100);
```
jonkoops, line 23: Can we put this into a single utility function and re-use it across the project?

jonkoops (changes requested):

jonkoops:
Perhaps a better solution for generating IDs in React components would be to use the [`useId()`](https://beta.reactjs.org/reference/react/useId) hook which was introduced in React 18.

jonkoops (approved):

jonkoops (approved):

